PR #8, opened 2020-01-29
Use official docker image

ebyhr:
Additionally, modified the existing test_dbapi.py to suppress errors by version upgrade.


integration_tests/fixtures.py
```diff
@@ -64,33 +48,15 @@ def get_local_port():
...
-        ".",
-    ]
-    if not with_cache:
-        docker_build.append("--no-cache")
-    subprocess.check_call(docker_build)
-
-
 def get_default_presto_image_tag():
-    image_name = "presto-server"
+    image_name = "prestosql/presto"
```
findepi: inline:

```suggestion
    return "prestosql/presto:" + PRESTO_VERSION
```

integration_tests/fixtures.py
```diff
@@ -31,21 +29,7 @@
...
-    try:
-        return xml_root.find("versioning").find("release").text.strip()
-    except AttributeError:
-        raise Exception("not release found")
-
-
-# Temporary fix because can't start 0.203
-# Use get_latest_release() instead
-PRESTO_VERSION = os.environ.get("PRESTO_VERSION") or get_latest_release()
+PRESTO_VERSION = os.environ.get("PRESTO_VERSION") or "latest"
```
findepi: "latest" is not a presto version, so

```suggestion
PRESTO_IMAGE_VERSION = os.environ.get("PRESTO_VERSION") or "latest"
```

integration_tests/fixtures.py
```diff
@@ -109,7 +75,7 @@ def start_presto(image_tag=None, build=True, with_cache=True):
     return (container_id, run, "localhost", local_port)
 
 
-def wait_for_presto_workers(host, port, timeout=30):
+def wait_for_presto_workers(host, port, timeout=180):
```
findepi: is this a related change? 
separate commit?

integration_tests/fixtures.py
```diff
@@ -149,15 +115,15 @@ def start_local_presto_server(image_tag, build=True, with_cache=True):
...
+def start_presto_and_wait(image_tag=None):
     container_id = None
     proc = None
     host = os.environ.get("PRESTO_RUNNING_HOST", None)
     if host:
         port = os.environ.get("PRESTO_RUNNING_PORT", DEFAULT_PORT)
     else:
         container_id, proc, host, port = start_local_presto_server(
-            image_tag, build, with_cache
+            image_tag
```
findepi: wrap to prev line?

integration_tests/test_dbapi.py
```diff
@@ -120,6 +119,7 @@ def test_select_tpch_1000(presto_connection):
 def test_cancel_query(presto_connection):
     cur = presto_connection.cursor()
     cur.execute("select * from tpch.sf1.customer")
+    cur.fetchone()
```
findepi: how is this related?

integration_tests/test_dbapi.py
```diff
@@ -47,7 +47,6 @@ def test_select_query(presto_connection):
     rows = cur.fetchall()
     assert len(rows) > 0
     row = rows[0]
-    assert row[0] == "test"
```
findepi: the test query above should be using concrete column names instead of `*` (but that would be a separate commit)

i don't understand what `row[0]` was referring to and why it gets removed

integration_tests/test_dbapi.py
```diff
@@ -47,7 +47,6 @@ def test_select_query(presto_connection):
...
     assert row[2].split("-")[0] == fixtures.PRESTO_VERSION
```
findepi: - can `fixtures.PRESTO_VERSION` be "latest" here?
- i think you can remove `.split("-")[0]` from here -- switching to dockerhub-provided image means the version won't contain a hyphen

integration_tests/test_dbapi.py
ebyhr: Sorry, I forgot to submit issue. It comes from  dispatcher . https://github.com/prestosql/presto/issues/2683

integration_tests/test_dbapi.py
ebyhr: It was node_id specified in etc/node.properties. It's not a fixed value after this commit. 
```
presto> select * from system.runtime.nodes;
   node_id    |        http_uri        | node_version | coordinator | state
--------------+------------------------+--------------+-------------+--------
 b8c38cccce9d | http://172.17.0.2:8080 | 329          | true        | active
```

integration_tests/test_dbapi.py
ebyhr: Exactly, the value won't "latest" here. Let me fix it.

.gitignore
```diff
@@ -0,0 +1,2 @@
+.idea
```
findepi: i merged this commit, please rebase

integration_tests/fixtures.py
findepi: let's actually have
```
PRESTO_VERSION = os.environ.get("PRESTO_VERSION") or "329"
```

here

integration_tests/test_dbapi.py
findepi: let's add 

```
# TODO (https://github.com/prestosql/presto/issues/2683) test with and without .fetchone
```

Dockerfile
```diff
@@ -1,38 +0,0 @@
...
-RUN curl -o /$PRESTO_PKG $PRESTO_PKG_URL
-RUN tar -zxf /$PRESTO_PKG
-
-# Create directory for Presto data
-RUN mkdir -p /var/lib/presto/data
-
-# Add Presto configuration
-WORKDIR /presto-server-$PRESTO_VERSION
-RUN mkdir etc
-ADD etc/jvm.config etc/
```
findepi: we should remove etc/ as well

integration_tests/fixtures.py
ebyhr: It's unrelated and unnecessary. Reverted.

ebyhr:
Comments addressed.

PR #16, opened 2020-02-24
Added support for query params to dbapi module

akuzmichov:
Added support for query params to dbapi module, Cursor's method execute. This includes ParamsEscaper class.


martint:
@cla-bot check

akuzmichov:
I sent my signed CLA to cla@prestosql.io

presto/dbapi.py
```diff
@@ -41,6 +42,89 @@
...
+
+        :param Iterable sequence: The sequence to be escaped
+        :return: String containing the escaped given collection, with parentheses to make a proper SQL array
+        """
+        items = map(str, map(self._escape_item, sequence))
+        return '(' + ", ".join(items) + ')'
+
+    def _escape_item(self, item):
+        """
+        Escapes a single item of the parameters.
```
electrum: This seems incorrect, as NULL in SQL should be without quotes. This is transforming it into a string “NULL” (which is a valid non-null string).

akuzmichov: I see what you mean, and I see why this is misleading. It must be correct, however. These parameters are embedded into the SQL query in method `execute` using % operator - see line 320. And that SQL query (called `operation` there) actually IS a string in Python. So embedding a string 'NULL' into a string query is correct.

Hm, on the other hand, there can be usage inconsistency. If you write queries with `WHERE id = %(value)s` or `WHERE id IN %(values)s`, NULL doesn't work as a parameter, because NULL isn't equal to anything in SQL. The check for NULL is `IS NULL`. If we add a NULL parameter after the `IS` keyword, other parameter values won't work there. Which brings me to conclusion that checking NULL value is usually a specific enough situation to be written explicitly, without parameters. Also, SQL can be written like `WHERE phone IN %(values)s OR phone IS NULL`.
So I guess we should ignore None whatsoever, and make users move this logic to their queries. What do you think?

akuzmichov:
> Thanks for submitting this PR. Can you add tests, especially around the handling for None / null? We should execute queries with parameters and validate they return the expected results. The existing tests should helpfully provide a good starting point.

Sure, I will write tests later. I'm in GMT+3, so it is too late for today :)

presto/dbapi.py
electrum: Ah, I see what you mean. Thanks for explaining. How do the drivers for MySQL and PostgreSQL handle this?

akuzmichov: I don't have a working MySQL instance to check it in.
But I've just checked it in Postgres Python client psycopg2. Apparently it converts a Python None to NULL in the SQL query. It works in queries like `WHERE id IS %(none_param)s`, and doesn't work in queries like `WHERE id = %(none_param)s` - because NULL equality is never true in SQL standard.
Besides, I didn't find special `None` treatment in psycopg2's source code in C except converting Python's `None` to `NULL`.

See this source code file:
https://github.com/psycopg/psycopg2/blob/master/psycopg/cursor_type.c
Line 376 starts the cursor's execute method `_psyco_curs_execute`.
In the line 399, it calls function `_mogrify` which validates and prepares the query and vars (its parameters). The function `_mogrify` is lines 130-305. Lines 205-213 are of the most interest, where it says that "None (Py_None) is always converted to NULL".
After `_mogrify` call, the function `_psyco_curs_merge_query_args` is called in line 404, which is defined in lines 316-371. I didn't find any special Py_None or NULL processing in it.

All and all, my observation matches to my findings in psycopg2 source code. I guess we can do the same and keep converting None value to NULL in SQL.

akuzmichov:
@cla-bot check

integration_tests/fixtures.py
```diff
@@ -87,7 +87,7 @@ def wait_for_presto_workers(host, port, timeout=180):
         time.sleep(1)
 
 
-def wait_for_presto_coordinator(stream, timeout=180):
+def wait_for_presto_coordinator(stream, timeout=360):
```
akuzmichov: I could not pull image and run container in time otherwise. I guess I might not be alone with this issue (slow internet?)

akuzmichov: I could not pull image and run container in time otherwise. I guess I might not be alone with this issue (slow internet connection?)

presto/dbapi.py
```diff
@@ -41,6 +46,93 @@
...
+        Numbers returned as they are (they don't require escaping).
+        Strings are escaped in _escape_string().
+        Iterable collections are processed recursively down to individual booleans, numbers or strings.
+
+        :param item: The item to escape
+        :return: An int, a float, or a string
+        """
+        if item is None:
+            return 'NULL'
+        elif isinstance(item, bool):
```
akuzmichov: `True` and `False` are actually instances of `int` in Python, and they can be embedded into %s and %(named)s parameters via the % operator. I added this `isinstance(item, bool)` branch just for explicity.

akuzmichov:
Hey @electrum , I have added tests for parameters of the following types: bool, int, float, str, lists and tuples, NoneType. All these parameter types are tested using a single SQL query but with multiple asserts afterwards. There are 3 test functions: one for params dict, one for tuple, one for list. All the integration tests have passed in a local container on my mahcine.
I left the None->'NULL' conversion as is, because this is apparently what PostgreSQL Python client does.
Please take a look when you have the time.

Also, the Travis CI build is stuck in progress for some reason.

akuzmichov:
Hey @electrum , any chance you could take another look at this PR?

rushton:
fixes #21, thanks!

akuzmichov:
Hey @electrum , I'm sorry for a late reply too :)
I have fixed the conflict with master. The tests have passed successfully.

akuzmichov:
FYI, this page says "Only those with write access to this repository can merge pull requests."
So I need your help, @electrum :)

presto/dbapi.py
```diff
@@ -41,6 +46,93 @@
...
+    """
+
+    def escape_args(self, parameters):
+        """
+        Escapes string parameters for an SQL query in Presto. String parameters are escaped in _escape_string.
+        Boolean values are transformed to either 'TRUE' or 'FALSE' to be added to SQL query. Numbers in the
+        parameters are returned as they are. Iterable collections are processed recursively down to individual
+        booleans, numbers or strings. Python's None is transformed to 'NULL' string to be added as NULL to SQL query.
+
+        :param parameters: Either a dict, or a tuple, or a string with the parameters. If the query has unnamed
```
losipiuk, line 61: drop "or a string".  Only dict and tuple is supported

presto/dbapi.py
```diff
@@ -231,8 +323,9 @@ def setinputsizes(self, sizes):
     def setoutputsize(self, size, column):
         raise presto.exceptions.NotSupportedError
 
-    # TODO support params
-    def execute(self, operation):
+    def execute(self, operation, params=None):
+        if params:
+            operation = operation % ParamEscaper().escape_args(parameters=params)
```
losipiuk, line 328: nit: consider making escape_args a static method.

presto/dbapi.py
```diff
@@ -41,6 +46,93 @@
 logger = presto.logging.get_logger(__name__)
 
 
+class ParamEscaper(object):
+    """
+    Escapes parameters for an SQL query in Presto. The entry point is the public method escape_args.
+    """
+
+    def escape_args(self, parameters):
```
losipiuk, line 54: Could you please add a comprehensive unit test for this one, so we are sure all the corner cases are exercised?

losipiuk:
@akuzmichov please let me know when you have chance to address the comments, so we can merge this :)

findepi:
Thank you for your contribution. We ended up merging https://github.com/trinodb/trino-python-client/pull/41 instead.
I can understand that it's not the most enjoyable outcome and I do not remember why this happened.

I am closing this one, but please reopen or create a new PR if there is something here I am missing.


Anyway, @akuzmichov thank you for your contribution and looking forward to more!

akuzmichov:
Hi Piotr

That is fine, I am not upset. On the contrary, I feel sorry that I forgot
to finish this one. My apologies for that.
I may submit another PR if we in our team find some functionality lacking.

Good luck, and have a great day!

вт, 12 янв. 2021 г. в 18:53, Piotr Findeisen <notifications@github.com>:

PR #18, opened 2020-03-31
Add support for polling.

matthewwardrop:
This PR was originally put up against the original prestodb version of this library [https://github.com/prestodb/presto-python-client/pull/90] (which now seems to be defunct?); and adds support for polling the status of presto queries, decoupling it from collection of results (fixing #63 on that repository). This is important especially for long-running queries.

Changes:
- Removed `PrestoResult` and merged its functionality with `PrestoQuery`. Keeping both led to some weird logic loops as each tried to keep the other up to date once polling was added.
- Made `PrestoQuery.fetch` a private method `PrestoQuery._fetch`, since calling it directly leads to the iterator not being in sync.
- Made the `PrestoQuery` object cache the query results, otherwise there is some weird behaviour whereby if you stop iterating over the rows, you throw away all of the rows left over in the last retrieved chunk.
- Switched the `fetchone` / `fetchmany` relationship so that `fetchone` is a special case of `fetchmany` (because the changes above allow multiple iterators to be created over time, so long as there is only one at time, and in this way we don't create a new generator for every row).

### DIFF
--- a/presto/client.py
+++ b/presto/client.py
@@ -423,42 +423,14 @@ class PrestoRequest(object):
         )
 
 
-class PrestoResult(object):
+class PrestoQuery(object):
     """
-    Represent the result of a Presto query as an iterator on rows.
+    Represent the execution of a SQL statement by Presto.
 
-    This class implements the iterator protocol as a generator type
+    Results of the query can be extracted by iterating over this class, since it
+    implements the iterator protocol as a generator type
     https://docs.python.org/3/library/stdtypes.html#generator-types
     """
-
-    def __init__(self, query, rows=None):
-        self._query = query
-        self._rows = rows or []
-        self._rownumber = 0
-
-    @property
-    def rownumber(self):
-        # type: () -> int
-        return self._rownumber
-
-    def __iter__(self):
-        # Initial fetch from the first POST request
-        for row in self._rows:
-            self._rownumber += 1
-            yield row
-        self._rows = None
-
-        # Subsequent fetches from GET requests until next_uri is empty.
-        while not self._query.is_finished():
-            rows = self._query.fetch()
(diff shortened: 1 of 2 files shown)


matthewwardrop:
@ebyhr I've fixed things. There was a trivial oversight in keeping track of progress when requesting one row at a time. I've pushed the fix, and signed the CLA.

martint:
@cla-bot check

matthewwardrop:
@ebyhr @martint Any updates here?

mathieuseguin:
We're running into the same issue with long pulling queries.

Any chance we could see this being merged in a near future?

amitds1997:
@matthewwardrop Are you planning to work on this soon?

matthewwardrop:
@amitds1997 I'm not sure. My original PRs to the presto-python-client and then this one have "enjoyed" very slow review times, making it challenging to make much headway (by the time a review occurs, I've not thought much about the patch for many months). @findepi's comments were valid ones, but as far as I can tell I already answered these concerns in the original PR description. I'm happy to restage the PR as a series of commits if that is helpful, but note that it won't be entirely straightforward: these changes hang together, and each standalone commit won't make a tonne of sense without the context of the others. I'm also happy if someone else closer to the project wants to reimplement things in a different way, but after a review of this patch I still feel this is a pretty clean implementation. It does make several "unnecessary" changes, but only in the sense that you without the other changes there are weird epicycles in the code and the risk of losing data depending on the order in which you call methods on the class instances.

I originally wrote this patch so I could migrate to this more official client from `pyhive`, since I thought things would wind down there.... but things on that side have continued kicking along, and their Presto client already features this polling behaviour, so I haven't been too motivated to push hard on this.

PR #19, opened 2020-04-01
Developer improvements

matthewwardrop:
This scratches a few development itches that I encountered while working on the project.

1) The gitignore did not cover the usual things like tox environments. I replaced it with the GitHub template.
2) Pytest was not set up to report verbosely, meaning that when errors occurred it was more difficult to identify exactly which test was failing.

I also added coverage for Python 3.7 (everything passes, so why not?), and set up tox to read the test dependencies from the setup.py configuration (there were some missing dependencies otherwise for local tox runs, such as requests_kerberos).

Feel free to accept or reject any of these "improvements".

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,5 @@
 envlist = py27,py35,py36,pypy2
 
 [testenv]
-deps = pytest
-       httpretty
+extras = tests
 commands = pytest -s tests/ integration_tests/
--- a/tox.ini
+++ b/tox.ini
@@ -3,4 +3,4 @@ envlist = py27,py35,py36,pypy2
 
 [testenv]
 extras = tests
-commands = pytest -s tests/ integration_tests/
+commands = pytest -sv tests/ integration_tests/
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py35,py36,pypy2
+envlist = py27,py35,py36,py37,pypy2
 
 [testenv]
 extras = tests
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,143 @@
-.idea/
-/dist
-/.venv/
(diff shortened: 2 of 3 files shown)


martint:
@cla-bot check

tox.ini
```diff
@@ -1,7 +1,6 @@
 [tox]
-envlist = py27,py35,py36,pypy2
+envlist = py27,py35,py36,py37,pypy2
```
findepi, line 2: Should we modify .travis.yml too?

matthewwardrop: Good catch!

tox.ini
```diff
@@ -1,7 +1,6 @@
...
 
 [testenv]
-deps = pytest
-       httpretty
-commands = pytest -s tests/ integration_tests/
+extras = tests
+commands = pytest -sv tests/ integration_tests/
```
findepi, line 6: I know this commit is just 4 lines, but this is 3 separate changes

* adding py3.7
* adding `-v`
* removing httppretty

all changes are trivial, but all (except py 3.7) should come with some explanation (eg i should know that httppretty was not used, or that -v is needed to see certain info in test output)

would you be able to split them into 3 commits, each with such explanation?

i know this is additional work for you, but this is just in case we need to return back to those commits looking for  "why so?"

matthewwardrop: A more than fair request. :). I'll restage them tomorrow!

findepi:
Merged as f7c3b4b2f56cb86944e03c2a64a8ec474ee927f8, thanks!

I updated the messages to align them with our guidelines (https://chris.beams.io/posts/git-commit/).
I will add a ref to them in the README too (https://github.com/prestosql/presto-python-client/pull/20).

Thank you for your contribution!

matthewwardrop:
Roger that! Thanks @findepi.

PR #23, opened 2020-04-26
Remove default redirect_handler

findepi:
Let the HTTP session handler the redirects with `allow_redirects=True`
by default.

Fixes https://github.com/prestosql/presto-python-client/issues/17


electrum:
I think the purpose of `GatewayRedirectHandler` is to canonicalize the hostname in the URL, likely for a reason related to Kerberos or TLS certificates. However, this seems like the wrong place to handle it. I suggest removing the redirect handler support entirely.

findepi:
@electrum i agree, my assessment of the redirect handler is the same. 
Let's do small steps, just in case someone relied on this.
I will merge as is, and create follow up issue to remove the redirect handler altogether.

PR #35, opened 2020-06-30
Make Cursor implement iterable

### LINKED ISSUE #33: Cursor should implement __iter__
The Cursor class in `dbapi.py` doesn't implement `__iter__` today meaning that code like `csv_writer.writerows(cursor)` or anything that tries to loop over the Cursor will not work.

hashhar:
Fixes #33.

### DIFF
--- a/integration_tests/test_dbapi.py
+++ b/integration_tests/test_dbapi.py
@@ -68,6 +68,21 @@ def test_select_query_result_iteration(presto_connection):
     assert len(list(rows0)) == len(rows1)
 
 
+def test_select_cursor_iteration(presto_connection):
+    cur0 = presto_connection.cursor()
+    cur0.execute("select nationkey from tpch.sf1.nation")
+    rows0 = []
+    for row in cur0:
+        rows0.append(row)
+
+    cur1 = presto_connection.cursor()
+    cur1.execute("select nationkey from tpch.sf1.nation")
+    rows1 = cur1.fetchall()
+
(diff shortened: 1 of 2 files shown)


hashhar:
Tests are pending.

Also, the current impl ends up delegating to `PrestoResult::__iter__` which means that it's somewhat lazily evaluated.

hashhar:
Existing tests are failing because co-ordinator was unable to start within 180 seconds on Travis. Tests are working locally though.

integration_tests/test_dbapi.py
```diff
@@ -68,6 +68,20 @@ def test_select_query_result_iteration(presto_connection):
...
+    cur0.execute("select custkey from tpch.sf1.customer LIMIT 10")
+    rows0 = []
+    for row in cur0:
+        rows0.append(row)
+
+    cur1 = presto_connection.cursor()
+    cur1.execute("select custkey from tpch.sf1.customer LIMIT 10")
+    rows1 = cur1.fetchall()
+
+    assert len(rows0) == len(rows1)
```
findepi, line 82: - you need ORDER BY to get stable results
- you can read from `nation`, test will be faster (then no need for LIMIT)

hashhar: Changed to use `nation` and also asserts on the actual rows to in addition to length now.

findepi:
Merged, thanks!

findepi:
Released in `0.301.0`

PR #36, opened 2020-08-17
Support self-signed SSL certificates

mik-laj:
Hello,

I would like to add support for self-signed SSL certificates. Requests can ignore verifying the SSL certificate if we set verify to False: Unfortunately, this is not possible in this library. Unfortunately, in this library it is not possible due to one incorrect condition in if-statement.
https://requests.readthedocs.io/en/master/user/advanced/#ssl-cert-verification
This is what I need to work on integration testing for Kerberos + Presto. I currently have a prepared environment.
https://github.com/mik-laj/presto-kerberos-docker
In the next step, I would like to integrate it with [Apache Airflow](https://github.com/apache/airflow).

**Question for the project maintainer:**
Would you like to use this environment in your CI/CD as well to test authorizations with Kerberos? Currently, they are very easy to configure. Just call ``start.sh`` and a full environment in Docker will be set up.

Best regards,
Kamil

### DIFF
--- a/presto/auth.py
+++ b/presto/auth.py
@@ -85,7 +85,7 @@ class KerberosAuthentication(Authentication):
             delegate=self._delegate,
             service=self._service_name,
         )
-        if self._ca_bundle:
+        if self._ca_bundle is not None:
             http_session.verify = self._ca_bundle
         return http_session
 


presto/auth.py
```diff
@@ -85,7 +85,7 @@ def set_http_session(self, http_session):
             delegate=self._delegate,
             service=self._service_name,
         )
-        if self._ca_bundle:
+        if self._ca_bundle is not None:
```
findepi, line 88: Is your intention to have _ca_bundle=False here so that `http_session.verify` will be set to `False`?

findepi:
> This is what I need to work on integration testing for Kerberos + Presto. I currently have a prepared environment.
> https://github.com/mik-laj/presto-kerberos-docker
> In the next step, I would like to integrate it with [Apache Airflow](https://github.com/apache/airflow).
> 
> **Question for the project maintainer:**
> Would you like to use this environment in your CI/CD as well to test authorizations with Kerberos? Currently, they are very easy to configure. Just call `start.sh` and a full environment in Docker will be set up.

This is definitely interesting!

Let's start a new issue about this and we will exchange thoughts there.

presto/auth.py
mik-laj: Yes. Exactly.

findepi:
Indeed, this might be confusing, but there might be a reason for that. @electrum can you comment?

presto/auth.py
hashhar: Does something like how PostgreSQL (and others) construct their URLs make sense? i.e. `?ssl=true&sslmode=verify&sslcert=cert.crt`

`ssl` (http/https) and `sslcert` (ca bundle) are already there.
`sslmode` can be introduced as a new parameter.

PR #41, opened 2020-08-27
Added query params functionality to Cursor.execute via PREPARE statement

harph:


presto/dbapi.py
```diff
@@ -22,7 +22,16 @@
 from __future__ import print_function
```
hashhar, line 22: Since we are using `?` as the parameter style, we should add a module level variable `paramstyle` set to `?` to ensure we comply with the DB-API (since most tools assume the default to be `%(name)` style parameters).

See https://www.python.org/dev/peps/pep-0249/#paramstyle

hashhar:
@harph The code solves the core problem correctly and the ideas look good.

But this might need some changes around usability (prepared statements are not re-used nor can the user try to re-use them, signature change of public methods allowing to pass headers etc.) and DB-API compliance.

I'll follow up with a detailed review in a while.

presto/dbapi.py
```diff
@@ -236,10 +245,111 @@ def setinputsizes(self, sizes):
...
+
+        raise presto.exceptions.FailedToObtainAddedPrepareHeader
+
+    def _get_added_prepare_statement_presto_query(
+        self,
+        added_prepare_header,
+        statement_name,
+        params
+    ):
+        sql = 'EXECUTE ' + statement_name + ' USING ' + ','.join(map(str, params))
```
hashhar, line 290: This method implicitly plays with the string representation of `params` which can cause issues (large numbers which python prints using scientific notation, arrays etc.).

presto/dbapi.py
```diff
@@ -236,10 +245,111 @@ def setinputsizes(self, sizes):
     def setoutputsize(self, size, column):
         raise presto.exceptions.NotSupportedError
 
-    # TODO support params
-    def execute(self, operation):
-        self._query = presto.client.PrestoQuery(self._request, sql=operation)
-        result = self._query.execute()
+    def _execute_prepare_statement(self, operation, statement_name):
```
hashhar, line 248: Maybe rename to `prepare_statement` since we don't actually `EXECUTE` here.

presto/dbapi.py
```diff
@@ -236,10 +245,111 @@ def setinputsizes(self, sizes):
...
+        for _ in result:
+            response_headers = result.response_headers
+
+            if constants.HEADER_DEALLOCATED_PREPARE in response_headers:
+                return response_headers[constants.HEADER_DEALLOCATED_PREPARE]
+
+        raise presto.exceptions.FailedToObtainDeallocatedPrepareHeader
+
+    def execute(self, operation, params=None):
+        if params:
```
hashhar, line 319: @findepi @electrum Should we start looking at how other DB-API drivers in Python solve this by providing multiple Cursor classes and over-ride the `execute()` and related functions accordingly in the sub-classes?

Will keep things cleaner and allow adding new Cursors that can maybe return the data as list of dicts etc.

hashhar:
In my opinion, we should consider adding a new Cursor subclass and the `connect()` method can take an argument to return the correct Cursor class.
This will allow the `PreparedCursor` class to cache prepared statements and transparently re-use them (see https://github.com/mysql/mysql-connector-python/blob/master/lib/mysql/connector/cursor.py#L1089 for the idea).

What I am aiming for is to allow the user control over the preparing, execution and deallocation of statements. Right now all of them are coupled within the `execute()` method.

findepi:
> This will allow the PreparedCursor class to cache prepared statements and transparently re-use them

I think this is not as important for Presto as it may be for  eg MySQL.
MySQL is specializing in low latency queries, so caching is more important for them.

Also, Presto re-plans prepared staement every time currently. Idk whether MySQL does the same.

> What I am aiming for is to allow the user control over the preparing, execution and de-allocation of statements. Right now all of them are coupled within the execute() method.

More control is also more complexity (for users, for maintainers).
I wouldn't "offer" more complexity to users unless this is really needed.

> Another possibility is to merge and take this in a follow up PR (risky since after release people will start depending on the current behaviour).

 what is the end-user visible behavior users could become attached to?

hashhar:
> Also, Presto re-plans prepared staement every time currently. Idk whether MySQL does the same.

Then I think we can go with the current implementation since the benefit I was thinking of of re-using prepared statements doesn't really exist.

> More control is also more complexity (for users, for maintainers).
> I wouldn't "offer" more complexity to users unless this is really needed.

Same as above since the control I was thinking about was something like a user can prepare 2 queries ahead of time and fire them within a loop based on some condition. I wanted to change:

```
for i in 1 to 10:
  if i % 2:
    s1 := PREPARE
    EXECUTE(s1)
    DEALLOCATE(s1)
  else:
    s2 := PREPARE
    EXECUTE(s2)
    DEALLOCATE(s2)
```
to 

```
s1 := PREPARE
s2 := PREPARE
for i in 1 to 10:
  if i % 2:
    EXECUTE(s1)
  else:
    EXECUTE(s2)

DEALLOCATE(s1)
DEALLOCATE(s2)
```
> > Another possibility is to merge and take this in a follow up PR (risky since after release people will start depending on the current behaviour).

The `additional_headers` argument that the `client.execute()` method now takes. I can imagine people using that to pass authentication headers/custom headers on a per-query basis to Presto.

With the above clarification around re-planning the query and no benefit to caching the PREPAREd statement I think the only thing left to discuss is the `additional_headers` on the client.

findepi:
> The additional_headers argument that the client.execute() method now takes.

Can we avoid this?
Ideally we keep the user-facing code surface minimal, for better future of our selves :)

hashhar:
> > The additional_headers argument that the client.execute() method now takes.
> 
> Can we avoid this?
> Ideally we keep the user-facing code surface minimal, for better future of our selves :)

One way could be model it similar to transactions? We add an attribute (dict) to `PrestoRequest` to track the prepared statement name to the value of `X-Presto-Added-Prepare`.

The flow can then be:

1. User calls `cursor.execute(sql, params)`
2. If `params` are present, generate a `statement_name` and call `cursor._prepare_statement(sql, statement_name)`
3. In `cursor._prepare_statement(sql, statement_name)`:
    - Create a copy of `self._request` (ctor vs deepcopy?)
    - Execute `PREPARE {statement_name} FROM {sql}` using the copied `request`
    - Create a new `self._request` via constructor and pass in the `X-Presto-Added-Prepare` response value as `prepared_statement` to PrestoRequest
    - Return the new `request` object that can be used to execute queries with the prepared statement (similar to Transactions)
4. Call `cursor._execute_prepare_statement(sql, request, params)` which execute `EXECUTE {request.prepared_statement_name} USING {params}` using the new `request`
5. Call `cursor._deallocate_prepare(request)` which execute `DEALLOCATE PREPARE {request.prepared_statement_name}`

Another option could be to wrap the behaviour from 3 to 5 inside a `PreparedStatement` class. Helps keep things compartmentalized.

In this scenario the only user-visible change would be the new `params` property to `cursor.execute()`.

presto/dbapi.py
harph: @hashhar Are you suggesting adding `paramstyle = '?'` as a  module variable/constant so that people can reference it when writing queries?

hashhar: Yes. Both for users writing queries and external tools that generate queries (I've seen some for PostgreSQL).

presto/dbapi.py
harph: Good point. What do you suggest in this case?

I think it's fair to assume that any complex object in the `params` the  must implement the `__str__` function  in order to be treated properly. Otherwise, they should be parsed/converted on the user's implementation before appending them to the `params`.

presto/dbapi.py
harph: Sounds good. I'll do the same to with the deallocate part of it.

presto/dbapi.py
harph: Perfect. I'll make that change.

findepi:
> In this scenario the only user-visible change would be the new `params` property to `cursor.execute()`.

that's the goal, so i'm fine with this :)


> The additional_headers argument that the client.execute() method now takes

i mis-read this perviously, sorry. I hope we can consider `client` as the impl details of the module, 
so we do not expose new contract fo users here.

Is my assumption now sound?

presto/dbapi.py
hashhar: @harph You are correct. I think we can leave as is in that case but clarify this in the docstring of the method to avoid people like me being caught by surprise.

**That still leaves one concern and that is regarding `params` whose values themselves include a `,`.**

Been a little long since I used Python with any DB. Almost every DB interprets the values passes as parameters as `LITERALS` by converting to a string.

hashhar:
@findepi Yes, if we consider `client` to be internal (the module docstring makes it **sound** public) then the PR can be merged once the `paramstyle`, tests for params with `,` and new method names are renamed.

Sorry for the unnecessary theoretical journey in that case. :sweat_smile:

presto/dbapi.py
hashhar: https://github.com/prestosql/presto-python-client/pull/16/files#diff-e6510cf8abf3e99495eee2b0d39cd839R54 might be useful.

@harph

presto/dbapi.py
```diff
@@ -236,10 +245,111 @@ def setinputsizes(self, sizes):
...
+            response_headers = result.response_headers
+
+            if constants.HEADER_ADDED_PREPARE in response_headers:
+                return response_headers[constants.HEADER_ADDED_PREPARE]
+
+        raise presto.exceptions.FailedToObtainAddedPrepareHeader
+
+    def _get_added_prepare_statement_presto_query(
+        self,
+        added_prepare_header,
```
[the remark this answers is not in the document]
harph: good catch. I'll remove it

presto/dbapi.py
findepi: >I think it's fair to assume that any complex object in the params the must implement the __str__ function in order to be treated properly.

__str__ has different contract and different requirements than Presto's parser.

I'd rather err on the side of not supporting a type, rather than call str on something I do not know.

I assume that supporting totally custom types is not a requirement, so we do not have to worry about them.

presto/exceptions.py
```diff
@@ -203,3 +203,19 @@ class DataError(DatabaseError):
...
+    """
+    pass
+
+
+class FailedToObtainDeallocatedPrepareHeader(Error):
+    """
+    Raise this exception when unable to find the 'X-Presto-Deallocated-Prepare'
+    header in the response of a DEALLOCATED statement request.
+    """
+    pass
```
losipiuk, line 221: add newline

presto/dbapi.py
```diff
@@ -236,10 +245,111 @@ def setinputsizes(self, sizes):
...
+            result = self._query.execute(
+                additional_http_headers={
+                    constants.HEADER_PREPARED_STATEMENT: added_prepare_header
+                }
+            )
+
+            # Send deallocate statement
+            # At this point the query can be deallocated since it has already
+            # been executed
+            self._execute_deallocate_prepare(added_prepare_header, statement_name)
```
losipiuk, line 348: shouldn't we wrap section above in `try:` and execute `_execute_deallocate_prepare` also for flow with exeption in `finally:`?

presto/dbapi.py
```diff
@@ -236,10 +245,111 @@ def setinputsizes(self, sizes):
...
+        raise presto.exceptions.FailedToObtainDeallocatedPrepareHeader
+
+    def execute(self, operation, params=None):
+        if params:
+            assert isinstance(params, (list, tuple)), (
+                'params must be a list or tuple containing the query '
+                'parameter values'
+            )
+
+            # Generate a unique statement name as convinient way to avoid
```
losipiuk, line 325: instead having a comment here I would suggest extracting `_generate_unique_statement_name` function. Calling it will be self explanatory.

presto/dbapi.py
```diff
@@ -236,10 +245,111 @@ def setinputsizes(self, sizes):
...
+                operation, statement_name
+            )
+
+            # Send execute statement and assign the return value to `results`
+            # is it will be returned by the function
+            self._query = self._get_added_prepare_statement_presto_query(
+                added_prepare_header, statement_name, params
+            )
+            result = self._query.execute(
+                additional_http_headers={
```
losipiuk, line 340: I was thinking that maybe we should change headers map in  `self._query`. But passing it explicitly here seems fine too.

tests/test_client.py
```diff
@@ -581,3 +621,71 @@ def test_gateway_redirect(monkeypatch):
...
+            request=req,
+            sql=sql
+        )
+        result = query.execute(additional_http_headers=additional_headers)
+
+        # Validate the the post function was called with the right argguments
+        mock_post.assert_called_once_with(sql, additional_headers)
+
+        # Validate the result is an instance of PrestoResult
+        assert isinstance(result, PrestoResult)
```
losipiuk, line 691: new line

presto/dbapi.py
```diff
@@ -236,10 +245,111 @@ def setinputsizes(self, sizes):
...
+            header.
+        """
+        sql = 'PREPARE {statement_name} FROM {operation}'.format(
+            statement_name=statement_name,
+            operation=operation
+        )
+
+        # Send prepare statement. Copy the _request object to avoid poluting the
+        # one that is going to be used to execute the actual operation.
+        query = presto.client.PrestoQuery(copy.deepcopy(self._request), sql=sql)
```
losipiuk, line 271: Can't we use `self._create_request()` here?
Is it important to call `PREPARE` using request containing information regarding current transaction?

tests/test_client.py
harph: done

presto/exceptions.py
harph: done

presto/dbapi.py
harph: The reason why I decided to do it this way is that the `Cursor` class receives a request object which can technically be different from what `_create_request` originally returns, if we assume that `Cursor` class can be instantiated directly. Also `Cursor` does no have access to the `_create_request` function since that one belongs to the `Connection` class.

presto/dbapi.py
harph: Good point. I'll do that.

presto/dbapi.py
harph: Addressed.

presto/dbapi.py
harph: done

presto/client.py
```diff
@@ -551,3 +563,7 @@ def cancel(self):
     def is_finished(self):
         # type: () -> bool
         return self._finished
+
+    @property
+    def response_headers(self):
```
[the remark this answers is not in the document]
harph: That does not apply to `post`, `get`, and `delete` because they belong to the `PrestoRequest` instead of the `PrestoQuery` class. Did you mean `execute` and `cancel` instead?

presto/dbapi.py
harph: Done

presto/dbapi.py
harph: Done

harph:
@hashhar @findepi Just to confirm, are we ok with leaving `addtional_headers` as parameter of the `client.execute` function given that is consider part of the module implementation?

findepi:
> @hashhar @findepi Just to confirm, are we ok with leaving `addtional_headers` as parameter of the `client.execute` function given that is consider part of the module implementation?

i am ok

presto/client.py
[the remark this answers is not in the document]
hashhar: You are correct. I should've said `execute` and `cancel`. I checked the headers returned on `execute` and `cancel` but they didn't look useful enough generally (yet) so we can let this go.

presto/dbapi.py
```diff
@@ -236,10 +246,113 @@ def setinputsizes(self, sizes):
...
+            statement_name = self._generate_unique_statement_name()
+
+            try:
+                # Send prepare statement
+                added_prepare_header = self._prepare_statement(
+                    operation, statement_name
+                )
+
+                # Send execute statement and assign the return value to `results`
+                # is it will be returned by the function
```
hashhar, line 338: nit: typo

```suggestion
                # as it will be returned by the function
```

hashhar (approved):
Looks good to go % the discussion about `_str_` vs what was proposed in [ParamEscaper](https://github.com/prestosql/presto-python-client/pull/16/files#diff-e6510cf8abf3e99495eee2b0d39cd839R49) in another PR.

presto/dbapi.py
harph: Fixed

presto/dbapi.py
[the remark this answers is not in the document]
harph: Done

harph:
What's the plan for merging and releasing this?

findepi:
@hashhar any other comments?
@losipiuk do you want to take a look?

otherwise I can merge & release

hashhar:
> Using `_str_` vs what was proposed in [ParamEscaper](https://github.com/prestosql/presto-python-client/pull/16/files#diff-e6510cf8abf3e99495eee2b0d39cd839R49) in another PR.

@findepi @harph should this be a follow up PR?

findepi:
> Converting params by calling str vs making the code type-aware and type-safe.
> 
> @findepi @harph should this be a follow up PR?

i'm fine, please make sure we have issues covering this

harph:
> > Using `_str_` vs what was proposed in [ParamEscaper](https://github.com/prestosql/presto-python-client/pull/16/files#diff-e6510cf8abf3e99495eee2b0d39cd839R49) in another PR.
> 
> Converting params by calling str vs making the code type-aware and type-safe.
> 
> @findepi @harph should this be a follow up PR?

I'm ok addressing it on this one.

losipiuk:
I am fine with merging this as is and do improvements in followup PR.

losipiuk (approved):

hashhar (approved):
LGTM :shipit:

findepi:
Released in `0.301.0`

ps-cbsantos:
Hi all,

I'm trying to use %s as the placeholder to my stmt, but always getting the error bellow:
`TrinoUserError(type=USER_ERROR, name=SYNTAX_ERROR, message="line 1:130: mismatched input '='. Expecting: <predicate>", query_id=20211221_205415_00100_hau44)`

The query I'm trying to run:
`cursor.execute("select  * from system.runtime.nodes where status=%s", ['active'])`

I followed the ways as described in the [test](https://github.com/trinodb/trino-python-client/pull/16/files#diff-e6510cf8abf3e99495eee2b0d39cd839R49), but could not get it done.

harph:
@ps-cbsantos take a look at this test that show you how it works: https://github.com/trinodb/trino-python-client/blob/a7438556213bcf954f8a7a5063c7823bbb177edb/integration_tests/test_dbapi.py#L73

You basically have to replace `%s` by `?`

Issue #48, opened 2020-10-14
Query parameters broken for string parameters


rushton:
The following assertion fails:
```python
      cur.execute(
          """
          select * from (
              values
              (1, 'one', 'a'),
              (2, 'two', 'b'),
              (3, 'three', 'c'),
              (4, 'four', 'd'),
              (5, 'five', 'e')
          ) x (id, name, letter)
          where name >= ?
          """,
          params=("one",),  # expecting all the rows with id >= 3
      )

      rows = cur.fetchall()

      assert len(rows) == 1
      assert rows[0][1] == "one"
```
with the following error:

```
presto.exceptions.PrestoUserError: PrestoUserError(type=USER_ERROR, name=EXPRESSION_NOT_CONSTANT, message="line 1:51: Constant expression cannot contain column references", query_id=20201014_222020_00005_ssyxh)
```

the generated statement is incorrect because it does not quote the string argument properly:

```sql
EXECUTE st_0c49e903f9cb4331ab305ac778bbae43 USING one
```

should be:
```sql
EXECUTE st_0c49e903f9cb4331ab305ac778bbae43 USING 'one'
```

rushton:
possibly related to #45, a result of changes made in #41

ebyhr:
Already fixed in #50

PR #50, opened 2020-10-14
Fix use of string parameters in prepared statements

rushton:
Prior to this commit parameters passed to execute were added to the
EXECUTE statement as naked strings. When using strings this causes a
syntax error in the query because Presto thinks it's a constant.
Example:
```
EXECUTE "st-XXX" USING Bob
```
results in:
```
Query 20201014_232548_02077_8sytd failed: line 1:22: Constant expression cannot contain column references
```
In this commit, string parameters passed to the execute are single
quoted to form a proper query. Care has been taken to escape single
quotes to avoid other query syntax errors and SQL injection.

### DIFF
--- a/integration_tests/test_dbapi.py
+++ b/integration_tests/test_dbapi.py
@@ -12,8 +12,6 @@
 from __future__ import absolute_import, division, print_function
 
 import fixtures
-import presto
-import pytest
 from fixtures import run_presto
 import pytest
 
@@ -89,12 +87,22 @@ def test_select_query_result_iteration_statement_params(presto_connection):
         params=(3,)  # expecting all the rows with id >= 3
     )
 
-    rows = cur.fetchall()
-    assert len(rows) == 3
+def test_select_query_result_string_statement_params(presto_connection):
+    """
+    Tests string parameters in prepared statements
+    """
+    cur = presto_connection.cursor()
+    # test a simple string parameter and
+    # a parameter with a quote in it
+    cur.execute(
+        "select ?",
+        params=("six'",),
+    )
 
-    for row in rows:
-        # Validate that all the ids of the returned rows are greather or equals than 3
-        assert row[0] >= 3
+    rows = cur.fetchall()
+
+    assert len(rows) == 1
+    assert rows[0][0] == "six'"
 
 
 @pytest.mark.parametrize('params', [
--- a/presto/dbapi.py
+++ b/presto/dbapi.py
@@ -281,16 +281,31 @@ class Cursor(object):
 
         raise presto.exceptions.FailedToObtainAddedPrepareHeader
 
+
     def _get_added_prepare_statement_presto_query(
         self,
         statement_name,
         params
     ):
-        sql = 'EXECUTE ' + statement_name + ' USING ' + ','.join(map(str, params))
+        sql = 'EXECUTE ' + statement_name + ' USING ' + ','.join(map(self._format_prepared_param, params))
 
         # No need to deepcopy _request here because this is the actual request
         # operation
         return presto.client.PrestoQuery(self._request, sql=sql)
+
+    def _format_prepared_param(self, param):
+        """
+        Formats parameters to be passed in an
+        EXECUTE statement.
+        """
+        if isinstance(param, str):
+            return ("'%s'" % param.replace("'", "''"))
+        else:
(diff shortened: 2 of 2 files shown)


rushton:
Fixes #48

rushton:
@harph I know this issue was mentioned in the original MR for query parameters, but seems use with string parameters is broken. Let me know if I'm missing the larger direction for this or if my implementation is being too naive.

presto/dbapi.py
```diff
@@ -281,12 +281,24 @@ def _prepare_statement(self, operation, statement_name):
...
+        """
+        Formats parameters to be passed in an
+        EXECUTE statement.
+        """
+        if isinstance(param, str):
+            # escape single quotes to protect against
+            # SQL injection in the execute
+            return ("'%s'" % param.replace("'", "''"))
+        else:
+            return str(param)
```
findepi, line 294: what else type `param` could be?

since we support python 2.7 still, this could be eg unicode.

also, there are many other cases which would need to be handled correctly (maps, lists, floating point infinity or NaN, maybe others), so please file an issue and add TODO here (with a link)

presto/dbapi.py
```diff
@@ -281,12 +281,24 @@ def _prepare_statement(self, operation, statement_name):
 
         raise presto.exceptions.FailedToObtainAddedPrepareHeader
 
+    def _format_prepared_param(self, param):
```
findepi, line 284: please move this below _get_added_prepare_statement_presto_query

findepi (commented):
@rushton thanks for working on this!

presto/dbapi.py
```diff
@@ -281,12 +281,24 @@ def _prepare_statement(self, operation, statement_name):
...
+        """
+        Formats parameters to be passed in an
+        EXECUTE statement.
+        """
+        if isinstance(param, str):
+            # escape single quotes to protect against
+            # SQL injection in the execute
```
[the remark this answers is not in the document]
rushton: removed

integration_tests/test_dbapi.py
```diff
@@ -82,26 +84,52 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+        "NOT A LIST OR TUPPLE",
+        {"invalid", "params"},
+        object,
+    ],
+)
 def test_select_query_invalid_params(presto_connection, params):
     cur = presto_connection.cursor()
     with pytest.raises(AssertionError):
-        cur.execute('select ?', params=params)
+        cur.execute("select ?", params=params)
```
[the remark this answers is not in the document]
rushton: reverted

integration_tests/test_dbapi.py
```diff
@@ -82,26 +84,52 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
-    {'invalid', 'params'},
-    object,
-])
+@pytest.mark.parametrize(
+    "params",
+    [
+        "NOT A LIST OR TUPPLE",
+        {"invalid", "params"},
+        object,
+    ],
```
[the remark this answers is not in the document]
rushton: reverted

integration_tests/test_dbapi.py
```diff
@@ -82,26 +84,52 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+            (2, 'two', 'b'),
+            (3, 'three', 'c'),
+            (4, 'four', 'd'),
+            (5, 'five', 'e'),
+            (6, 'six''', 'f')
+        ) x (id, name, letter)
+        where name = ? OR name = ?
+        order by id
+        """,
+        params=("one","six'"),
```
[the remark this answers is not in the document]
rushton: updated with shorter query

presto/dbapi.py
rushton: agree, I tagged https://github.com/prestosql/presto-python-client/issues/45 as the issue for this. I will try to follow up with another MR to address this.

presto/dbapi.py
rushton: moved

presto/dbapi.py
rushton: tried to address this in #51

findepi (approved):
thanks!

findepi:
Merged, thanks!

PR #51, opened 2020-10-15
 Handle python types for query parameters

rushton:
Prior to this commit, the only python types supported as query parameters were str and int. In this commit the types are expanded to bool, float, bytes, datetime, list, dict, and UUID.

### DIFF
--- a/integration_tests/fixtures.py
+++ b/integration_tests/fixtures.py
@@ -29,7 +29,7 @@ from presto.exceptions import TimeoutError
 logger = presto.logging.get_logger(__name__)
 
 
-PRESTO_VERSION = os.environ.get("PRESTO_VERSION") or "329"
+PRESTO_VERSION = os.environ.get("PRESTO_VERSION") or "344"
 PRESTO_HOST = "127.0.0.1"
 PRESTO_PORT = 8080
 
--- a/integration_tests/test_dbapi.py
+++ b/integration_tests/test_dbapi.py
@@ -10,10 +10,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from __future__ import absolute_import, division, print_function
+from datetime import datetime
+import uuid
+import math
 
 import fixtures
 from fixtures import run_presto
 import pytest
+import pytz
 
 import presto
 from presto.exceptions import PrestoQueryError
@@ -87,22 +91,140 @@ def test_select_query_result_iteration_statement_params(presto_connection):
         params=(3,)  # expecting all the rows with id >= 3
     )
 
-def test_select_query_result_string_statement_params(presto_connection):
-    """
-    Tests string parameters in prepared statements
-    """
-    cur = presto_connection.cursor()
-    # test a simple string parameter and
-    # a parameter with a quote in it
-    cur.execute(
-        "select ?",
-        params=("six'",),
-    )
 
+def test_none_query_param(presto_connection):
+    cur = presto_connection.cursor()
+    cur.execute("SELECT ?", params=(None,))
     rows = cur.fetchall()
 
-    assert len(rows) == 1
+    assert rows[0][0] == None
+
+
+def test_string_query_param(presto_connection):
+    cur = presto_connection.cursor()
+
+    cur.execute("SELECT ?", params=("six'",))
+    rows = cur.fetchall()
+
     assert rows[0][0] == "six'"
+
+
+def test_datetime_query_param(presto_connection):
+    cur = presto_connection.cursor()
+
+    cur.execute(
+            "SELECT ?", 
+            params=(datetime(2020, 1, 1, 0, 0, 0),)
+            )
+    rows = cur.fetchall()
+
+    assert rows[0][0] == "2020-01-01 00:00:00.000"
+
+    cur.execute(
+            "SELECT ?", 
+            params=(datetime(2020, 1, 1, 0, 0, 0, tzinfo=pytz.utc),)
+            )
+    rows = cur.fetchall()
+
+    assert rows[0][0] == "2020-01-01 00:00:00.000 UTC"
+    assert cur.description[0][1] == "timestamp with time zone"
+
+
+def test_array_query_param(presto_connection):
+    cur = presto_connection.cursor()
+
+    cur.execute("SELECT ?", params=([1, 2, 3],))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == [1, 2, 3]
+
+    cur.execute(
+            "SELECT ?", 
+            params=([[1, 2, 3],[4,5,6]],))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == [[1, 2, 3],[4,5,6]]
+
+    cur.execute("SELECT TYPEOF(?)", params=([1, 2, 3],))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == "array(integer)"
+
+
+def test_dict_query_param(presto_connection):
+    cur = presto_connection.cursor()
+
+    cur.execute("SELECT ?", params=({"foo": "bar"},))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == {"foo": "bar"}
+
+    cur.execute("SELECT TYPEOF(?)", params=({"foo": "bar"},))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == "map(varchar(3), varchar(3))"
+
+
+def test_boolean_query_param(presto_connection):
+    cur = presto_connection.cursor()
+
+    cur.execute("SELECT ?", params=(True,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == True
+
+    cur.execute("SELECT ?", params=(False,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == False
+
+
+def test_float_query_param(presto_connection):
+    cur = presto_connection.cursor()
+    cur.execute("SELECT ?", params=(1.1,))
+    rows = cur.fetchall()
+
+    assert cur.description[0][1] == "double"
+    assert rows[0][0] == 1.1
+
+
+@pytest.mark.skip(reason="Nan currently not returning the correct python type for nan")
+def test_float_nan_query_param(presto_connection):
+    cur = presto_connection.cursor()
+    cur.execute("SELECT ?", params=(float("nan"),))
+    rows = cur.fetchall()
+
+    assert cur.description[0][1] == "double"
+    assert isinstance(rows[0][0], float)
+    assert math.isnan(rows[0][0])
+
+
+@pytest.mark.skip(reason="Nan currently not returning the correct python type fon inf")
+def test_float_inf_query_param(presto_connection):
+    cur.execute("SELECT ?", params=(float("inf"),))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == float("inf")
+
+    cur.execute("SELECT ?", params=(-float("-inf"),))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == float("-inf")
+
+
+def test_int_query_param(presto_connection):
+    cur = presto_connection.cursor()
+    cur.execute("SELECT ?", params=(3,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == 3
+    assert cur.description[0][1] == "integer"
+
+    cur.execute("SELECT ?", params=(9223372036854775807,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == 9223372036854775807
+    assert cur.description[0][1] == "bigint"
 
 
 @pytest.mark.parametrize('params', [
--- a/presto/dbapi.py
+++ b/presto/dbapi.py
@@ -30,8 +30,10 @@ except ImportError:
     from urllib import urlencode
 
 import copy
-import datetime
 import uuid
+import datetime
+import re
+import math
 
 from presto import constants
 import presto.exceptions
@@ -298,13 +300,52 @@ class Cursor(object):
         Formats parameters to be passed in an
         EXECUTE statement.
         """
+        if param is None:
+            return "NULL"
+
+        if isinstance(param, bool):
+            return "true" if param else "false"
+
+        if isinstance(param, int):
+            # TODO represent numbers exceeding 64-bit (BIGINT) as DECIMAL
+            return "%d" % param
+
+        if isinstance(param, float):
+            if param == float("+inf"):
+                return "infinity()"
+            if param == float("-inf"):
+                return "-infinity()"
+            if math.isnan(param):
+                return "nan()"
+            return "DOUBLE '%s'" % param
+
         if isinstance(param, str):
             return ("'%s'" % param.replace("'", "''"))
-        else:
-            # TODO: param could have many other types which should
-            #       be covered including unicode, maps, lists, etc.
-            #       https://github.com/prestosql/presto-python-client/issues/45
-            return str(param)
+
+        if isinstance(param, bytes):
+            return "X'%s'" % param.hex()
+
+        if isinstance(param, datetime.datetime):
+            datetime_str = param.strftime("%Y-%m-%d %H:%M:%S.%f %Z")
+            # strip trailing whitespace if param has no zone
+            datetime_str = datetime_str.rstrip(" ")
+            return "TIMESTAMP '%s'" % datetime_str
+
+        if isinstance(param, list):
+            return "ARRAY[%s]" % ','.join(map(self._format_prepared_param, param))
+
+        if isinstance(param, dict):
+            keys = list(param.keys())
+            values = [param[key] for key in keys]
+            return "MAP(%s, %s)" % (
+                self._format_prepared_param(keys),
+                self._format_prepared_param(values)
+            )
+
+        if isinstance(param, uuid.UUID):
+            return "UUID '%s'" % param
+
+        raise presto.exceptions.NotSupportedError("Query parameter of type '%s' is not supported." % type(param))
 
 
     def _deallocate_prepare_statement(self, added_prepare_header, statement_name):
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ kerberos_require = ["requests_kerberos"]
 
 all_require = [kerberos_require]
 
-tests_require = all_require + ["httpretty", "pytest", "pytest-runner", "mock"]
+tests_require = all_require + ["httpretty", "pytest", "pytest-runner", "mock", "pytz"]
 
 py27_require = ["ipaddress", "typing"]
 


rushton:
does not yet work for older versions of Presto

rushton:
the fix fails with presto version >=329.

presto/dbapi.py
```diff
@@ -281,17 +282,44 @@ def _prepare_statement(self, operation, statement_name):
...
+            return "true" if param else "false"
+        if isinstance(param, float):
+            return "DOUBLE '%s'" % param
+        if isinstance(param, str):
+            return ("'%s'" % param.replace("'", "''"))
+        if isinstance(param, bytes):
+            return "X'%s'" % param.hex()
+        if isinstance(param, datetime.datetime):
+            datetime_str = param.strftime("%Y-%m-%d %H:%M:%S.%f %Z").rstrip(" ")
+            return "TIMESTAMP '%s'" % re.sub(r'\.([0-9]{3})([0-9]{3})', r'.\1', datetime_str)
```
rushton, line 312: prior to presto 335 the timestamp format only supports 1/100th of a second in the timestamp format, thus the last 3 digits of the subseconds are truncated (effectively dividing by 1000).

presto/dbapi.py
```diff
@@ -281,17 +282,46 @@ def _prepare_statement(self, operation, statement_name):
...
+        if isinstance(param, list):
+            return "ARRAY[%s]" % ','.join(map(self._format_prepared_param, param))
+        if isinstance(param, dict):
+            keys = self._format_prepared_param(list(param.keys()))
+            values = self._format_prepared_param(list(param.values()))
+            return "MAP(%s, %s)" % (keys, values)
+        if isinstance(param, uuid.UUID):
+            return "UUID '%s'" % str(param)
+
+        return str(param)
```
rushton, line 322: thought an exception might be appropriate here, rather than presto spitting out a random syntax error. Thoughts?

findepi: Agreed.

presto/dbapi.py
findepi: we can have a legacy mode for dealing with older versions, but we should be correct by default (no implicit truncation nor rounding) for the new versions

doing truncation via regular expression is very clever, but i am thinking we could perhaps round using some methods on `param`, or by providing desired output format to `strftime` directly.

integration_tests/fixtures.py
```diff
@@ -29,7 +29,7 @@
 logger = presto.logging.get_logger(__name__)
 
 
-PRESTO_VERSION = os.environ.get("PRESTO_VERSION") or "329"
+PRESTO_VERSION = os.environ.get("PRESTO_VERSION") or "330"
```
findepi, line 32: 344 is the latest

integration_tests/test_dbapi.py
```diff
@@ -87,22 +90,45 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
     Tests string parameters in prepared statements
     """
     cur = presto_connection.cursor()
+    expected_uuid = uuid.uuid1()
     # test a simple string parameter and
     # a parameter with a quote in it
     cur.execute(
-        "select ?",
-        params=("six'",),
+        "SELECT ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?",
```
findepi, line 102: This is efficient (one query, many assertions), but this does not "scale" when you add more test cases (the list of question marks will grow and grow), or more assertions -- for example we will want to verify not only how `SELECT ?` looks like back on our side, but also what Presto engine things of the type of `?` (eg `SELECT typeof(?)`)

Thus, I would rather go with separate query + assertions for every value tested.
This could be similar to what we do with JDBC tests.
(I am linking an example to a read path test, but i am sure you get what i mean)
https://github.com/prestosql/presto/blob/cbcba450481feb3d7b92fb862dd6692b6a921f7e/presto-jdbc/src/test/java/io/prestosql/jdbc/BaseTestJdbcResultSet.java#L188-L202

presto/dbapi.py
```diff
@@ -298,13 +299,27 @@ def _format_prepared_param(self, param):
         Formats parameters to be passed in an
         EXECUTE statement.
         """
+        if isinstance(param, bool):
```
findepi, line 302: add
```
if param is None:
    return "NULL"
```

as the first case

presto/dbapi.py
```diff
@@ -298,13 +299,27 @@ def _format_prepared_param(self, param):
...
+            return "DOUBLE '%s'" % param
         if isinstance(param, str):
             return ("'%s'" % param.replace("'", "''"))
-        else:
-            # TODO: param could have many other types which should
-            #       be covered including unicode, maps, lists, etc.
-            #       https://github.com/prestosql/presto-python-client/issues/45
-            return str(param)
+        if isinstance(param, bytes):
+            return "X'%s'" % param.hex()
```
findepi, line 309: can you please add support for unicode as well?
for now this can be as simple as the `str` case.
in the future we can go with `U&'...'` unicode escapes, but maybe we do not actually need to

presto/dbapi.py
```diff
@@ -298,13 +299,27 @@ def _format_prepared_param(self, param):
...
+        if isinstance(param, datetime.datetime):
+            datetime_str = param.strftime("%Y-%m-%d %H:%M:%S.%f %Z").rstrip(" ")
+            return "TIMESTAMP '%s'" % re.sub(r'\.([0-9]{3})([0-9]{3})', r'.\1', datetime_str)
+        if isinstance(param, list):
+            return "ARRAY[%s]" % ','.join(map(self._format_prepared_param, param))
+        if isinstance(param, dict):
+            keys = self._format_prepared_param(list(param.keys()))
+            values = self._format_prepared_param(list(param.values()))
```
findepi, line 317: Can `param.keys()` and `param.values()` ordering go out of sync? can it depend on `dict` subclass?
if possible, this could be a safe guard:
```
entries = list(param. items())
keys = self._format_prepared_param([k for (k, v) in entries])
values = self._format_prepared_param([v for (k, v) in entries])
```

integration_tests/test_dbapi.py
```diff
@@ -87,22 +90,45 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+        params=(
+            "six'",
+            datetime(2020,1,1,0,0,0),
+            datetime(2020,1,1,0,0,0,tzinfo=pytz.utc),
```
findepi, line 106: that's good test case to have

how would i construct "timestamp with time zone" equivalent with just standard library?
should we test for this as well?

integration_tests/test_dbapi.py
```diff
@@ -87,22 +90,45 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+    expected_uuid = uuid.uuid1()
     # test a simple string parameter and
     # a parameter with a quote in it
     cur.execute(
-        "select ?",
-        params=("six'",),
+        "SELECT ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?",
+        params=(
+            "six'",
+            datetime(2020,1,1,0,0,0),
```
findepi, line 105: nit: while we do not have a formal code style estabilished here, i personally would prefer a space after a comma

presto/dbapi.py
rushton: done, thanks!

presto/dbapi.py
rushton: added, used `NotSupportedError` as the exception. Let me know if that is the correct one to use  here.

integration_tests/fixtures.py
rushton: updated

presto/dbapi.py
rushton: Correct, they can be out of order, good catch.  I ended up using the keys to build the list of values.

integration_tests/test_dbapi.py
rushton: This does output a `timestamp with time zone` type from presto. I've added an assertion to verify that is the case.

integration_tests/test_dbapi.py
rushton: updated, thanks!

presto/dbapi.py
rushton: just to clarify, should we only support the later versions here and forego truncating the subseconds? or round depending on how the user constructed the `datetime`?

integration_tests/test_dbapi.py
rushton: broke out into multiple tests, thanks for the tip!

presto/dbapi.py
rushton: did some reading, in python 3 all `str` objects are unicode, thus should we always apply the `U&` prefix? also, any suggestions for doing unicode detection that is compatible with both  python2 and python3?

integration_tests/test_dbapi.py
```diff
@@ -87,22 +90,97 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+        params=([[1, 2, 3],[4,5,6]],))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == [[1, 2, 3],[4,5,6]]
+
+
+def test_dict_query_params(presto_connection):
+    cur = presto_connection.cursor()
+
+    cur.execute("SELECT ?", params=({"foo": "bar"},))
```
findepi, line 150: please add 
```
# TODO validate the type on the Presto side (e.g. `typeof(?)`)
```

integration_tests/test_dbapi.py
```diff
@@ -87,22 +90,97 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
-    assert len(rows) == 1
-    assert rows[0][0] == "six'"
+    assert rows[0][0] == "2020-01-01 00:00:00.000 UTC"
+    assert cur.description[0][1] == "timestamp with time zone"
+
+
+def test_array_query_params(presto_connection):
+    cur = presto_connection.cursor()
+
+    cur.execute("SELECT ?", params=([1, 2, 3],))
```
findepi, line 134: please add 
```
# TODO validate the type on the Presto side (e.g. `typeof(?)`)
```

integration_tests/test_dbapi.py
```diff
@@ -87,22 +90,97 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+
+    cur.execute("SELECT ?", params=(False,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == False
+
+
+def test_float_query_params(presto_connection):
+    cur = presto_connection.cursor()
+    cur.execute("SELECT ?", params=(1.1,))
```
findepi, line 172: add test cases with NaN, +inf and -inf

integration_tests/test_dbapi.py
```diff
@@ -87,22 +90,97 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+    rows = cur.fetchall()
+
+    assert rows[0][0] == 1.1
+
+
+def test_int_query_params(presto_connection):
+    cur = presto_connection.cursor()
+    cur.execute("SELECT ?", params=(3,))
```
findepi, line 180: add a test case with negative number

add a test case with number not fitting in 32-bit in, eg Long.MAX_VALUE ( 9,223,372,036,854,775,807 )

presto/dbapi.py
```diff
@@ -298,13 +299,35 @@ def _format_prepared_param(self, param):
         Formats parameters to be passed in an
         EXECUTE statement.
         """
+        if param is None:
+            return "NULL"
+        if isinstance(param, bool):
+            return "true" if param else "false"
+        if isinstance(param, int):
+            return "%d" % param
```
findepi, line 307: add
```
# TODO represent numbers exceeding 64-bit (BIGINT) as DECIMAL
```

presto/dbapi.py
findepi:
```
isinstance(param, type(u''))
```

or, perhaps more elegant way
```
try:
    from builtins import unicode
except ImportError:
    unicode = str
```

presto/dbapi.py
findepi: I do not see a need to support old versions here.

we can always revisit if there is enough demand

presto/dbapi.py
```diff
@@ -298,13 +299,35 @@ def _format_prepared_param(self, param):
...
             return ("'%s'" % param.replace("'", "''"))
-        else:
-            # TODO: param could have many other types which should
-            #       be covered including unicode, maps, lists, etc.
-            #       https://github.com/prestosql/presto-python-client/issues/45
-            return str(param)
+        if isinstance(param, bytes):
+            return "X'%s'" % param.hex()
+        if isinstance(param, datetime.datetime):
+            datetime_str = param.strftime("%Y-%m-%d %H:%M:%S.%f %Z").rstrip(" ")
```
findepi, line 315:
```
datetime_str = param.strftime("%Y-%m-%d %H:%M:%S.%f %Z")
# strip trailing whitespace if param has no zone
datetime_str = datetime_str.rstrip(" ")
```

presto/dbapi.py
```diff
@@ -298,13 +299,35 @@ def _format_prepared_param(self, param):
...
+        if isinstance(param, dict):
+            keys = list(param.keys())
+            values = [param[key] for key in keys]
+
+            return "MAP(%s, %s)" % (
+                self._format_prepared_param(keys),
+                self._format_prepared_param(values)
+            )
+        if isinstance(param, uuid.UUID):
+            return "UUID '%s'" % str(param)
```
findepi, line 328: is `str` redundant here?

findepi (approved):
thanks, minor comments

presto/dbapi.py
findepi: > should we always apply the U& prefix

let's not use U& for now. just handling `unicode` _type_ is enough for this PR.

as a follow up we can add some unicode test cases (with ascii, non-ascii, non-BMP, etc.)

integration_tests/test_dbapi.py
rushton: added the test case for this

integration_tests/test_dbapi.py
rushton: added the test case for this

integration_tests/test_dbapi.py
rushton: added tests and implementation for these cases, but they all fail on the results side. Using `infinity()` and `nan()` functions in presto convert to a `str` in the python client. I'm not sure where this is happening in the code, but willing to fix it if you have some insight. For now I've marked the tests as skip with an explanation.

integration_tests/test_dbapi.py
rushton: added

presto/dbapi.py
rushton: added

presto/dbapi.py
rushton: done

presto/dbapi.py
rushton: removed truncation of subseconds

presto/dbapi.py
rushton: yes, fixed

integration_tests/test_dbapi.py
findepi: > For now I've marked the tests as skip with an explanation.

Thanks, please file an issue to track this and let's thus unblock this PR

integration_tests/test_dbapi.py
```diff
@@ -87,24 +91,141 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+    assert rows[0][0] == "2020-01-01 00:00:00.000"
+
+    cur.execute(
+            "SELECT ?", 
+            params=(datetime(2020, 1, 1, 0, 0, 0, tzinfo=pytz.utc),)
+            )
+    rows = cur.fetchall()
+
+    assert rows[0][0] == "2020-01-01 00:00:00.000 UTC"
+    assert cur.description[0][1] == "timestamp with time zone"
```
findepi, line 130: Curious, why `rows` is `[0][0]` and description is `[0][1]`?

integration_tests/test_dbapi.py
```diff
@@ -87,24 +91,141 @@ def test_select_query_result_iteration_statement_params(presto_connection):
...
+
+    assert rows[0][0] == False
+
+
+def test_float_query_param(presto_connection):
+    cur = presto_connection.cursor()
+    cur.execute("SELECT ?", params=(1.1,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == 1.1
```
findepi, line 187: let's have `cur.description`  assertion for floats too 
(eg to "document" floats end up as being `double` in Presto)

presto/dbapi.py
```diff
@@ -298,13 +300,45 @@ def _format_prepared_param(self, param):
...
+            # TODO represent numbers exceeding 64-bit (BIGINT) as DECIMAL
+            return "%d" % param
+        if isinstance(param, float) and param == float("+inf"):
+            return "infinity()"
+        if isinstance(param, float) and param == float("-inf"):
+            return "-infinity()"
+        if isinstance(param, float) and math.isnan(param):
+            return "nan()"
+        if isinstance(param, float):
+            return "DOUBLE '%s'" % param
```
findepi, line 317: let's group this by type:
```
if isinstance(param, float):
    if param == float("+inf"):
       ....
    ....
    return "DOUBLE '%s'" % param
```

presto/dbapi.py
findepi: OK, let's keep unicode out of this PR. https://github.com/prestosql/presto-python-client/issues/53

presto/dbapi.py
```diff
@@ -298,13 +300,45 @@ def _format_prepared_param(self, param):
...
             return ("'%s'" % param.replace("'", "''"))
-        else:
-            # TODO: param could have many other types which should
-            #       be covered including unicode, maps, lists, etc.
-            #       https://github.com/prestosql/presto-python-client/issues/45
-            return str(param)
+        if isinstance(param, bytes):
+            return "X'%s'" % param.hex()
+        if isinstance(param, datetime.datetime):
+            datetime_str = param.strftime("%Y-%m-%d %H:%M:%S.%f %Z").rstrip(" ")
```
findepi, line 323: the `rstrip` is now on the next line

```suggestion
            datetime_str = param.strftime("%Y-%m-%d %H:%M:%S.%f %Z")
```

presto/dbapi.py
```diff
@@ -298,13 +300,45 @@ def _format_prepared_param(self, param):
...
+            # strip trailing whitespace if param has no zone
+            datetime_str = datetime_str.rstrip(" ")
+
```
findepi, line 326: remove this line

(i would rather keep empty line after each `return` than within the "then" block)

presto/dbapi.py
```diff
@@ -298,13 +300,45 @@ def _format_prepared_param(self, param):
...
+            return "TIMESTAMP '%s'" % datetime_str
+        if isinstance(param, list):
+            return "ARRAY[%s]" % ','.join(map(self._format_prepared_param, param))
+        if isinstance(param, dict):
+            keys = list(param.keys())
+            values = [param[key] for key in keys]
+
```
findepi, line 333: remove this line

findepi (approved):

integration_tests/test_dbapi.py
rushton: `description` is a list of tuples of the form: `[("<output_col_name>", "<output_col_type>"),...]`

findepi: obviously... i think i now understand the feeling you had seeing my question 😉

findepi (approved):

findepi:
many thanks!

integration_tests/test_dbapi.py
rushton: created: https://github.com/prestosql/presto-python-client/issues/54 thanks!

PR #55, opened 2020-10-22
Fix return type of NaN and Infinity values from presto API

### LINKED ISSUE #54: Infinity and nan functions return strings instead of floats
## Problem

the values returned from infinity and nan functions in presto get cast to strings in python. 

## Reproduction

```
cursor.execute("SELECT infinity(), nan()")
rows = cursor.fetchall()

type(rows[0][0]) # => str
type(rows[0][1]) # => str
```

these should return the python `float("inf")` and `float("nan")` respectively.

rushton:
Presto returns NaN and Infinity values as strings in the API response.
The python client doesn't do any work to coerece the values, thus their
python types differ from the type intended by presto. This commit fixes
that by detecting when these special values get returned and
transforming them into python's special values for NaN and infinity.

Fixes #54

### DIFF
--- a/integration_tests/test_dbapi.py
+++ b/integration_tests/test_dbapi.py
@@ -188,7 +188,6 @@ def test_float_query_param(presto_connection):
     assert rows[0][0] == 1.1
 
 
-@pytest.mark.skip(reason="Nan currently not returning the correct python type for nan")
 def test_float_nan_query_param(presto_connection):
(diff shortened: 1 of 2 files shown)


rushton:
Naive and inefficient approach here, could be sped up by caching indexes of fields which need to be checked. Also doesn't account for NaN/Infinity values in Maps/Arrays

PR #64, opened 2021-01-07
Complete transition to Trino

realknorke:
replaced almost every occurrence of "presto" with "trino".
README.md migrated, too -- except the Presto book link
changed default version of Trino for integration tests to 351

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,21 @@
-[![Build Status](https://github.com/prestosql/presto-python-client/workflows/ci/badge.svg)](https://github.com/prestosql/presto-python-client/actions?query=workflow%3Aci+event%3Apush+branch%3Amaster)
-[![Presto Slack](https://img.shields.io/static/v1?logo=slack&logoColor=959DA5&label=Slack&labelColor=333a41&message=join%20conversation&color=3AC358)](https://prestosql.io/slack.html)
+[![Build Status](https://github.com/trinodb/trino-python-client/workflows/ci/badge.svg)](https://github.com/trinodb/trino-python-client/actions?query=workflow%3Aci+event%3Apush+branch%3Amaster)
+[![Trino Slack](https://img.shields.io/static/v1?logo=slack&logoColor=959DA5&label=Slack&labelColor=333a41&message=join%20conversation&color=3AC358)](https://trino.io/slack.html)
 [![Presto: The Definitive Guide book download](https://img.shields.io/badge/Presto%3A%20The%20Definitive%20Guide-download-brightgreen)](https://www.starburstdata.com/oreilly-presto-guide-download/)
 
 # Introduction
 
-This package provides a client interface to query [Presto](https://prestosql.io/)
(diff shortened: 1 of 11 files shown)


realknorke:
I used the code to execute some queries on our Trino351 cluster. Works.

README.md
```diff
@@ -1,21 +1,21 @@
...
 
-This package provides a client interface to query [Presto](https://prestosql.io/)
+This package provides a client interface to query [Trino](https://trino.io/)
 a distributed SQL engine. It supports Python 2.7, 3.5, 3.6, and pypy.
 
 # Installation
 
 ```
-$ pip install presto-client
+$ pip install trino-client
```
hashhar, line 13:
```suggestion
$ pip install trino
```

https://pypi.org/project/trino/

integration_tests/test_dbapi.py
```diff
@@ -355,8 +355,8 @@ def test_transaction_rollback(presto_connection_with_transaction):
         assert len(rows) == 1000
 
 
-def test_transaction_multiple(presto_connection_with_transaction):
-    with presto_connection_with_transaction as connection:
+def test_transaction_multiple(trino_connection_with_transaction):
+    with trino_connection_with_transaction as connection:
```
hashhar, line 359: Split this change into two commits. One with just the rename and the next one with the refactoring.

Thanks for catching this.

trino/constants.py
```diff
@@ -17,7 +17,7 @@
 
 
 DEFAULT_PORT = 8080
-DEFAULT_SOURCE = "presto-python-client"
+DEFAULT_SOURCE = "Trino-python-client"
```
hashhar, line 20: Is the case change intentional?

hashhar (changes requested):
Thanks for picking this up. ❤️ 

There are some non-rename changes in the same commit. I'd like if you could extract them into a separate commit at the end.

Also, some changes requested.

trino/constants.py
realknorke: fixed. sharp eyes! :)
thanks

integration_tests/test_dbapi.py
hashhar: Nevermind about this, I misread the diff.

hashhar (approved):
Looks good to me.

Would love a second pair of eyes - specifically the HTTP header changes.

findepi:
Merged, thanks!

callado4:
Is there a way to configure the header values? It seems that the sever I'm using is returning headers with the presto value, such as `'X-Presto-Added-Prepare': 'st_...`

ebyhr:
@callado4 I would recommend upgrading your cluster. Trino Python client doesn't provide the way to configure those values.

hashhar:
@callado4 While you upgrade the cluster you can use `pip install presto-client==0.302.0` (the last version that works with PrestoSQL) - note that it doesn't have any of the fixes made to trino-python-client over the last year. I also highly recommend to upgrade your cluster.

PR #65, opened 2021-01-08
Rename Header

71nn:
From Presto To Trino


findepi:
There is also @realknorke 's #64 which does this and more.

findepi:
We ended up merging the earlier attempt (https://github.com/trinodb/trino-python-client/pull/64), so let me close this one.

@71nn thank you for your contribution & i looking forward for more in the future!

Issue #43, opened 2020-09-09
Make client module private


findepi:
rename client to _client.

relates to https://github.com/prestosql/presto-python-client/pull/41#issuecomment-689613917

cc @hashhar

hashhar:
Acknowledged. Will send a PR once #41 is merged to avoid to rebase/conflicts.

joshthoward:
@hashhar If we update the public facing API, can we import `connection` as a peer to the `dbapi` module? This matches what is done SQLAlchemy and PyODBC and should be able to be done in a backward compatible way.

findepi:
This is free to take.

PR #68, opened 2021-01-14
Added flake8 test and updated affected code

joshthoward:
Added flake8 to the GitHub CI tests and then updated the code to pass flake8 test. Some unused imports were removed.

The max line length is set to 120 and the W503 rule is ignored which is PEP 8 compliant.

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -34,14 +34,6 @@ TRINO_HOST = "127.0.0.1"
 TRINO_PORT = 8080
 
 
-def is_process_alive(pid):
-    try:
-        os.kill(pid, 0)
-    except ProcessLookupError:
-        return False
-    return True
-
-
 def get_local_port():
     with closing(socket.socket()) as s:
         s.bind(("localhost", 0))
--- a/integration_tests/test_dbapi.py
+++ b/integration_tests/test_dbapi.py
@@ -47,7 +47,7 @@ def trino_connection_with_transaction(run_trino):
 
 def test_select_query(trino_connection):
     cur = trino_connection.cursor()
-    cur.execute("select * from system.runtime.nodes")
+    cur.execute("SELECT * FROM system.runtime.nodes")
     rows = cur.fetchall()
     assert len(rows) > 0
     row = rows[0]
@@ -62,11 +62,11 @@ def test_select_query(trino_connection):
 
 def test_select_query_result_iteration(trino_connection):
     cur0 = trino_connection.cursor()
-    cur0.execute("select custkey from tpch.sf1.customer LIMIT 10")
+    cur0.execute("SELECT custkey FROM tpch.sf1.customer LIMIT 10")
     rows0 = cur0.genall()
 
     cur1 = trino_connection.cursor()
-    cur1.execute("select custkey from tpch.sf1.customer LIMIT 10")
+    cur1.execute("SELECT custkey FROM tpch.sf1.customer LIMIT 10")
     rows1 = cur1.fetchall()
 
     assert len(list(rows0)) == len(rows1)
@@ -76,7 +76,7 @@ def test_select_query_result_iteration_statement_params(trino_connection):
     cur = trino_connection.cursor()
     cur.execute(
         """
-        select * from (
+        SELECT * FROM (
             values
             (1, 'one', 'a'),
             (2, 'two', 'b'),
@@ -84,7 +84,7 @@ def test_select_query_result_iteration_statement_params(trino_connection):
             (4, 'four', 'd'),
             (5, 'five', 'e')
         ) x (id, name, letter)
-        where id >= ?
+        WHERE id >= ?
         """,
         params=(3,)  # expecting all the rows with id >= 3
     )
@@ -227,18 +227,18 @@ def test_int_query_param(trino_connection):
 def test_select_query_invalid_params(trino_connection, params):
     cur = trino_connection.cursor()
     with pytest.raises(AssertionError):
-        cur.execute('select ?', params=params)
+        cur.execute('SELECT ?', params=params)
 
 
 def test_select_cursor_iteration(trino_connection):
     cur0 = trino_connection.cursor()
-    cur0.execute("select nationkey from tpch.sf1.nation")
+    cur0.execute("SELECT nationkey FROM tpch.sf1.nation")
     rows0 = []
     for row in cur0:
         rows0.append(row)
 
     cur1 = trino_connection.cursor()
-    cur1.execute("select nationkey from tpch.sf1.nation")
+    cur1.execute("SELECT nationkey FROM tpch.sf1.nation")
     rows1 = cur1.fetchall()
 
     assert len(rows0) == len(rows1)
@@ -247,7 +247,7 @@ def test_select_cursor_iteration(trino_connection):
 
 def test_select_query_no_result(trino_connection):
     cur = trino_connection.cursor()
-    cur.execute("select * from system.runtime.nodes where false")
+    cur.execute("SELECT * FROM system.runtime.nodes WHERE false")
     rows = cur.fetchall()
     assert len(rows) == 0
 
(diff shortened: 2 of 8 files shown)


joshthoward:
CLA has been sent via email.

integration_tests/test_dbapi.py
```diff
@@ -171,12 +162,12 @@ def test_boolean_query_param(trino_connection):
     cur.execute("SELECT ?", params=(True,))
     rows = cur.fetchall()
 
-    assert rows[0][0] == True
+    assert rows[0][0]
```
hashhar, line 165: I wouldn't consider this equivalent. Can lead to silent bugs when copying tests.

This change is only safe because you know that the type is a boolean. Truthiness is different for other types.

Same at L170.

integration_tests/test_dbapi.py
```diff
@@ -201,6 +192,7 @@ def test_float_nan_query_param(trino_connection):
 
 @pytest.mark.skip(reason="Nan currently not returning the correct python type fon inf")
 def test_float_inf_query_param(trino_connection):
+    cur = trino_connection.cursor()
```
hashhar, line 195: Thanks. ❤️

integration_tests/test_dbapi.py
```diff
@@ -368,15 +360,15 @@ def test_transaction_multiple(trino_connection_with_transaction):
...
-    """
-    tests that an invalid query raises the correct exception
+    """Tests that an invalid query raises the correct exception
     """
     cur = trino_connection.cursor()
     with pytest.raises(TrinoQueryError):
         cur.execute(
             """
-            select * FRMO foo WHERE x = ?; 
+            select * FRMO foo WHERE x = ?;
```
hashhar, line 371: While we are at it, can you make SQL queries use uppercase keywords (or at-least be consistent within a single query) if it's not too much of an effort?

It should be a separate commit though.

setup.cfg
```diff
@@ -1,2 +1,6 @@
 [aliases]
 test=pytest
+
+[flake8]
+max-line-length = 120
+ignore = W503
```
hashhar, line 6: Please add a comment about why this rule is ignored (and what the rule says).

tests/test_client.py
```diff
@@ -214,8 +213,8 @@
                 "com.facebook.presto.sql.tree.Table.accept(Table.java:50)",
                 "com.facebook.presto.sql.tree.AstVisitor.process(AstVisitor.java:22)",
                 "com.facebook.presto.sql.analyzer.StatementAnalyzer.analyzeFrom(StatementAnalyzer.java:1413)",
-                "com.facebook.presto.sql.analyzer.StatementAnalyzer.visitQuerySpecification(StatementAnalyzer.java:670)",
-                "com.facebook.presto.sql.analyzer.StatementAnalyzer.visitQuerySpecification(StatementAnalyzer.java:166)",
+                "com.facebook.presto.sql.analyzer.StatementAnalyzer.visitQuerySpecification(StatementAnalyzer.java:670)",  # NOQA
```
hashhar, line 216: Is it possible to do this more cleanly? For the entire block at once?

trino/__init__.py
```diff
@@ -20,4 +20,6 @@
 from . import exceptions
 from . import logging
 
+__all__ = ['auth', 'dbapi', 'client', 'constants', 'exceptions', 'logging']
```
hashhar, line 23: I'm not sure if we consider all of these to be public API.

Most notably `client` isn't supposed to be public since it's an implementation detail of the driver. People can use it but it's not expected to be stable (which is what `__all__` indicates by convention.

`auth`, `dbapi`, `constants` and `exceptions` might be considered public.

trino/dbapi.py
```diff
@@ -23,16 +23,9 @@
 
 from typing import Any, List, Optional  # NOQA for mypy types
 
-try:
-    from urllib.parse import urlencode
-except ImportError:
-    # Python 2
-    from urllib import urlencode
```
hashhar, line 30: Was this unused?

This drops Python 2 support. The sad truth is that many people are still on 2.7.

integration_tests/conftest.py
```diff
@@ -34,14 +34,6 @@
 TRINO_PORT = 8080
 
 
-def is_process_alive(pid):
```
hashhar, line 37: It's surprising that this is unused. I'll check what changed to make this unused.

hashhar (commented):
A few comments.

Thanks for your contribution. ❤️

trino/dbapi.py
findepi: @hashhar per the diff, looks like unused import?

tests/test_client.py
joshthoward: Not to my knowledge. You can make this cleaner on line continuations, but not for arrays. Another option would be for the whole file, but I don't think that is better at this point. 

I am going to update this to only remove checks for line length though.

trino/__init__.py
joshthoward: `__all__` is updated here to indicate to flake8 that the unused imports are part of a public API. Otherwise, we might want to remove `client` and `logging` from `__init__.py`. What do you think?

trino/dbapi.py
joshthoward: It was unused.

integration_tests/test_dbapi.py
```diff
@@ -49,11 +47,11 @@ def trino_connection_with_transaction(run_trino):
 
 def test_select_query(trino_connection):
     cur = trino_connection.cursor()
-    cur.execute("select * from system.runtime.nodes")
+    cur.execute("SELECT * FROM system.runtime.nodes")
```
findepi, line 50: is this being checked too?

setup.cfg
```diff
@@ -1,2 +1,11 @@
 [aliases]
 test=pytest
+
+[flake8]
+max-line-length = 120
+# W503 raises a warning when there is a line break before a binary operator.
+# This is best practice according to PEP 8 and the rule should be ignored. 
+#
+# https:#www.flake8rules.com/rules/W503.html
+# https:#www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
```
findepi, line 10: # -> // in url

trino/__init__.py
hashhar: Good point. Let's leave it out of this PR - I'll follow up on this in https://github.com/trinodb/trino-python-client/issues/43

integration_tests/conftest.py
hashhar: This function was never called but has been here since the beginning. 🤷‍♂️ 

Let's remove it then. 👍

hashhar (approved):
LGTM 🚀

integration_tests/test_dbapi.py
joshthoward: It is not since it isn't "Python" code.

setup.cfg
joshthoward: Fixed.

joshthoward:
@findepi, @hashhar can either of you guys merge my PR? I think we should be good to close here.

setup.cfg
```diff
@@ -6,6 +6,6 @@ max-line-length = 120
 # W503 raises a warning when there is a line break before a binary operator.
 # This is best practice according to PEP 8 and the rule should be ignored. 
 #
-# https:#www.flake8rules.com/rules/W503.html
-# https:#www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
+# https://www.flake8rules.com/rules/W503.html
+# https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
```
findepi, line 10: This should be squashed into first commit.

findepi (approved):
LGTM!

One minor comment about commit boundaries.

setup.cfg
joshthoward: @findepi I was being lazy 😔 This is fixed now.

findepi (approved):

findepi:
Merged, thanks!

Issue #4, opened 2020-01-12
Make code Python3 only


ggreg:

joshthoward:
@ggreg completely agree with you...

For anyone else looking at this, the Python Software Foundation has [deprecated](https://www.python.org/doc/sunset-python-2/) Python 2.

findepi:
Well... it was deprecated long ago. 
What matters though are facts on the ground.
Sadly, I do not know of a good way to know whether there are users still on Python 2 and how many.

findepi:
@joshthoward maybe https://packaging.python.org/guides/analyzing-pypi-package-downloads/ could help, and https://packaging.python.org/guides/analyzing-pypi-package-downloads/#python-versions-over-time in particular?
the `trino` PyPI package is very new, but we could check for `presto-client` there.

joshthoward:
@findepi that would be an analytical way to validate that there are few Python 2 users, but it seems like a moot point given that the version is deprecated and also not supported by basic analytical libraries like Pandas or Numpy.

nineinchnick:
It's not only deprecated but actually not supported anymore.

findepi:
I agree with you that it would be great to drop Python 2 compatibility.
I am not convinced checking Python 2 deprecation status is sufficient for that. 

Java 8 is EOL'd and not supported (unless you pay $$$), but we don't feel comfortable dropping Java 8 support from JDBC or CLI tools.
Python 2 was "deprecated" like 10 years ago (or am i wrong here?).

to be clear: i am not afraid of forcing one straggler to upgrade. I just want that we make sure we don't leave some % of users in a bad situation.

yuokada:
I believe pyupgrade is useful for this issue.
https://github.com/asottile/pyupgrade

joshthoward:
@findepi Fair enough. The BigQuery link you posted was surprisingly easy to follow.  

Here are the results for the `presto-client` package over the last 6 months...

Query:
```
#standardSQL
SELECT
  REGEXP_EXTRACT(details.python, r"[0-9]+\.[0-9]+") AS python_version,
  COUNT(*) AS num_downloads,
FROM `the-psf.pypi.file_downloads`
WHERE
  file.project = 'presto-client'
  -- Only query the last 6 months of history
  AND DATE(timestamp)
    BETWEEN DATE_TRUNC(DATE_SUB(CURRENT_DATE(), INTERVAL 6 MONTH), MONTH)
    AND CURRENT_DATE()
[...]

PR #71, opened 2021-01-15
Drop python3.5 support

yuokada:
Python 3.5 was reached end-of-life in 2020. So we can drop python3.5 support.
- https://www.python.org/downloads/release/python-3510/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py35,py36,py37,pypy2
+envlist = py27,py36,py37,py38,py39,pypy2
 
 [testenv]
 extras = tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ jobs:
       matrix:
         python-version: [
           2.7,
-          3.5,
           3.6,
           3.7,
           3.8,


findepi (approved):

joshthoward:
Hey guys. What is the rationale behind dropping Python 3.5 before Python 2? 

From this [issue](https://github.com/trinodb/trino-python-client/issues/4), we get roughly 6x the number of 3.5 downloads compared to 2.7.

yuokada:
To be honest, I don't have the big reason to drop python3.5 support.
Since python3.5 was reached end-of-life, I sent the pr mechanically

One of the reason why dropping python3.5 support is a relationship to dependency libraries.
Pytest  has already dropped python3.5 support. So we might be unable to run tests on python 3.5 properly in the near future.
https://pypi.org/project/pytest/

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py35,py36,py37,pypy2
+envlist = py27,py36,py37,py38,py39,pypy2
```
electrum, line 2: This seems to also be adding support for newer versions. Can you create a separate PR for that? It should not be controversial (compared to dropping an old version) and we can merge it right away.

electrum:
Since this version is working today, I suggest we keep it for now. We can revisit the decision when there is a need, such as wanting to use new syntax or new dependency versions, or if something stops working. Then we can weigh the positive/negative, whereas dropping it now seems to only be a negative.

tox.ini
yuokada: Created
https://github.com/trinodb/trino-python-client/pull/72

yuokada:
@electrum I see. I will close the pr.

findepi:
> Hey guys. What is the rationale behind dropping Python 3.5 before Python 2?
> 
> From this [issue](https://github.com/trinodb/trino-python-client/issues/4), we get roughly 6x the number of 3.5 downloads compared to 2.7.

@joshthoward 
there are two reasons

- in my not so intensive python development in past 2-3 years i've witnessed problems present in 3.5 and fixed in 3.6. Enough for me to think there is little value supporting Python 3.5 (this is based on memory though)
  - on the other hand, python  2.7 is very solid (even though no longer supported)
- there is clean upgrade path for users from Python 3.5 to 3.6+
   - this is a big difference between 3.5 and 2.7
  
i did not know the download numbers when approving.
Knowing the numbers, I would still consider removing 3.5 support for the reasons outlined above, but i can be perhaps convinced to think differently.

PR #66, opened 2021-01-13
Add python3.8 & python3.9 to test targets

yuokada:

PR #72, opened 2021-01-17
Add python3.8 & python3.9 to envlist in tox.ini

yuokada:
follow up #66


ebyhr (approved):
Could you update `setup.py` too? Currently, version 3.7 is missing, so you can separate a commit likes "Add 3.7, 3.8 and 3.9 to classifiers in setup.py". 
https://pypi.org/pypi?%3Aaction=list_classifiers

PR #74, opened 2021-01-18
Pin runs-os to Ubuntu 18.04

yuokada:
The ubuntu-latest is assigned to ubuntu-20.04 from ubuntu-18.04 in the near future.

 Github action reports a warning message like below.
> Ubuntu-latest workflows will use Ubuntu-20.04 soon. For more details, see https://github.com/actions/virtual-environments/issues/1816

e.g: https://github.com/trinodb/trino-python-client/actions/runs/492701210

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     strategy:
       fail-fast: false
       matrix:


hashhar:
That is just a warning and I don't think anything should break once the OS is updated. We can pin in case something breaks after the upgrade.

findepi:
@yuokada thanks for taking care of the CI!

note that we shouldn't pin unless necessary.
Thus let's not apply any change here and act retroactively if any problems arise.

yuokada:
Got it. I will close.

PR #76, opened 2021-02-03
Added initial SQLAlchemy dialect for Trino

joshthoward:


joshthoward:
Once Python 2.7 and 3.5 support are dropped we should be able to merge.

findepi:
> joshthoward requested a review from findepi 2 hours ago


i understand this is blocked yet. So will appreciate ping once unblocked. Thanks for understanding.

joshthoward:
@findepi understandable, but it's ready for review none the less. I would appreciate your opinion on the changes to the client.py and dbapi.py files in order to achieve compatibility as well as the sqlalchemy connection string (tests/test_sqalchemy_url.py). Please take a look when you find the time.

trino/sqlalchemy.py
```diff
@@ -0,0 +1,166 @@
...
+            )
+            regex = r'Table\ \'.*{}\'\ does\ not\ exist'.format(re.escape(table_name))
+            if msg and re.search(regex, msg):
+                raise exc.NoSuchTableError(table_name)
+            else:
+                raise
+
+    def has_table(self, connection, table_name, schema=None):
+        """Determine whether table with schema exists"""
+        try:
```
dungdm93, line 134: Should use `SHOW TABLES 'table_name'` instead

trino/sqlalchemy.py
```diff
@@ -0,0 +1,166 @@
...
+            self._get_table_columns(connection, table_name, schema)
+            return True
+        except exc.NoSuchTableError:
+            return False
+
+    def get_indexes(self, connection, table_name, schema=None, **kw):
+        """Return columns which are partition keys"""
+        rows = self._get_table_columns(connection, table_name, schema)
```
dungdm93, line 142: Should get columns from `<table>$partitions`. See [Hive special tables](https://trino.io/docs/current/connector/hive.html#special-tables)

dungdm93 (changes requested):

setup.py
```diff
@@ -29,7 +29,9 @@
 
 kerberos_require = ["requests_kerberos"]
 
-all_require = [kerberos_require]
+sqlalchemy_require = ["SQLAlchemy==1.3.22"]
```
dungdm93, line 32: Should we use dependency range `~=` or `>=` instead of specific version like this?

PR #70, opened 2021-01-15
Drop python2 support

yuokada:
Resolve #4

PR #77, opened 2021-02-03
Drop python3.5 support

joshthoward:


joshthoward:
Depends on #70. Required for #76.

PR #79, opened 2021-02-20
Add support for JWT Authentication 

aakashnand:
The recent version of trino has the JWT authentication method so this PR adds support for JWT authentication in `trino-python-client`. I have updated documentation in `README` as well for the usage of JWT authentication.  Please let me know if any improvement is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,25 @@ cur.execute('SELECT * FROM system.runtime.nodes')
 rows = cur.fetchall()
 ```
 
+# JWT Authentication
+The `JWTAuthentication` can be used to connect to a JWT token authentication enabled Trino cluster:
+```python
+import trino
+JWT_TOKEN = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoiSm9obiBEb2UifQ.DjwRE2jZhren2Wt37t5hlVru6Myq4AhpGLiiefF69u8'
+conn = trino.dbapi.connect(
+    host='coordinator-url',
+    port=8443,
+    user='the-user',
+    catalog='the-catalog',
+    schema='the-schema',
+    http_scheme='https',
+    auth=trino.auth.JWTAuthentication(JWT_TOKEN),
+)
+cur = conn.cursor()
+cur.execute('SELECT * FROM system.runtime.nodes')
+rows = cur.fetchall()
+```
+
 # Transactions
 The client runs by default in *autocommit* mode. To enable transactions, set
 *isolation_level* to a value different than `IsolationLevel.AUTOCOMMIT`:
--- a/trino/auth.py
+++ b/trino/auth.py
@@ -18,6 +18,7 @@ import os
 
 from six import with_metaclass
 from typing import Any, Optional, Text  # NOQA
+from requests.auth import AuthBase
 
 
 class Authentication(with_metaclass(abc.ABCMeta)):  # type: ignore
@@ -120,6 +121,41 @@ class BasicAuthentication(Authentication):
             raise RuntimeError("unable to import requests.auth")
 
         http_session.auth = requests.auth.HTTPBasicAuth(self._username, self._password)
+        return http_session
+
+    def setup(self, trino_client):
+        self.set_client_session(trino_client.client_session)
+        self.set_http_session(trino_client.http_session)
+
+    def get_exceptions(self):
+        return ()
+
+    def handle_error(self, handle_error):
+        pass
+
+
+class _BearerAuth(AuthBase):
+    """
+    Custom implementation of Authentication class for bearer token
+    """
+    def __init__(self, token):
+        self.token = token
+
+    def __call__(self, r):
+        r.headers["Authorization"] = "Bearer " + self.token
+        return r
+
+
+class JWTAuthentication(Authentication):
+
+    def __init__(self, token):
+        self.token = token
+
+    def set_client_session(self, client_session):
+        pass
+
+    def set_http_session(self, http_session):
+        http_session.auth = _BearerAuth(self.token)
         return http_session
 
     def setup(self, trino_client):


trino/auth.py
```diff
@@ -131,3 +132,38 @@ def get_exceptions(self):
...
+
+class BearerAuth(AuthBase):
+    """
+    Custom implementation of Authentication class for bearer token
+    """
+    def __init__(self, token):
+        self.token = token
+
+    def __call__(self, r):
+        r.headers["authorization"] = "Bearer " + self.token
```
dain, line 145: I don't think it matters, but the header is normally capitalized https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

dain (commented):
The authorization header looks correct to me

trino/auth.py
```diff
@@ -131,3 +132,38 @@ def get_exceptions(self):
 
     def handle_error(self, handle_error):
         pass
+
+
+class BearerAuth(AuthBase):
```
joshthoward, line 137: Can we make this a private class in order to prevent direct use?

README.md
```diff
@@ -57,6 +57,23 @@ cur = conn.cursor()
...
+```python
+import trino
+conn=trino.dbapi.connect(
+    host='coordinator-url',
+    port=8443,
+    user='the-user',
+    catalog='the-catalog',
+    schema='the-schema',
+    http_schema='https',
+    auth=trino.auth.JWTAuthentication("jwt-token"),
```
joshthoward, line 71: Could we sub the "jwt-token" example string for an actual JWT token? e.g.  `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoiSm9obiBEb2UifQ.DjwRE2jZhren2Wt37t5hlVru6Myq4AhpGLiiefF69u8`? As is, it looks like a string based config at first glance. If not, maybe angle brackets?

joshthoward (changes requested):

aakashnand:
@dain @joshthoward thanks for the suggestions. I am waiting for my CLA to be updated otherwise the cla-bot will fail again.

trino/auth.py
aakashnand: Making class ‘private’ is enough or methods should also be made private? @joshthoward

aakashnand: Sorry my bad, understood

README.md
aakashnand: How about this @joshthoward 

```python
import trino
JWT_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoiSm9obiBEb2UifQ.DjwRE2jZhren2Wt37t5hlVru6Myq4AhpGLiiefF69u8 
conn=trino.dbapi.connect(
    host='coordinator-url',
    port=8443,
    user='the-user',
    catalog='the-catalog',
    schema='the-schema',
    http_schema='https',
    auth=trino.auth.JWTAuthentication(JWT_TOKEN),
)
cur = conn.cursor()
cur.execute('SELECT * FROM system.runtime.nodes')
rows = cur.fetchall()
```

joshthoward: That works (just quote it haha)

martint:
@cla-bot check

aakashnand:
@joshthoward @dain I have applied the changes as you suggested. Please take a moment to review again.

cc: @findepi

aakashnand:
I did not use separate commits for the suggestion as I thought it would be unnecessary to add new commits for minor changes. Please let me know if this is the wrong practice so that I can improve in future.

joshthoward:
LGTM

dain:
I don't know much about the python client, but the headers looks correct to me.  I would suggest squashing into a single commit.

README.md
```diff
@@ -57,6 +57,24 @@ cur = conn.cursor()
 cur.execute('SELECT * FROM system.runtime.nodes')
 rows = cur.fetchall()
 ```
+# JWT Authentication
+The `JWTAuthentication` can be used to connect to a JWT token authentication enabled Trino cluster:
+```python
+import trino
+JWT_TOKEN='eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoiSm9obiBEb2UifQ.DjwRE2jZhren2Wt37t5hlVru6Myq4AhpGLiiefF69u8'
+conn=trino.dbapi.connect(
```
ebyhr, line 65: nit: Add space 
```suggestion
JWT_TOKEN = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoiSm9obiBEb2UifQ.DjwRE2jZhren2Wt37t5hlVru6Myq4AhpGLiiefF69u8'
conn = trino.dbapi.connect(
```

README.md
```diff
@@ -57,6 +57,24 @@ cur = conn.cursor()
...
+    host='coordinator-url',
+    port=8443,
+    user='the-user',
+    catalog='the-catalog',
+    schema='the-schema',
+    http_schema='https',
```
ebyhr, line 71: Fix typo http_schema → http_scheme

README.md
```diff
@@ -57,6 +57,24 @@ cur = conn.cursor()
 cur.execute('SELECT * FROM system.runtime.nodes')
 rows = cur.fetchall()
 ```
+# JWT Authentication
```
ebyhr, line 60: nit: Add a new line between L59 and L60

ebyhr (commented):
Left only minor comments. Did you test this PR with your Trino cluster?

aakashnand:
@ebyhr Thanks for the detailed review. I will make those changes. I was already using JWT token authentication on my local trino cluster. It is working without any problem. One question though as @dain suggested me to squash commits into one, does it mean that every PR should have single commit only?

ebyhr:
@aakashnand No, it depends on the situation. We usually don't separate commits for code and docs change, so single commit in this PR is reasonable.

aakashnand:
@ebyhr Got it. I will make changes soon and will request a final review again from you. Thanks

trino/auth.py
```diff
@@ -131,3 +132,38 @@ def get_exceptions(self):
...
+
+    def setup(self, trino_client):
+        self.set_client_session(trino_client.client_session)
+        self.set_http_session(trino_client.http_session)
+
+    def get_exceptions(self):
+        return ()
+
+    def handle_error(self, handle_error):
+        pass
```
findepi, line 169: This is default behavior  of Authentication, so seems can be omitted here

trino/auth.py
```diff
@@ -18,6 +18,7 @@
 
 from six import with_metaclass
 from typing import Any, Optional, Text  # NOQA
+from requests.auth import AuthBase
```
findepi, line 21: That's out of scope of this PR, but we should define an __ALL__ here, so that `from trino.auth import *` doesn't pull things like `AuthBase`... or we say we don't care.

findepi (approved):

PR #80, opened 2021-03-24
eager-loading cursor.description

dungdm93:
Currently, `cursor.description` only loaded when data is fetched.
Screenshot from 2021-03-24 18-33-33
That make Trino's sqlalchemy dialect don't work. See [sqlalchemy's ResultProxy](https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_3_23/lib/sqlalchemy/engine/result.py#L793-L811)

Signed-off-by: Đặng Minh Dũng <dungdm93@live.com>

### DIFF
--- a/integration_tests/test_dbapi.py
+++ b/integration_tests/test_dbapi.py
@@ -9,14 +9,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from datetime import datetime
 import math
+from datetime import datetime
 
-from conftest import TRINO_VERSION
 import pytest
 import pytz
 
 import trino
+from conftest import TRINO_VERSION
 from trino.exceptions import TrinoQueryError
 from trino.transaction import IsolationLevel
 
@@ -371,3 +371,26 @@ def test_invalid_query_throws_correct_error(trino_connection):
             """,
             params=(3,),
         )
+
+
+def test_eager_loading_cursor_description(trino_connection):
+    description_expected = [
+        ('node_id', 'varchar', None, None, None, None, None),
+        ('http_uri', 'varchar', None, None, None, None, None),
+        ('node_version', 'varchar', None, None, None, None, None),
+        ('coordinator', 'boolean', None, None, None, None, None),
+        ('state', 'varchar', None, None, None, None, None),
+    ]
+    cur = trino_connection.cursor()
+    cur.execute('SELECT * FROM system.runtime.nodes')
+    description_before = cur.description
+
+    assert description_before is not None
+    assert len(description_before) == len(description_expected)
+    assert all([b == e] for b, e in zip(description_before, description_expected))
+
+    cur.fetchone()
+    description_after = cur.description
+    assert description_after is not None
+    assert len(description_after) == len(description_expected)
+    assert all([a == e] for a, e in zip(description_after, description_expected))
--- a/trino/client.py
+++ b/trino/client.py
@@ -37,17 +37,15 @@ import copy
 import os
 from typing import Any, Dict, List, Optional, Text, Tuple, Union  # NOQA for mypy types
 
-import trino.logging
 import requests
+
+import trino.logging
 from trino import constants, exceptions
 from trino.transaction import NO_TRANSACTION
 
-
 __all__ = ["TrinoQuery", "TrinoRequest"]
 
-
 logger = trino.logging.get_logger(__name__)
-
 
 MAX_ATTEMPTS = constants.DEFAULT_MAX_ATTEMPTS
 SOCKS_PROXY = os.environ.get("SOCKS_PROXY")
@@ -451,7 +449,7 @@ class TrinoResult(object):
         self._rows = None
 
         # Subsequent fetches from GET requests until next_uri is empty.
-        while not self._query.is_finished():
+        while not self._query.finished:
             rows = self._query.fetch()
             for row in rows:
                 self._rownumber += 1
@@ -487,6 +485,11 @@ class TrinoQuery(object):
 
     @property
     def columns(self):
+        if self.query_id:
+            while not self._columns and not self.finished and not self.cancelled:
+                # Columns don't return immediate after query is summited.
+                # Continue fetching data until columns are available and push fetched rows into buffer.
+                self._result._rows += self.fetch()
         return self._columns
 
     @property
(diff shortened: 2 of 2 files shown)


joshthoward:
@dungdm93 If it is helpful, please check out my [PR](https://github.com/trinodb/trino-python-client/pull/76/files#diff-fab1dd3ac4ba8d573ac4c1ddf02b844dd50aac9f942197db99f37c9e812d4d0bR206-R211). The changes in `dbapi.py` and `client.py` are to achieve this purpose.

dungdm93:
Thanks @joshthoward, I updated my PR after refer your code :+1:

integration_tests/test_dbapi.py
```diff
@@ -372,3 +373,13 @@ def test_invalid_query_throws_correct_error(trino_connection):
             """,
             params=(3,),
         )
+
+
+def test_eager_loading_cursor_description(trino_connection):
```
joshthoward, line 378: I would explicitly check that the cursor description is as expected before and after fetching a row.

trino/client.py
```diff
@@ -559,6 +562,10 @@ def is_finished(self):
         # type: () -> bool
         return self._finished
 
+    def is_cancelled(self):
```
joshthoward, line 565: Can you make this a property?

joshthoward (changes requested):
Minor changes. Looks good overall.

trino/client.py
dungdm93: Yeah, sure. And `is_finished` as wel?

joshthoward: In general we should align on using properties since that is what has been done for most attributes, but it's optional since you didn't touch it haha

dungdm93: I create `finished` property and now `is_finished()` marked as deprecated.

dungdm93:
Hello @joshthoward, @findepi, could you please review this one

joshthoward (approved):
LGTM

joshthoward (approved):

integration_tests/test_dbapi.py
```diff
@@ -371,3 +371,14 @@ def test_invalid_query_throws_correct_error(trino_connection):
...
+    cur = trino_connection.cursor()
+    cur.execute('SELECT * FROM system.runtime.nodes')
+    des_before = cur.description
```
hashhar, line 379: nit: avoid abbreviations. `before` and `after` are good alternatives (and still clear).

trino/client.py
```diff
@@ -541,7 +544,7 @@ def fetch(self):
     def cancel(self):
         # type: () -> None
         """Cancel the current query"""
-        if self.query_id is None or self.is_finished():
+        if self.query_id is None or self.finished:
             return
 
         self._cancelled = True
```
hashhar, line 550: Isn't this prone to incorrectness? The flag should be set AFTER we get a response from co-ordinator.

Pre-existing but worth noting.

trino/client.py
```diff
@@ -508,9 +511,9 @@ def execute(self, additional_http_headers=None):
         This is the first HTTP request sent to the coordinator.
         It sets the query_id and returns a Result object used to
         track the rows returned by the query. To fetch all rows,
-        call fetch() until is_finished is true.
+        call fetch() until finished.
```
hashhar, line 514: nit: `until finished is true`

integration_tests/test_dbapi.py
```diff
@@ -371,3 +371,14 @@ def test_invalid_query_throws_correct_error(trino_connection):
...
+    assert des_before is not None
+    cur.fetchone()
+    des_after = cur.description
```
hashhar, line 382: I would like if the check was a bit more strict and explicitly verified that the columns in the description match the ones requested from the query.

The current check will still pass if we add a bug in the part where cursor.description is set.

hashhar (commented):
commit msg refers to `edge-load` vs `eager load`.

Some comments - one about the test which I think might be important.

trino/client.py
joshthoward: You mean that `self.query_id` should be set before cancelling?

hashhar: No I'm talking about line 550.

We set the cancelled flag before actually cancelling the query. This can lead to abandoned queries in case the API call to co-ordinator fails for any reason or the client is interrupted.

hashhar (approved):

dungdm93:
Hello guys, can we merge this one?
@findepi, @joshthoward, @hashhar

hashhar (approved):
@dungdm93 Please squash the commits (so that test coverage is part of the implementation commit).

LGTM % squash. cc: @ebyhr @findepi

trino/client.py
hashhar: ping @joshthoward . Does what I am saying make sense or did I misunderstand the flow? In case this is an issue we can address this in a follow-up.

dungdm93:
> @dungdm93 Please squash the commits (so that test coverage is part of the implementation commit).
> 
> LGTM % squash. cc: @ebyhr @findepi

squashed

trino/client.py
```diff
@@ -555,9 +558,20 @@ def cancel(self):
         self._request.raise_response_error(response)
 
     def is_finished(self):
+        import warnings
+        warnings.warn("is_finished is deprecated, use finished instead", DeprecationWarning)
+        return self.finished
```
findepi, line 563: Why we change `is_finished` -> `finished` in this PR?
Why are you adding `cancelled` property? Is this some API we're filling?

trino/client.py
```diff
@@ -487,6 +487,9 @@ def __init__(
 
     @property
     def columns(self):
+        if self.query_id and not self.cancelled:
+            while not self._columns and not self.finished:
+                self._result._rows += self.fetch()
```
findepi, line 492: Add a comment why we're doing this

trino/client.py
```diff
@@ -487,6 +487,9 @@ def __init__(
 
     @property
     def columns(self):
+        if self.query_id and not self.cancelled:
+            while not self._columns and not self.finished:
```
findepi, line 491: should the loop include `not self.cancelled`?

trino/client.py
dungdm93: I believe `cancelled` should be public, not just internal use (like `is_finished()`).
And I also change from `is_finished` to `finished` to align with new `cancelled` properties. You could see previous discussion with @joshthoward

trino/client.py
dungdm93: I just thought we can't cancel the query while doing other operator (get cursor description in this case)

trino/client.py
dungdm93: Comment added

trino/client.py
joshthoward: That does make sense, but let's fix this in a different PR. This one has already been sitting for too long.

joshthoward (approved):

trino/client.py
hashhar: I agree @joshthoward . Just wanted to make sure we don't forget about this - I'll create an issue to track.

findepi:
Merged, thanks!

trino/client.py
joshthoward: @hashhar don't worry about it raising a PR.

Issue #13, opened 2020-02-16
SQLAlchemy compatibility


findepi:
The client should be compatible with SQLAlchemy and there should be tests for this.

esemeniuc:
Please add some examples or documentation for SQLAlchemy :)

esemeniuc:
This seems relevant: https://docs.sqlalchemy.org/en/13/core/connections.html#registering-new-dialects

andreas-eberle:
It looks like @dungdm93 is working on something like this: https://github.com/dungdm93/sqlalchemy-trino

dungdm93:
Thank you for interesting my project.

findepi:
@dungdm93 this is great!
do you have plans to contribute SQLAlchemy support to the main repo here?
cc @joshthoward

dungdm93:
@findepi I think it's better to keep it in a separated repo like [other dialects](https://docs.sqlalchemy.org/en/14/dialects/).  
However if possible, I also would like to transfer ownership of that repo to Trino community.

joshthoward:
@findepi, @dungdm93 IMO we should keep the SQL Alchemy Dialect w/ the DB API implementation. In the link provided tools like `pyhive` or `pydruid` do this. I've also noticed that there are a few differences in what SQL Alchemy expects from a DB API implementation vs. what we have (hence the need for bundling them in a single repo).

shawnzhu:
I had been using the [`pyHive[presto]`](https://github.com/dropbox/PyHive) as the DB-API and SQLAlchemy driver for presto for a while, it works very well, for instance:

```Python
import pandas as pd

sqlalchemy_presto_connection_url = "presto://username:password@presto-coordinator:8080/hive/schema-name"

df = pd.read_sql_query("SHOW tables", sqlalchemy_presto_connection_url)
```

I'm also interested in this feature since it will my workflow much better.

findepi:
@prakharcode there is a PR
 @joshthoward knows the current state better

dungdm93:
Hi @prakharcode, I start moving my code to this repo. First one is just a patch to make `cursor.description` eager-loading #80

HassanOuda:
[...]

PR #81, opened 2021-04-02
SQLAlchemy support

dungdm93:
See #13

### DIFF
--- a/tests/sqlalchemy/conftest.py
+++ b/tests/sqlalchemy/conftest.py
@@ -1,0 +1,46 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from assertpy import add_extension, assert_that
+from sqlalchemy.sql.sqltypes import ARRAY
+
+from trino.sqlalchemy.datatype import MAP, ROW, SQLType
+
+
+def assert_sqltype(this: SQLType, that: SQLType):
+    if isinstance(this, type):
+        this = this()
+    if isinstance(that, type):
+        that = that()
+    assert_that(type(this)).is_same_as(type(that))
+    if isinstance(this, ARRAY):
+        assert_sqltype(this.item_type, that.item_type)
+        if this.dimensions is None or this.dimensions == 1:
+            # ARRAY(dimensions=None) == ARRAY(dimensions=1)
+            assert_that(that.dimensions).is_in(None, 1)
+        else:
+            assert_that(this.dimensions).is_equal_to(this.dimensions)
+    elif isinstance(this, MAP):
+        assert_sqltype(this.key_type, that.key_type)
+        assert_sqltype(this.value_type, that.value_type)
+    elif isinstance(this, ROW):
+        assert_that(len(this.attr_types)).is_equal_to(len(that.attr_types))
+        for (this_attr, that_attr) in zip(this.attr_types, that.attr_types):
+            assert_that(this_attr[0]).is_equal_to(that_attr[0])
+            assert_sqltype(this_attr[1], that_attr[1])
+    else:
+        assert_that(str(this)).is_equal_to(str(that))
+
+
+@add_extension
+def is_sqltype(self, that):
+    this = self.val
+    assert_sqltype(this, that)
--- a/tests/sqlalchemy/test_datatype_parse.py
+++ b/tests/sqlalchemy/test_datatype_parse.py
@@ -1,0 +1,183 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pytest
+from assertpy import assert_that
+from sqlalchemy.sql.sqltypes import (
+    CHAR, VARCHAR,
+    ARRAY,
+    INTEGER, DECIMAL,
+    DATE, TIME, TIMESTAMP
+)
+from sqlalchemy.sql.type_api import TypeEngine
+
+from trino.sqlalchemy import datatype
+from trino.sqlalchemy.datatype import MAP, ROW
+
+
+@pytest.mark.parametrize(
+    'type_str, sql_type',
+    datatype._type_map.items(),
+    ids=datatype._type_map.keys()
+)
+def test_parse_simple_type(type_str: str, sql_type: TypeEngine):
+    actual_type = datatype.parse_sqltype(type_str)
+    if not isinstance(actual_type, type):
+        actual_type = type(actual_type)
+    assert_that(actual_type).is_equal_to(sql_type)
+
+
+parse_cases_testcases = {
+    'char(10)': CHAR(10),
+    'Char(10)': CHAR(10),
+    'char': CHAR(),
+    'cHaR': CHAR(),
+    'VARCHAR(10)': VARCHAR(10),
+    'varCHAR(10)': VARCHAR(10),
+    'VARchar(10)': VARCHAR(10),
+    'VARCHAR': VARCHAR(),
+    'VaRchAr': VARCHAR(),
+}
+
+
+@pytest.mark.parametrize(
+    'type_str, sql_type',
+    parse_cases_testcases.items(),
+    ids=parse_cases_testcases.keys()
+)
+def test_parse_cases(type_str: str, sql_type: TypeEngine):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
+
+
+parse_type_options_testcases = {
+    'CHAR(10)': CHAR(10),
+    'VARCHAR(10)': VARCHAR(10),
+    'DECIMAL(20)': DECIMAL(20),
+    'DECIMAL(20, 3)': DECIMAL(20, 3),
+    # TODO: support parametric timestamps (https://github.com/trinodb/trino-python-client/issues/107)
+}
+
+
+@pytest.mark.parametrize(
+    'type_str, sql_type',
+    parse_type_options_testcases.items(),
+    ids=parse_type_options_testcases.keys()
+)
+def test_parse_type_options(type_str: str, sql_type: TypeEngine):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
+
+
+parse_array_testcases = {
+    'array(integer)': ARRAY(INTEGER()),
+    'array(varchar(10))': ARRAY(VARCHAR(10)),
+    'array(decimal(20,3))': ARRAY(DECIMAL(20, 3)),
+    'array(array(varchar(10)))': ARRAY(VARCHAR(10), dimensions=2),
+    'array(map(char, integer))': ARRAY(MAP(CHAR(), INTEGER())),
+    'array(row(a integer, b varchar))': ARRAY(ROW([("a", INTEGER()), ("b", VARCHAR())])),
+}
+
+
+@pytest.mark.parametrize(
+    'type_str, sql_type',
+    parse_array_testcases.items(),
+    ids=parse_array_testcases.keys()
+)
+def test_parse_array(type_str: str, sql_type: ARRAY):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
+
+
+parse_map_testcases = {
+    'map(char, integer)': MAP(CHAR(), INTEGER()),
+    'map(varchar(10), varchar(10))': MAP(VARCHAR(10), VARCHAR(10)),
+    'map(varchar(10), decimal(20,3))': MAP(VARCHAR(10), DECIMAL(20, 3)),
+    'map(char, array(varchar(10)))': MAP(CHAR(), ARRAY(VARCHAR(10))),
+    'map(varchar(10), array(varchar(10)))': MAP(VARCHAR(10), ARRAY(VARCHAR(10))),
+    'map(varchar(10), array(array(varchar(10))))': MAP(VARCHAR(10), ARRAY(VARCHAR(10), dimensions=2)),
+}
+
+
+@pytest.mark.parametrize(
+    'type_str, sql_type',
+    parse_map_testcases.items(),
+    ids=parse_map_testcases.keys()
+)
+def test_parse_map(type_str: str, sql_type: ARRAY):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
+
+
+parse_row_testcases = {
+    'row(a integer, b varchar)':
+        ROW(attr_types=[
+            ("a", INTEGER()),
+            ("b", VARCHAR()),
+        ]),
+    'row(a varchar(20), b decimal(20,3))':
+        ROW(attr_types=[
+            ("a", VARCHAR(20)),
+            ("b", DECIMAL(20, 3)),
+        ]),
+    'row(x array(varchar(10)), y array(array(varchar(10))), z decimal(20,3))':
+        ROW(attr_types=[
+            ("x", ARRAY(VARCHAR(10))),
+            ("y", ARRAY(VARCHAR(10), dimensions=2)),
+            ("z", DECIMAL(20, 3)),
+        ]),
+    'row(min timestamp(6) with time zone, max timestamp(6) with time zone)':
+        ROW(attr_types=[
+            ("min", TIMESTAMP(timezone=True)),
+            ("max", TIMESTAMP(timezone=True)),
+        ]),
+    'row("first name" varchar, "last name" varchar)':
+        ROW(attr_types=[
+            ("first name", VARCHAR()),
+            ("last name", VARCHAR()),
+        ]),
+    'row("foo,bar" varchar, "foo(bar)" varchar, "foo\\"bar" varchar)':
+        ROW(attr_types=[
+            (r'foo,bar', VARCHAR()),
+            (r'foo(bar)', VARCHAR()),
+            (r'foo"bar', VARCHAR()),
+        ]),
+}
+
+
+@pytest.mark.parametrize(
+    'type_str, sql_type',
+    parse_row_testcases.items(),
+    ids=parse_row_testcases.keys()
+)
+def test_parse_row(type_str: str, sql_type: ARRAY):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
+
+
+parse_datetime_testcases = {
+    # TODO: support parametric timestamps (https://github.com/trinodb/trino-python-client/issues/107)
+    'date': DATE(),
+    'time': TIME(),
+    'time with time zone': TIME(timezone=True),
+    'timestamp': TIMESTAMP(),
+    'timestamp with time zone': TIMESTAMP(timezone=True),
+}
+
+
+@pytest.mark.parametrize(
+    'type_str, sql_type',
+    parse_datetime_testcases.items(),
+    ids=parse_datetime_testcases.keys()
+)
+def test_parse_datetime(type_str: str, sql_type: ARRAY):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
--- a/tests/sqlalchemy/test_dialect.py
+++ b/tests/sqlalchemy/test_dialect.py
@@ -1,0 +1,63 @@
+from typing import List, Any, Dict
+from unittest import mock
+
+import pytest
+from assertpy import assert_that
+from sqlalchemy.engine import make_url
+from sqlalchemy.engine.url import URL
+
+from trino.auth import BasicAuthentication
+from trino.dbapi import Connection
+from trino.sqlalchemy.dialect import TrinoDialect
+from trino.transaction import IsolationLevel
+
+
+class TestTrinoDialect:
+    def setup(self):
+        self.dialect = TrinoDialect()
+
+    # TODO: Test more authentication methods and URL params (https://github.com/trinodb/trino-python-client/issues/106)
+    @pytest.mark.parametrize(
+        'url, expected_args, expected_kwargs',
+        [
+            (make_url('trino://user@localhost'),
+             list(), dict(host='localhost', catalog='system', user='user')),
+
+            (make_url('trino://user@localhost:8080'),
+             list(), dict(host='localhost', port=8080, catalog='system', user='user')),
+
+            (make_url('trino://user:pass@localhost:8080'),
+             list(), dict(host='localhost', port=8080, catalog='system', user='user',
+                          auth=BasicAuthentication('user', 'pass'), http_scheme='https')),
+        ],
+    )
+    def test_create_connect_args(self, url: URL, expected_args: List[Any], expected_kwargs: Dict[str, Any]):
+        actual_args, actual_kwargs = self.dialect.create_connect_args(url)
+
+        assert_that(actual_args).is_equal_to(expected_args)
+        assert_that(actual_kwargs).is_equal_to(expected_kwargs)
+
+    def test_create_connect_args_missing_user_when_specify_password(self):
+        url = make_url('trino://:pass@localhost')
+        assert_that(self.dialect.create_connect_args).raises(ValueError) \
+            .when_called_with(url) \
+            .is_equal_to('Username is required when specify password in connection URL')
+
+    def test_create_connect_args_wrong_db_format(self):
+        url = make_url('trino://abc@localhost/catalog/schema/foobar')
+        assert_that(self.dialect.create_connect_args).raises(ValueError) \
+            .when_called_with(url) \
+            .is_equal_to('Unexpected database format catalog/schema/foobar')
+
+    def test_get_default_isolation_level(self):
+        isolation_level = self.dialect.get_default_isolation_level(mock.Mock())
+        assert_that(isolation_level).is_equal_to('AUTOCOMMIT')
+
+    def test_isolation_level(self):
+        dbapi_conn = Connection(host="localhost")
+
+        self.dialect.set_isolation_level(dbapi_conn, "SERIALIZABLE")
+        assert_that(dbapi_conn._isolation_level).is_equal_to(IsolationLevel.SERIALIZABLE)
+
+        isolation_level = self.dialect.get_isolation_level(dbapi_conn)
+        assert_that(isolation_level).is_equal_to("SERIALIZABLE")
--- a/trino/auth.py
+++ b/trino/auth.py
@@ -36,16 +36,16 @@ class Authentication(metaclass=abc.ABCMeta):
 
 class KerberosAuthentication(Authentication):
     def __init__(
-        self,
-        config: Optional[str] = None,
-        service_name: str = None,
-        mutual_authentication: bool = False,
-        force_preemptive: bool = False,
-        hostname_override: Optional[str] = None,
-        sanitize_mutual_error_response: bool = True,
-        principal: Optional[str] = None,
-        delegate: bool = False,
-        ca_bundle: Optional[str] = None,
+            self,
+            config: Optional[str] = None,
+            service_name: str = None,
+            mutual_authentication: bool = False,
+            force_preemptive: bool = False,
+            hostname_override: Optional[str] = None,
+            sanitize_mutual_error_response: bool = True,
+            principal: Optional[str] = None,
+            delegate: bool = False,
+            ca_bundle: Optional[str] = None,
     ) -> None:
         self._config = config
         self._service_name = service_name
@@ -87,6 +87,19 @@ class KerberosAuthentication(Authentication):
         except ImportError:
             raise RuntimeError("unable to import requests_kerberos")
 
+    def __eq__(self, other):
+        if not isinstance(other, KerberosAuthentication):
+            return False
+        return (self._config == other._config
+                and self._service_name == other._service_name
+                and self._mutual_authentication == other._mutual_authentication
+                and self._force_preemptive == other._force_preemptive
+                and self._hostname_override == other._hostname_override
+                and self._sanitize_mutual_error_response == other._sanitize_mutual_error_response
+                and self._principal == other._principal
+                and self._delegate == other._delegate
+                and self._ca_bundle == other._ca_bundle)
+
 
 class BasicAuthentication(Authentication):
     def __init__(self, username, password):
@@ -105,11 +118,17 @@ class BasicAuthentication(Authentication):
     def get_exceptions(self):
         return ()
 
+    def __eq__(self, other):
+        if not isinstance(other, BasicAuthentication):
+            return False
+        return self._username == other._username and self._password == other._password
+
 
 class _BearerAuth(AuthBase):
     """
     Custom implementation of Authentication class for bearer token
     """
+
     def __init__(self, token):
         self.token = token
 
@@ -130,6 +149,11 @@ class JWTAuthentication(Authentication):
     def get_exceptions(self):
         return ()
 
+    def __eq__(self, other):
+        if not isinstance(other, JWTAuthentication):
+            return False
+        return self.token == other.token
+
 
 def handle_redirect_auth_url(auth_url):
     print("Open the following URL in browser for the external authentication:")
@@ -140,6 +164,7 @@ class _OAuth2TokenBearer(AuthBase):
     """
     Custom implementation of Trino Oauth2 based authorization to get the token
     """
+
     MAX_OAUTH_ATTEMPTS = 5
 
     class _AuthStep(Enum):
@@ -160,16 +185,16 @@ class _OAuth2TokenBearer(AuthBase):
             self._thread_local.auth_step = self._AuthStep.GET_REDIRECT_SERVER
 
         if self._thread_local.token:
-            r.headers['Authorization'] = "Bearer " + self._thread_local.token
+            r.headers['Authorization'] = f"Bearer {self._thread_local.token}"
 
         r.register_hook('response', self.__authenticate)
 
         return r
 
     def __authenticate(self, r, **kwargs):
-        if (self._thread_local.auth_step == self._AuthStep.GET_REDIRECT_SERVER):
+        if self._thread_local.auth_step == self._AuthStep.GET_REDIRECT_SERVER:
             self.__process_get_redirect_server(r)
-        elif (self._thread_local.auth_step == self._AuthStep.GET_TOKEN):
+        elif self._thread_local.auth_step == self._AuthStep.GET_TOKEN:
             self.__process_get_token(r)
         return r
 
@@ -249,3 +274,8 @@ class OAuth2Authentication(Authentication):
 
     def get_exceptions(self):
         return ()
+
+    def __eq__(self, other):
+        if not isinstance(other, OAuth2Authentication):
+            return False
+        return self._redirect_auth_url == other._redirect_auth_url
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -29,10 +29,44 @@ from trino import constants
 import trino.exceptions
 import trino.client
 import trino.logging
-from trino.transaction import Transaction, IsolationLevel, NO_TRANSACTION
+from trino.transaction import (
+    Transaction,
+    IsolationLevel,
+    NO_TRANSACTION
+)
+from trino.exceptions import (
+    Warning,
+    Error,
+    InterfaceError,
+    DatabaseError,
+    DataError,
+    OperationalError,
+    IntegrityError,
+    InternalError,
+    ProgrammingError,
+    NotSupportedError,
+)
 
-
-__all__ = ["connect", "Connection", "Cursor"]
+__all__ = [
+    # https://www.python.org/dev/peps/pep-0249/#globals
+    "apilevel",
+    "threadsafety",
+    "paramstyle",
+    "connect",
+    "Connection",
+    "Cursor",
+    # https://www.python.org/dev/peps/pep-0249/#exceptions
+    "Warning",
+    "Error",
+    "InterfaceError",
+    "DatabaseError",
+    "DataError",
+    "OperationalError",
+    "IntegrityError",
+    "InternalError",
+    "ProgrammingError",
+    "NotSupportedError",
+]
 
 
 apilevel = "2.0"
--- a/trino/sqlalchemy/compiler.py
+++ b/trino/sqlalchemy/compiler.py
@@ -1,0 +1,143 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from sqlalchemy.sql import compiler
+
+# https://trino.io/docs/current/language/reserved.html
+RESERVED_WORDS = {
+    "alter",
+    "and",
+    "as",
+    "between",
+    "by",
+    "case",
+    "cast",
+    "constraint",
+    "create",
+    "cross",
+    "cube",
+    "current_catalog",
+    "current_date",
+    "current_path",
+    "current_role",
+    "current_schema",
+    "current_time",
+    "current_timestamp",
+    "current_user",
+    "deallocate",
+    "delete",
+    "describe",
+    "distinct",
+    "drop",
+    "else",
+    "end",
+    "escape",
+    "except",
+    "execute",
+    "exists",
+    "extract",
+    "false",
+    "for",
+    "from",
+    "full",
+    "group",
+    "grouping",
+    "having",
+    "in",
+    "inner",
+    "insert",
+    "intersect",
+    "into",
+    "is",
+    "join",
+    "left",
+    "like",
+    "localtime",
+    "localtimestamp",
+    "natural",
+    "normalize",
+    "not",
+    "null",
+    "on",
+    "or",
+    "order",
+    "outer",
+    "prepare",
+    "recursive",
+    "right",
+    "rollup",
+    "select",
+    "skip",
+    "table",
+    "then",
+    "true",
+    "uescape",
+    "union",
+    "unnest",
+    "using",
+    "values",
+    "when",
+    "where",
+    "with",
+}
+
+
+class TrinoSQLCompiler(compiler.SQLCompiler):
+    pass
+
+
+class TrinoDDLCompiler(compiler.DDLCompiler):
+    pass
+
+
+class TrinoTypeCompiler(compiler.GenericTypeCompiler):
+    def visit_FLOAT(self, type_, **kw):
+        precision = type_.precision or 32
+        if 0 <= precision <= 32:
+            return self.visit_REAL(type_, **kw)
+        elif 32 < precision <= 64:
+            return self.visit_DOUBLE(type_, **kw)
+        else:
+            raise ValueError(f"type.precision must be in range [0, 64], got {type_.precision}")
+
+    def visit_DOUBLE(self, type_, **kw):
+        return "DOUBLE"
+
+    def visit_NUMERIC(self, type_, **kw):
+        return self.visit_DECIMAL(type_, **kw)
+
+    def visit_NCHAR(self, type_, **kw):
+        return self.visit_CHAR(type_, **kw)
+
+    def visit_NVARCHAR(self, type_, **kw):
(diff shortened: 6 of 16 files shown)


findepi:
Let me know when this becomes non-draft.

dungdm93:
@findepi Do you know why CI say `ModuleNotFoundError: No module named 'trino.sqlalchemy'`?
I run test in my local machine, it works fine.

hashhar:
@dungdm93 Could it be due to https://github.com/trinodb/trino-python-client/blob/16feda14efb1a0985053d92338313f32548d1095/trino/__init__.py (no trino.sqlalchemy entry there).

cc: @joshthoward

trino/__init__.py
```diff
@@ -17,6 +17,6 @@
 from . import exceptions
 from . import logging
 
-__all__ = ['auth', 'dbapi', 'client', 'constants', 'exceptions', 'logging']
+__all__ = ['auth', 'dbapi', 'client', 'constants', 'exceptions', 'logging', 'sqlalchemy']
```
hashhar, line 20: This has no effect. the`__all__` only serves as documentation/convention. The `from . import sqlalchemy` seems to be the missing piece but Josh might know better.

findepi: `__all__` seems to define the meaning for `from trino import *`, so it's kind of part of public API.

i think it's fair to require that `trino.sqlalchemy` be explicitly imported by a user

dungdm93:
Hello @findepi, @hashhar 
I'll add more tests later, but you can review it now.

findepi:
> feat: support python 3.5

this shouldn't be needed. i just merged https://github.com/trinodb/trino-python-client/pull/83 so that you can drop this one

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,301 @@
...
+        """Trino has no support for foreign keys. Returns an empty list."""
+        return []
+
+    def get_schema_names(self, connection: Connection, **kw) -> List[str]:
+        query = 'SHOW SCHEMAS'
+        res = connection.execute(sql.text(query))
+        return [row.Schema for row in res]
+
+    def get_table_names(self, connection: Connection, schema: str = None, **kw) -> List[str]:
+        query = 'SHOW TABLES'
```
hashhar, line 136: Should we use `information_schema` everywhere instead?

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,301 @@
...
+        res = connection.execute(sql.text(query))
+        return [row.Table for row in res]
+
+    def get_temp_table_names(self, connection: Connection, schema: str = None, **kw) -> List[str]:
+        """Trino has no support for temporary tables. Returns an empty list."""
+        return []
+
+    def get_view_names(self, connection: Connection, schema: str = None, **kw) -> List[str]:
+        schema = schema or self._get_default_schema_name(connection)
+        if schema is None:
```
hashhar, line 148: The sqlalchemy API docs say the schema is optional - we should treat it as optional in our method too.

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,301 @@
...
+        except error.TrinoQueryError as e:
+            if e.error_name in (
+                    error.TABLE_NOT_FOUND,
+                    error.SCHEMA_NOT_FOUND,
+                    error.CATALOG_NOT_FOUND,
+            ):
+                return False
+            raise
+
+    def has_table(self, connection: Connection,
```
hashhar, line 230: The API docs say that the table being searched is an exact table instead of a pattern - so we should be escaping the `table_name` properly here.

Same in `has_schema`.

Refer to the postgresql dialect here: https://github.com/sqlalchemy/sqlalchemy/blob/93c8f1df8bb68820624559506f6713a16a9e2250/lib/sqlalchemy/dialects/postgresql/base.py#L3302-L3355

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,301 @@
...
+        pass
+
+    def do_recover_twophase(self, connection: Connection) -> None:
+        pass
+
+    def set_isolation_level(self, dbapi_conn: trino_dbapi.Connection, level) -> None:
+        dbapi_conn._isolation_level = getattr(trino_dbapi.IsolationLevel, level)
+
+    def get_isolation_level(self, dbapi_conn: trino_dbapi.Connection) -> str:
+        level_names = ['AUTOCOMMIT',
```
hashhar, line 298: Should the dbapi declare the levels as an exported field instead of repeating them here?

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,301 @@
...
+                       'READ_UNCOMMITTED',
+                       'READ_COMMITTED',
+                       'REPEATABLE_READ',
+                       'SERIALIZABLE']
+        return level_names[dbapi_conn.isolation_level]
+
+    def _get_full_table(self, table_name: str, schema: str = None, quote: bool = True) -> str:
+        table_part = self.identifier_preparer.quote_identifier(table_name) if quote else table_name
```
hashhar, line 306: nit: To me the Python ternary style isn't very readable.

trino/sqlalchemy/datatype.py
```diff
@@ -0,0 +1,169 @@
...
+    def __init__(self, attr_types: Dict[str, SQLType]):
+        for name, attr_type in attr_types.items():
+            if isinstance(attr_type, type):
+                attr_type = attr_type()
+                attr_types[name] = attr_type
+        self.attr_types: Dict[str, TypeEngine] = attr_types
+
+    @property
+    def python_type(self):
+        return dict
```
hashhar, line 97: An ordered dict impl would be more like a `ROW`.

trino/sqlalchemy/datatype.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def split(string: str, delimiter: str = ',',
+          quote: str = '"', escaped_quote: str = r'\"',
+          open_bracket: str = '(', close_bracket: str = ')') -> Iterator[str]:
+    """
+    A split function that is aware of quotes and brackets/parentheses.
```
hashhar, line 104: Since you have already added a comment, an example (or a self-explanatory method name) would help even more because the method impl is not very straightforward.

trino/sqlalchemy/datatype.py
```diff
@@ -0,0 +1,169 @@
...
+    if not match:
+        util.warn(f"Could not parse type name '{type_str}'")
+        return sqltypes.NULLTYPE
+    type_name = match.group("type")
+    type_opts = match.group("options")
+
+    if type_name == "array":
+        item_type = parse_sqltype(type_opts)
+        if isinstance(item_type, sqltypes.ARRAY):
+            dimensions = (item_type.dimensions or 1) + 1
```
hashhar, line 145: Will this not return 2 as the minimum value instead of the expected 1?

trino/sqlalchemy/datatype.py
```diff
@@ -0,0 +1,169 @@
...
+        return sqltypes.ARRAY(item_type)
+    elif type_name == "map":
+        key_type_str, value_type_str = split(type_opts)
+        key_type = parse_sqltype(key_type_str)
+        value_type = parse_sqltype(value_type_str)
+        return MAP(key_type, value_type)
+    elif type_name == "row":
+        attr_types: Dict[str, SQLType] = {}
+        for attr_str in split(type_opts):
+            name, attr_type_str = split(attr_str.strip(), delimiter=' ')
```
hashhar, line 156: What about `ROW("a space" VARCHAR, "another space" INT)`? splitting on spaces seems incorrect?
Note that types can also be implicit as in `MAP(ARRAY['foo', 'bar'], ARRAY[1, 2])`.

My examples may not be valid though because I haven't yet understood who calls this class/method and with what kind of inputs.

tests/sqlalchemy/test_datatype_parse.py
```diff
@@ -0,0 +1,127 @@
...
+)
+def test_parse_map(type_str: str, sql_type: ARRAY):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
+
+
+parse_row_testcases = {
+    'row(a integer, b varchar)': ROW(dict(a=INTEGER(), b=VARCHAR())),
+    'row(a varchar(20), b decimal(20,3))': ROW(dict(a=VARCHAR(20), b=DECIMAL(20, 3))),
+    'row(x array(varchar(10)), y array(array(varchar(10))), z decimal(20,3))':
```
hashhar, line 96: Also test row types with fields having special characters like `,`, `(` or whitespace.

tests/sqlalchemy/test_datatype_parse.py
```diff
@@ -0,0 +1,127 @@
...
+    'type_str, sql_type',
+    parse_row_testcases.items(),
+    ids=parse_row_testcases.keys()
+)
+def test_parse_row(type_str: str, sql_type: ARRAY):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
+
+
+parse_datetime_testcases = {
```
hashhar, line 111: Parametrized timestamps? Or does trino-python-client not support them yet?

trino/__init__.py
```diff
@@ -10,13 +10,4 @@
...
 # limitations under the License.
 
-from . import auth
-from . import dbapi
-from . import client
-from . import constants
-from . import exceptions
-from . import logging
-
-__all__ = ['auth', 'dbapi', 'client', 'constants', 'exceptions', 'logging']
```
hashhar, line 20: Why was this change needed?

integration_tests/test_dbapi.py
```diff
@@ -307,7 +307,7 @@ def test_cancel_query(trino_connection):
 def test_session_properties(run_trino):
     _, host, port = run_trino
 
-    connection = trino.dbapi.Connection(
```
hashhar, line 310: nit: separate commit.

.github/workflows/ci.yml
```diff
@@ -23,10 +22,10 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install libkrb5-dev
-          pip install .[tests]
+          pip install -e .[tests]
       - name: Run tests
         run: |
-          pytest -s tests/ integration_tests/
+          pytest -s .
```
hashhar, line 28: Is there any way to verify that the tests running before this change also run after this change? i.e. none are skipped.

trino/sqlalchemy/compiler.py
```diff
@@ -96,7 +96,44 @@ class TrinoDDLCompiler(compiler.DDLCompiler):
...
 class TrinoTypeCompiler(compiler.GenericTypeCompiler):
-    pass
+    def visit_FLOAT(self, type_, **kw):
+        precision = type_.precision or 32
+        if 0 <= precision <= 32:
+            return self.visit_REAL(type_, **kw)
+        elif 32 < precision <= 64:
+            return self.visit_DOUBLE(type_, **kw)
+        else:
+            raise ValueError(f"type.precision={type_.precision} is invalid")
```
hashhar, line 106: Can we also include the min and max expected precisions in the error message?

trino/sqlalchemy/dialect.py
```diff
@@ -260,6 +261,15 @@ def _get_default_schema_name(self, connection: Connection) -> Optional[str]:
...
         return dbapi_connection.schema
 
+    def do_execute(self, cursor: Cursor, statement: str, parameters: Tuple[Any, ...],
+                   context: DefaultExecutionContext = None):
+        cursor.execute(statement, parameters)
+        if context and context.should_autocommit:
+            # SQL statement only submitted to Trino server when cursor.fetch*() is called.
+            # For DDL (CREATE/ALTER/DROP) and DML (INSERT/UPDATE/DELETE) statement, call cursor.description
+            # to force submit statement immediately.
+            cursor.description # noqa
```
hashhar, line 271: This relies on the impl. detail that description calls fetch. why not directly call fetch here?

hashhar (commented):
Sorry for the delay in reviewing. Thanks a lot for your work.

A few comments.

I did not review [feat: TrinoTypeCompiper datatypes](https://github.com/trinodb/trino-python-client/pull/81/commits/8abfc0acbf97a55c1cfdc75a0a01a791b851f7d9) yet.

Also, I did not completely understand `datatype.py` added in https://github.com/trinodb/trino-python-client/pull/81/commits/66f6666265aab1b20bfa90717092279b383ef0f8.

trino/sqlalchemy/compiler.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from sqlalchemy.sql import compiler
+
+# https://trino.io/docs/current/language/reserved.html
+RESERVED_WORDS = {
```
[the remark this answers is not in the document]
dungdm93: Yeah. All other dialect also hard coded this list:
* [MySQL](https://github.com/sqlalchemy/sqlalchemy/blob/88c0ff61fa07cfa1d9caf24e6ead11851894d098/lib/sqlalchemy/dialects/mysql/base.py#L1001)
* [PostgreSQL](https://github.com/sqlalchemy/sqlalchemy/blob/88c0ff61fa07cfa1d9caf24e6ead11851894d098/lib/sqlalchemy/dialects/postgresql/base.py#L1402)
or 3rd party dialects:
* [cockroachdb](https://github.com/cockroachdb/sqlalchemy-cockroachdb/blob/d41697911716f62bf9cfe3a9ce14745e44b3836c/cockroachdb/sqlalchemy/stmt_compiler.py#L7)

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,301 @@
...
+    def dbapi(cls):
+        """
+        ref: https://www.python.org/dev/peps/pep-0249/#module-interface
+        """
+        return trino_dbapi
+
+    def create_connect_args(self, url: URL) -> Tuple[List[Any], Dict[str, Any]]:
+        args, kwargs = super(TrinoDialect, self).create_connect_args(url)  # type: List[Any], Dict[str, Any]
+
+        db_parts = kwargs.pop('database', 'hive').split('/')
```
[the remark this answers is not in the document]
dungdm93: Because [PyHive use this value](https://github.com/dropbox/PyHive/blob/b21c507a24ed2f2b0cf15b0b6abb1c43f31d3ee0/pyhive/sqlalchemy_presto.py#L93). But I just merge a PR to change database default value to `system`.

trino/sqlalchemy/dialect.py
dungdm93: Is there any reason to avoid using `SHOW` command?

trino/sqlalchemy/dialect.py
dungdm93: Because `cursor.description` also put fetch result into buffer (if any)

trino/sqlalchemy/dialect.py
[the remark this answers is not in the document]
hashhar: Is a default database required? i.e. what would fail if no default database was set?
The dbapi implementation doesn't set any default database today.
Note that the user may not have permission to access whatever default catalog we pick.

If it's required then `system` seems like the correct choice because it'll always exist.

hashhar:
@dungdm93 Did you get a chance to address the comments? I would really love to see this merged. Let me know if there's anything I can help with to push this over the finish line.

dungdm93:
Sorry @hashhar to let you waiting. I so busy recently. I’ll try to finish this task next week.

dungdm93:
Sorry everyone for making you wait so long. I'm back with this PR now.

trino/sqlalchemy/dialect.py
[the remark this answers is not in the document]
dungdm93: I have change default database to `system`

.github/workflows/ci.yml
dungdm93: I change it back

trino/__init__.py
dungdm93: The revert question is why does we need this?
I remove those lines because I don't wanna add an import for every new files.

trino/sqlalchemy/dialect.py
dungdm93: When `schema=None`, `SELECT * FROM information_schema.tables` return tables in all schemas but what we wan't is get tables in current schema.
Furthermore, other dialects like [MySQL](https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_3_24/lib/sqlalchemy/dialects/mysql/base.py#L2734) also use `SHOW` command. I don't see any problem if we are going to the same direction.

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,301 @@
...
+        if len(db_parts) == 1:
+            kwargs['catalog'] = db_parts[0]
+        elif len(db_parts) == 2:
+            kwargs['catalog'] = db_parts[0]
+            kwargs['schema'] = db_parts[1]
+        else:
+            raise ValueError(f'Unexpected database format {url.database}')
+
+        username = kwargs.pop('username', 'anonymous')
```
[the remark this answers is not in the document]
dungdm93: If username is empty, we get this error:
```
trino.exceptions.HttpError: error 401: b'Basic authentication or X-Trino-User must be sent'
```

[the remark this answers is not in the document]
ebyhr: @findepi Do you think we should use OS username by default in Python client as well?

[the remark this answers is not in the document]
dungdm93: @ebyhr is that implemented any where else (like CLI or JDBC)?

trino/sqlalchemy/datatype.py
dungdm93: `2` is expected value. `None` means `dimensions=1`.

You could see example in my test-cases:
```
parse_array_testcases = {
    'array(integer)': ARRAY(INTEGER()),
    'array(varchar(10))': ARRAY(VARCHAR(10)),
    'array(decimal(20,3))': ARRAY(DECIMAL(20, 3)),
    'array(array(varchar(10)))': ARRAY(VARCHAR(10), dimensions=2),
}
```

trino/sqlalchemy/dialect.py
dungdm93: Optional means schema already provided in the `connection`.

trino/sqlalchemy/dialect.py
dungdm93: Yeah, but LIKE statement don't use pattern. So it's exactly match.

FYI, I already changed to using `information_schema` instead of `SHOW`.

trino/sqlalchemy/dialect.py
dungdm93: Yeah, I just refactor `IsolationLevel` to `Enum`. So no need to define `level_names` here :+1:

trino/sqlalchemy/dialect.py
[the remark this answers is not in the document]
findepi: @ebyhr @dungdm93 yes, we should default to os username, as CLI does, or require username to be set.
JDBC requires `user` to be set: https://github.com/trinodb/trino/blob/ceb2d1b8238c4b6949abf2e7f506aab83bf1ea02/client/trino-jdbc/src/main/java/io/trino/jdbc/ConnectionProperties.java#L157

trino/sqlalchemy/datatype.py
```diff
@@ -0,0 +1,184 @@
...
+    :param delimiter: string defining where to split, usually a comma or space
+    :param maxsplit: Maximum number of splits to do. -1 (default) means no limit.
+    :param quote: string, either a single or a double quote
+    :param escaped_quote: string representing an escaped quote
+    :param open_bracket: string, either [, {, < or (
+    :param close_bracket: string, either ], }, > or )
+    """
+    parens = 0
+    quotes = False
+    i = 0
```
hashhar, line 121: nit: rename to more explanatory name.

trino/sqlalchemy/datatype.py
```diff
@@ -0,0 +1,184 @@
...
+    if maxsplit < -1:
+        raise ValueError(f"maxsplit must be a positive integer or -1, got {maxsplit}")
+    elif maxsplit == 0:
+        yield string
+        return
+    for j, character in enumerate(string):
```
hashhar, line 127: nit: rename `j` to more explanatory name.

trino/sqlalchemy/datatype.py
hashhar: Add a comment that None means 1-dimensional. It's not obvious.

trino/sqlalchemy/datatype.py
```diff
@@ -0,0 +1,184 @@
...
+    elif type_name == "row":
+        attr_types: Dict[str, SQLType] = {}
+        for attr_str in split(type_opts):
+            name, attr_type_str = split(attr_str.strip(), delimiter=' ', maxsplit=1)
+            attr_type = parse_sqltype(attr_type_str)
+            attr_types[name] = attr_type
+        return ROW(attr_types)
+
+    if type_name not in _type_map:
+        util.warn(f"Did not recognize type '{type_name}'")
```
hashhar, line 177: Should this be an error? Or does returning `NULLTYPE` make sure errors are raised?

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,314 @@
...
+                name=record.column_name,
+                type=datatype.parse_sqltype(record.data_type),
+                nullable=record.is_nullable == 'YES',
+                default=record.column_default,
+            )
+            columns.append(column)
+        return columns
+
+    def get_pk_constraint(self, connection: Connection,
+                          table_name: str, schema: str = None, **kw) -> Dict[str, Any]:
```
hashhar, line 116: Any specific way we are splitting arguments across lines? 

Maybe we should introduce an autoformatter like `black` here?

(No changes requested)

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def get_primary_keys(self, connection: Connection,
+                         table_name: str, schema: str = None, **kw) -> List[str]:
+        pk = self.get_pk_constraint(connection, table_name, schema)
+        return pk.get('constrained_columns')  # type: List[str]
+
+    def get_foreign_keys(self, connection: Connection,
+                         table_name: str, schema: str = None, **kw) -> List[Dict[str, Any]]:
+        """Trino has no support for foreign keys. Returns an empty list."""
+        return []
```
hashhar, line 128: Reminder to self: verify behaviour expected by sqlalchemy (for this and other methods)

trino/sqlalchemy/dialect.py
hashhar: `LIKE 'a_table'` will match `abtable` too.

Also, I still see SHOW TABLES here - did you forget to push some commits?

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,314 @@
...
+            # For DDL (CREATE/ALTER/DROP) and DML (INSERT/UPDATE/DELETE) statement, call cursor.description
+            # to force submit statement immediately.
+            cursor.description  # noqa
+
+    def do_rollback(self, dbapi_connection: trino_dbapi.Connection):
+        if dbapi_connection.transaction is not None:
+            dbapi_connection.rollback()
+
+    def do_begin_twophase(self, connection: Connection, xid: str):
+        pass
```
hashhar, line 278: Should these `pass` or `raise` something? We don't want an application to try to call `do_rollback_twophase` and then notice that the operation didn't get rolled back.

(Reminder to self - see what other clients do).

trino/sqlalchemy/dialect.py
hashhar: Reminder - looks like we still have `level_names` used here.

hashhar (commented):
Reviewed 1st commit "feat: add trino sqlalchemy dialect"

I will do another review after taking a look at some other sqlalchemy dialect impl and the sqlalchemy docs to see expected semantics of some methods.

trino/sqlalchemy/dialect.py
hashhar: Nevermind, see it's fixed in next commit.

tests/sqlalchemy/test_dialect.py
```diff
@@ -0,0 +1,34 @@
...
+    @pytest.mark.parametrize(
+        'url, expected_args, expected_kwargs',
+        [
+            (make_url('trino://localhost'),
+             list(), dict(host='localhost', catalog='system', user='anonymous')),
+            (make_url('trino://1.2.3.4:4321/mysql/sakila'),
+             list(), dict(host='1.2.3.4', port=4321, catalog='mysql', schema='sakila', user='anonymous')),
+            (make_url('trino://user@localhost:8080'),
+             list(), dict(host='localhost', port=8080, catalog='system', user='user')),
+            (make_url('trino://user:pass@localhost:8080'),
```
hashhar, line 28: Does the python client accept and URL parameters?

tests/sqlalchemy/test_dialect.py
```diff
@@ -0,0 +1,34 @@
...
+             list(), dict(host='localhost', port=8080, catalog='system', user='user',
+                          auth=BasicAuthentication('user', 'pass'), http_scheme='https')),
```
hashhar, line 30: Can we test with JWK/JWT as well?

Also with the ssl_verify (don't remember exactly) parameter?

trino/auth.py
```diff
@@ -128,11 +128,18 @@ def get_exceptions(self):
     def handle_error(self, handle_error):
         pass
 
+    def __eq__(self, other):
```
hashhar, line 131: Do all `Authentication` impls need to implement this? If so, let's move this to `Authentication` with a default that `raise`s an exception maybe?

hashhar (commented):
3rd commit: "test: dialect.create_connect_args"

tests/sqlalchemy/test_dialect.py
```diff
@@ -32,3 +35,16 @@ def test_create_connect_args(self, url: URL, expected_args: List[Any], expected_
...
+        il = self.dialect.get_default_isolation_level(mock.Mock())
+        assert_that(il).is_equal_to('AUTOCOMMIT')
+
+    def test_isolation_level(self):
+        dbapi_conn = Connection(host="localhost")
+
+        self.dialect.set_isolation_level(dbapi_conn, "SERIALIZABLE")
+        assert_that(dbapi_conn._isolation_level).is_equal_to(IsolationLevel.SERIALIZABLE)
+
+        il = self.dialect.get_isolation_level(dbapi_conn)
```
hashhar, line 49: nit: avoid abbreviations. `il` -> `isolation_level`.

trino/transaction.py
```diff
@@ -35,16 +35,16 @@ class IsolationLevel(object):
 
     @classmethod
     def levels(cls) -> Iterable[str]:
-        return {k for k, v in cls.__dict__.items() if not k.startswith("_") and isinstance(v, int)}
+        return {il.name for il in IsolationLevel}
```
hashhar, line 38: Avoid abbreviations. Use `isolation_level` instead.

hashhar: Here and other places.

trino/transaction.py
```diff
@@ -74,16 +72,14 @@ def begin(self):
...
 
     def commit(self):
         query = trino.client.TrinoQuery(self._request, COMMIT)
         try:
             list(query.execute())
         except Exception as err:
-            raise trino.exceptions.DatabaseError(
-                "failed to commit transaction {}: {}".format(self._id, err)
-            )
+            raise trino.exceptions.DatabaseError(f"failed to commit transaction {self._id}") from err
```
hashhar, line 82: What does `from err` do? First time seeing this. Interesting.

hashhar (commented):
4th commit: "refactor: change IsolationLevel to Enum"

Can we extract the formatting related changes to a separate first commit? It makes it much easier to see the places where actual changes have happenned.

A few comments. I'll do an in-depth review once again after reading up about the sqlalchemy api expectations.

tests/sqlalchemy/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from assertpy import add_extension, assert_that
+from sqlalchemy.sql.sqltypes import ARRAY
+
+from trino.sqlalchemy.datatype import MAP, ROW, SQLType
+
+
+def assert_sqltype(this: SQLType, that: SQLType):
+    if isinstance(this, type):
```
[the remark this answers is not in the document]
dungdm93: It's python built-in class.  It check `this` is a class or not.
In SQLAlchemy, SQLType can be either a class or an instance, e.g:
```
class User(Base):
    firstname = Column(String)
    lastname = Column(String())
```

tests/sqlalchemy/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+        this = this()
+    if isinstance(that, type):
+        that = that()
+    assert_that(type(this)).is_same_as(type(that))
+    if isinstance(this, ARRAY):
+        assert_sqltype(this.item_type, that.item_type)
+        if this.dimensions is None or this.dimensions == 1:
+            assert_that(that.dimensions).is_in(None, 1)
```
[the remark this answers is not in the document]
dungdm93: It's not special, it's because of an edge case `ARRAY(dimensions=None) == ARRAY(dimensions=1)`

trino/transaction.py
dungdm93: It's equal `throw new Exception("something went wrong", cause)` in Java.
You could see more info here [PEP-3134](https://www.python.org/dev/peps/pep-3134/)

trino/auth.py
dungdm93: Like java, `__eq__` available in all object, but by default, it check object identity.   
I added `__eq__` to all `Authentication` impls, but I don't think raise an exception in base `Authentication` class is good idea, because it's public API and may introduce breaking change if some one has they own `Authentication`

tests/sqlalchemy/test_dialect.py
dungdm93: To keep this PR small, I think others authentication methods e.g Kerberos, JWT,.. should be add later, in other PRs.

tests/sqlalchemy/test_datatype_parse.py
```diff
@@ -0,0 +1,129 @@
...
+    actual_type = datatype.parse_sqltype(type_str)
+    if not isinstance(actual_type, type):
+        actual_type = type(actual_type)
+    assert_that(actual_type).is_equal_to(sql_type)
+
+
+parse_type_options_testcases = {
+    'VARCHAR(10)': VARCHAR(10),
+    'DECIMAL(20)': DECIMAL(20),
+    'DECIMAL(20, 3)': DECIMAL(20, 3),
```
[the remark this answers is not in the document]
dungdm93: Yes, TODO. Timestamp w/ precision is not supported yet.

trino/sqlalchemy/dialect.py
hashhar: Seems this is fixed in a later commit.

tests/sqlalchemy/conftest.py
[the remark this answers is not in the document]
hashhar: Add a comment that both None and 1 mean a 1-dimensional array. It'll not be obvious to other readers.

tests/sqlalchemy/test_dialect.py
hashhar: Create a GitHub issue and add a TODO so that we remember to do that.

trino/sqlalchemy/compiler.py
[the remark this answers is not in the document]
dungdm93: Hello @hashhar , I don't see `MATCH`, and `RECOGNIZE` keyword in this page https://trino.io/docs/current/language/reserved.html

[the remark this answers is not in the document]
hashhar: cc: @kasiafi Are MATCH and RECOGNIZE (and other keywords that got added recently) "reserved"? If so we should update the documentation too.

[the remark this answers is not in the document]
kasiafi: The only reserved keyword that was added along with the new features was `SKIP`. Other keywords, including `MATCH_RECOGNIZE`, were added as non-reserved, which is an omission. The spec defines it as reserved.

IIRC, we have an automated check that all new reserved keywords appear in the docs.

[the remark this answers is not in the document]
mosabua: I am not sure about that check working to be honest.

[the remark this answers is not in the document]
ebyhr: We have below check for reserved keywords:
- https://github.com/trinodb/trino/blob/master/docs/pom.xml#L68-L81
- https://github.com/trinodb/trino/blob/master/core/trino-parser/src/main/java/io/trino/sql/ReservedIdentifiers.java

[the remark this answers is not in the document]
kasiafi: I extracted the keywords that are reserved by spec from the recently added features:
https://github.com/trinodb/trino/pull/9172

trino/sqlalchemy/dialect.py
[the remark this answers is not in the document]
dungdm93: @findepi @ebyhr @hashhar I think we should going to the same direction as JDBC does.  
So `user` is required.

tests/sqlalchemy/test_dialect.py
dungdm93: @hashhar #106

tests/sqlalchemy/test_dialect.py
dungdm93: Not yet. But it should be.  
I create #108 to tracking this issue

tests/sqlalchemy/test_datatype_parse.py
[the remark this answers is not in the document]
dungdm93: Keep track in #107

tests/sqlalchemy/conftest.py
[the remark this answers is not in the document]
dungdm93: Comment added

tests/sqlalchemy/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            assert_that(this.dimensions).is_equal_to(this.dimensions)
+    elif isinstance(this, MAP):
+        assert_sqltype(this.key_type, that.key_type)
+        assert_sqltype(this.value_type, that.value_type)
+    elif isinstance(this, ROW):
+        assert_that(len(this.attr_types)).is_equal_to(len(that.attr_types))
+        for name, this_attr in this.attr_types.items():
```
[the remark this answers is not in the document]
dungdm93: Thank @hashhar
I just change ROW presentation from `Dict` to `List[Tuple[name, type]]`. You wanna take a look.

trino/sqlalchemy/compiler.py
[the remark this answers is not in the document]
dungdm93: @ebyhr, @kasiafi So this page is up to date, right? https://trino.io/docs/current/language/reserved.html

@hashhar I just update the list of keywords base on the page above.

tests/sqlalchemy/test_datatype_parse.py
dungdm93: #107

trino/sqlalchemy/datatype.py
dungdm93: Could you have any better name?

tests/sqlalchemy/test_datatype_parse.py
```diff
@@ -0,0 +1,129 @@
...
+def test_parse_type_options(type_str: str, sql_type: TypeEngine):
+    actual_type = datatype.parse_sqltype(type_str)
+    assert_that(actual_type).is_sqltype(sql_type)
+
+
+parse_array_testcases = {
+    'array(integer)': ARRAY(INTEGER()),
+    'array(varchar(10))': ARRAY(VARCHAR(10)),
+    'array(decimal(20,3))': ARRAY(DECIMAL(20, 3)),
+    'array(array(varchar(10)))': ARRAY(VARCHAR(10), dimensions=2),
```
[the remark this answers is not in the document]
dungdm93: Done

tests/sqlalchemy/test_datatype_parse.py
dungdm93: I added 2 more testcases bellow.

trino/sqlalchemy/datatype.py
dungdm93: Other dialects also return `NULLTYPE`

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,314 @@
...
+    supports_sequences = False
+    supports_comments = True
+    inline_comments = True
+    supports_default_values = False
+
+    # DDL
+    supports_alter = True
+
+    # DML
+    supports_empty_insert = False
```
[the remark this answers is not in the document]
dungdm93: I guest  `INSERT INTO some_table VALUES ()` or `INSERT INTO some_table VALUES (NULL)`. I'm not very so sure.

trino/sqlalchemy/dialect.py
dungdm93: All commits are pushed, could you see it?

trino/sqlalchemy/dialect.py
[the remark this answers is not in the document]
hashhar: Looks like it's `INSERT () VALUES ()` according to https://docs.sqlalchemy.org/en/14/core/internals.html#sqlalchemy.engine.default.DefaultDialect.supports_empty_insert

trino/sqlalchemy/dialect.py
hashhar: Other dialects also return empty list if no FKs exist.

trino/sqlalchemy/dialect.py
hashhar: Other clients (which don't support 2PC transactions) `raise NotImplementedError()` from this method (and all twophase methods).
Please change to `raise NotImplementedError()`.

dungdm93: @hashhar Yeah, I just remove all twophase methods to leave it to default (`raise NotImplementedError()`)

tests/sqlalchemy/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+            that_attr = this.attr_types[name]
+            assert_sqltype(this_attr, that_attr)
+    else:
+        assert_that(str(this)).is_equal_to(str(that))
```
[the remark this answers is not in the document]
dungdm93: `str` is simplest ways to compare 2 datatypes. It'll return string representation in SQL.

trino/sqlalchemy/datatype.py
dungdm93: It's renamed to `aware_split` to distinguish it from `str.split()`

trino/sqlalchemy/datatype.py
dungdm93: I don't think it's necessary. It's local var and `i` & `j` also very common for index var.

trino/sqlalchemy/datatype.py
dungdm93: #107

trino/sqlalchemy/compiler.py
```diff
@@ -0,0 +1,140 @@
...
+
+class TrinoTypeCompiler(compiler.GenericTypeCompiler):
+    def visit_FLOAT(self, type_, **kw):
+        precision = type_.precision or 32
+        if 0 <= precision <= 32:
+            return self.visit_REAL(type_, **kw)
+        elif 32 < precision <= 64:
+            return self.visit_DOUBLE(type_, **kw)
+        else:
+            raise ValueError(f"type.precision must be in range [0, 64], got {type_.precision}")
```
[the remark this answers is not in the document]
dungdm93: It's because `GenericTypeCompiler` has default impl for `visit_FLOAT`.  
Other behavior is `raise Exception("not support FLOAT data type")`, so which one you prefer?

esemeniuc:
How is the current state of the port? Is it mostly usable?

tests/sqlalchemy/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+            # ARRAY(dimensions=None) == ARRAY(dimensions=1)
+            assert_that(that.dimensions).is_in(None, 1)
+        else:
+            assert_that(this.dimensions).is_equal_to(this.dimensions)
+    elif isinstance(this, MAP):
+        assert_sqltype(this.key_type, that.key_type)
+        assert_sqltype(this.value_type, that.value_type)
+    elif isinstance(this, ROW):
+        assert_that(len(this.attr_types)).is_equal_to(len(that.attr_types))
+        for (this_attr, that_attr) in zip(this.attr_types, that.attr_types):
```
hashhar, line 36: Does this maintain order? i.e. `ROW(a INT, b INT)` != `ROW(b INT, a INT)`.

tests/sqlalchemy/test_datatype_parse.py
```diff
@@ -0,0 +1,180 @@
...
+        actual_type = type(actual_type)
+    assert_that(actual_type).is_equal_to(sql_type)
+
+
+parse_cases_testcases = {
+    'char(10)': CHAR(10),
+    'Char(10)': CHAR(10),
+    'VARCHAR(10)': VARCHAR(10),
+    'varCHAR(10)': VARCHAR(10),
+    'VARchar(10)': VARCHAR(10),
```
hashhar, line 43: Can we test unbounded types too? `CHAR` and `VARCHAR`.

trino/__init__.py
hashhar: cc: @lukasz-walkiewicz any opinions?

trino/dbapi.py
```diff
@@ -454,7 +488,8 @@ def genall(self):
         return self._query.result
 
     def fetchall(self) -> List[List[Any]]:
-        return list(self.genall())
+        res = list(self.genall())
+        return res
```
hashhar, line 492: Why this change?

trino/sqlalchemy/compiler.py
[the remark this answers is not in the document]
hashhar: Discussed offline. The reason it is this way is because SqlAlchemy has FLOATS that can be 64 bits and we are only mapping to REAL or DOUBLE depending on what precision is provided.

trino/sqlalchemy/datatype.py
hashhar: begin and end sound like logical names since that's what they are representing.

tests/sqlalchemy/conftest.py
nineinchnick: It does

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,297 @@
...
+
+    @classmethod
+    def dbapi(cls):
+        """
+        ref: https://www.python.org/dev/peps/pep-0249/#module-interface
+        """
+        return trino_dbapi
+
+    def create_connect_args(self, url: URL) -> Tuple[Sequence[Any], Mapping[str, Any]]:
+        args = list()
```
hovaesco, line 62: why args is needed here?

dungdm93: For return.

trino/dbapi.py
dungdm93: Rollback. I changed it to set break-point while debugging.

trino/sqlalchemy/dialect.py
hovaesco: I see now that it is required return here: https://docs.sqlalchemy.org/en/14/core/internals.html#sqlalchemy.engine.Dialect.create_connect_args

tests/sqlalchemy/test_datatype_parse.py
dungdm93: Added test for unbounded types

hashhar (commented):
Looks mostly good to me. Couple of comments.

@findepi Do you want to take a look?

hashhar:
Also, I can't see any way to logically keep the commits separate - please squash them into a single one.

trino/__init__.py
dungdm93: Maybe, but the fact that all current modules are exposed.  
So the restriction is unnecessary.

trino/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,297 @@
...
+        return dbapi_connection.schema
+
+    def do_execute(self, cursor: Cursor, statement: str, parameters: Tuple[Any, ...],
+                   context: DefaultExecutionContext = None):
+        cursor.execute(statement, parameters)
+        if context and context.should_autocommit:
+            # SQL statement only submitted to Trino server when cursor.fetch*() is called.
+            # For DDL (CREATE/ALTER/DROP) and DML (INSERT/UPDATE/DELETE) statement, call cursor.description
+            # to force submit statement immediately.
+            cursor.description  # noqa
```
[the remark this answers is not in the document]
dungdm93: Because it'll fails in this scenario
```python
cur = connection.execute(text("UPDATE purchases SET status = 'OVERDUE' WHERE ship_date IS NULL;"))

# User may wanna get total number of records updated by call `cur.fetchall()`
# But it'll return null because cursor already fetchall in `do_execute` and ignore the result.
print(f"Total record updated {cur.fetchall()}")
```

tests/sqlalchemy/test_datatype_parse.py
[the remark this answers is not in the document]
dungdm93: `ARRAY(ROW)` added

trino/sqlalchemy/datatype.py
dungdm93: Comment added

dungdm93:
@hashhar updated as your comments.
I also squash all commits and has some feedback.

hashhar:
Merged as https://github.com/trinodb/trino-python-client/commit/da1441f1083ef1e908f4a8d68d2ef80491157944 (and preceding commits). There were some conflicts that I resolved (and split into commits). Your authorship is retained on all commits.

Thanks so much for continuing to work on this @dungdm93 and getting it merged. ❤️  This will be available in the next release which should follow soon.

PR #83, opened 2021-04-04
Drop Python 3.5 support

joshthoward:
This PR includes the following updates:
1. Dropping Python 3.5 support
2. Removing Python 2 type annotations

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [
-          3.5,
           3.6,
           3.7,
           3.8,
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 # Introduction
 
 This package provides a client interface to query [Trino](https://trino.io/)
-a distributed SQL engine. It supports Python>=3.5 and pypy.
+a distributed SQL engine. It supports Python>=3.6 and pypy.
 
 # Installation
 
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,6 @@ setup(
         "Operating System :: Microsoft :: Windows",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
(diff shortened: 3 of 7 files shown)


hashhar:
For removing Python 3.5 you also need to update the setup.py python_requires AFAIK.

joshthoward:
@hashhar disregard Mac OS and Windows support... I need to think about whether we want to run docker on each of those platforms or if we should have another service up and running for integration testing.

hashhar (commented):
I'm not so sure about the f-string changes (others may disagree) - they are readable in some cases but not in others. Also, can IDEs/static analysis tools catch cases where a f-string references a non-existent/mis-spelled variable?

e.g.
```
return "MAP({}, {})".format(
    self._format_prepared_param(keys),
    self._format_prepared_param(values)
)
```

vs

```
return f"MAP({self._format_prepared_param(keys)},
    {self._format_prepared_param(values)})"
```.

Also, are we using mypy in CI to actually verify the type annotations we have?
What end-user behaviour is for people who try to install the package for an unsupported version of Python?

**Review by hashhar — APPROVED** [review]
Looks good.

@joshthoward What does a user trying to install on Python3.5 see? Do get the older version? Or does it silently fail?

**Comment by joshthoward** [comment]
@hashhar It complains if you try to use Python 3.5. See below:

```bash
% pyenv virtualenv 3.5.10 test
Requirement already satisfied: setuptools in /Users/joshthoward/.pyenv/versions/3.5.10/envs/test/lib/python3.5/site-packages
Requirement already satisfied: pip in /Users/joshthoward/.pyenv/versions/3.5.10/envs/test/lib/python3.5/site-packages
% pyenv activate test
% pip install .
Processing /Users/joshthoward/Documents/GitHub/trino-python-client
trino requires Python '>=3.6' but the running Python is 3.5.10
```

hashhar:
Awesome! It's a useful error message. I wondered if it was something cryptic and would make the 3.5 users confused about why installs broke.

PR #86, opened 2021-04-20
Set query cancelled property only after cancelling

joshthoward:
Moved the setter for the cancelled property following the check that the cancellation is successful.


hashhar (approved):
Thanks. ❤️

Issue #69, opened 2021-01-14
Mypy static type checking


joshthoward:
Posting this here to gauge interest...

Does the community think that mypy static type checking would be a benefit? For users, this would enable type hints in modern IDEs. For developers, this would enable yet another check to ensure correctness while building. 

I'm happy to implement this BTW, just wanting to ensure that it would be well received prior to expending the effort. 

Thoughts?

nineinchnick:
There should be no negative side effects and since SQL is strongly typed it makes sense to add type hints to a driver like this :-)

joshthoward:
Part of the challenge is the difference in type annotations for Python 2 and Python 3 ([old Python 2 example is here](https://mypy.readthedocs.io/en/stable/python2.html)). If we would like to enable static type checking, we should consider removing Python 2 support ([see issue here](https://github.com/trinodb/trino-python-client/issues/4)).

pgagnon:
> Part of the challenge is the difference in type annotations for Python 2 and Python 3 ([old Python 2 example is here](https://mypy.readthedocs.io/en/stable/python2.html)). If we would like to enable static type checking, we should consider removing Python 2 support ([see issue here](https://github.com/trinodb/trino-python-client/issues/4)).

If we *must* keep Python 2 support we could put a stub in typeshed instead, but I'd highly suggest removing it anyway as Python 2 is EOL and has been deprecated for over 10 years.

electrum:
As someone who prefers statically typed languages and developing in powerful IDEs, I’m all for this. But as a casual Python user, I don’t know if there are any downsides.

From my quick research, it seems there is the PEP 484 standard for type hints, and then tools like Mypy or Pyright that utilize them. If that’s accurate, I’d like to understand why we’d pick one tool over another, or if that is more of a concern for users of the library.

I also found https://github.com/Instagram/MonkeyType which looks like it might help with adding the annotations.

pgagnon:
> As someone who prefers statically typed languages and developing in powerful IDEs, I’m all for this. But as a casual Python user, I don’t know if there are any downsides.

From a runtime perspective there is no downside as type hints do not alter runtime behavior at all so this is strictly a linting/CI/IDE concern.

> From my quick research, it seems there is the PEP 484 standard for type hints, and then tools like Mypy or Pyright that utilize them. If that’s accurate, I’d like to understand why we’d pick one tool over another, or if that is more of a concern for users of the library.

[...]

PR #91, opened 2021-07-18
Add pre-commit as a Flake8 and MyPy delegate

arturdryomov:
Relates to #69. Also see [`pre-commit`](https://pre-commit.com/).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,21 @@ name: ci
 on: [push, pull_request]
 
 jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout the source code"
+        uses: actions/checkout@v2
+
+      - name: "Install Python"
+        uses: actions/setup-python@v2
+
+      - name: "Install pre-commit"
+        run: pip install pre-commit
+
+      - name: "Run pre-commit checks"
+        run: pre-commit run --all-files
+
   build:
     runs-on: ubuntu-latest
     strategy:
@@ -27,6 +42,3 @@ jobs:
       - name: Run tests
         run: |
           pytest -s tests/ integration_tests/
-      - name: Run linter
-        run: |
-          flake8
--- a/setup.py
+++ b/setup.py
@@ -22,16 +22,16 @@ _version_re = re.compile(r"__version__\s+=\s+(.*)")
 
 
 with open("trino/__init__.py", "rb") as f:
-    version = str(
-        ast.literal_eval(_version_re.search(f.read().decode("utf-8")).group(1))
-    )
+    trino_version = _version_re.search(f.read().decode("utf-8"))
+    assert trino_version is not None
+    version = str(ast.literal_eval(trino_version.group(1)))
 
 
 kerberos_require = ["requests_kerberos"]
 
-all_require = [kerberos_require]
+all_require = kerberos_require + []
 
-tests_require = all_require + ["httpretty", "pytest", "pytest-runner", "pytz", "flake8", "click"]
+tests_require = all_require + ["httpretty", "pytest", "pytest-runner", "pytz", "click"]
 
 setup(
     name="trino",
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,15 @@
+repos:
+  - repo: "https://github.com/PyCQA/flake8"
+    rev: "3.9.2"
+    hooks:
+      - id: "flake8"
+        name: "Python: analysis"
+
+  - repo: "https://github.com/pre-commit/mirrors-mypy"
+    rev: "v0.910"
+    hooks:
+      - id: "mypy"
+        name: "Python: types"
+        additional_dependencies:
+          - "types-pytz"
+          - "types-requests"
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -3,6 +3,18 @@ name: ci
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout the source code"
+        uses: actions/checkout@v2
+
+      - name: "Install Python"
+        uses: actions/setup-python@v2
+
+      - name: "Run pre-commit checks"
+        uses: pre-commit/action@v2.0.3
```
arturdryomov, line 16: A dedicated GitHub action is used here but it’s possible to install `pre-commit` on a local machine and run checks before a commit or at will.

setup.py
```diff
@@ -22,16 +22,19 @@
 
 
 with open("trino/__init__.py", "rb") as f:
```
arturdryomov, line 24: Did a number of changes here and below but please note that this is a minimal config. Enabling a strict mode for MyPy or other options triggers dozens of warnings. These should be addressed in separate changes.

arturdryomov:
@ebyhr, @joshthoward, hey, can I ask you to check the PR and the CLA? I’ve sent the CLA email a week ago but the check is still failing so I assume it wasn’t reviewed.

ebyhr:
@martint Could you make sure CLA email from @arturdryomov ?

arturdryomov:
@ebyhr, the CLA was accepted this night. I’ll have to force-push to trigger the check I guess.

ebyhr:
@cla-bot check

ebyhr:
I can't find your account in contributor list.

> ll have to force-push to trigger the check I guess.

We can trigger the check by above comment.

arturdryomov:
@ebyhr, thanks! Yeah, seems like I was invited into the org but wasn’t added to `trinodb/cla/contributors`. Should I open a PR here or someone else does this on a regular basis?

ebyhr:
@arturdryomov You don't need to open a PR. @martint will add your account to the list.

.github/workflows/ci.yml
hashhar: Any reason to use `pre-commit` directly? Can we instead run all the style/static checks directly within the job (as we did earlier for flake8)?
It has the benefit that people can easily run the checks locally without having to install an additional tool.

+1 for splitting out static analysis and tests to separate jobs though.

arturdryomov: * It serves as an umbrella for such tools. Basically, each tool defines a hook so each consumer doesn’t need to know about invocation details.
* It introduces a clear statement about pre-commit checks (hence the name). This repo doesn’t have `flake8` mentions outside CI so ATM local runs feel a bit like a tribal knowledge.
* It reuses environments / tools across repos since `pre-commit` installs its own versions.
* Since it uses a separate config, such tools do not clutter requirements. As you’ve noted, static analysis is different from tests.

I was under an impression that `pre-commit` is actually quite popular among Python projects. For example, [Django uses it](https://github.com/django/django/blob/main/.pre-commit-config.yaml), [Flask](https://github.com/pallets/flask/blob/main/.pre-commit-config.yaml), [Superset](https://github.com/apache/superset/blob/master/.pre-commit-config.yaml), etc. Also, as an anecdotal evidence, we use it at Lyft across hundreds of repos (not exclusively Python ones).

ebyhr:
@cla-bot check

arturdryomov:
Rebased and the CLA seem to be in order.

@ebyhr, @findepi, PTAL. Please tag related people, I’ve chosen you based on [recent activity](https://github.com/trinodb/trino-python-client/graphs/contributors).

arturdryomov:
@hashhar, @ebyhr, @findepi, 🏓

.github/workflows/ci.yml
thof: FYI it seems the action is already deprecated https://github.com/pre-commit/action#readme

setup.py
```diff
@@ -22,16 +22,19 @@
...
-    version = str(
```
joshthoward, line 25: This seems like unnecessary error handling. We if the version is missing I would want an explicit failure.

joshthoward (changes requested):
Minor change requested. Otherwise, looks good.

.github/workflows/ci.yml
joshthoward: Yeah. IMHO the pre-commit thing looks suspect. The docs indicate that they will not charge for the hook, but https://github.com/marketplace/pre-commit-ci already has a pricing plan attached. This is a shame really.

hashhar: Let's do this the old fashioned way - pre-commit doesn't add much value IMO to be considered just yet.

setup.py
arturdryomov: Switched to `assert`, should fail and highlight the issue if it appears.

thof (approved):
LGTM but the discussion around pre-commit is not resolved yet, right (cc: @hashhar)? I'm fine with the current approach though.

.github/workflows/ci.yml
hashhar: Looks better, it can now be run locally as well by anybody new by looking at the ci config.

hashhar (approved):

.github/workflows/ci.yml
```diff
@@ -3,6 +3,21 @@ name: ci
 on: [push, pull_request]
 
 jobs:
+  pre-commit:
```
hashhar, line 6: Can you give this a name like `lint`? pre-commit is misleading.

hashhar (approved):
LGTM % workflow step name.

Sorry for the long back and forth @arturdryomov . Thanks for working on this.

arturdryomov:
@hashhar, done. Thanks for your help!

Issue #95, opened 2021-08-11
Trino-python-client not returning/capturing External Errors when using fetchone


subbareddydagumati:
Trino Python client not identifying the external errors and exiting successfully. Looks like it's not waiting until the end of transaction or relying on transaction status.

Connector: Hive 
Trino Version: 356
Trino-python Client: 0.305.0

Steps to Reproduce:
1) Create stage Table without partition
```
CREATE TABLE dl_prod.prod_tmp.trino_python_issue_stg (
id int,
c_name varchar,
file_date varchar
)
WITH (
   format = 'ORC',
        external_location='hdfs://datalakeprod/data/dl/test/trino_python_issue_stg/'
)
```

2) Insert Data into Stage Table
```
INSERT INTO dl_prod.prod_tmp.trino_python_issue_stg 
(id,c_name,file_date) VALUES(1,'subba','20210811')
,(2,'subba','20210812')

```
3) Create another table with partitions
```
CREATE TABLE dl_prod.prod_tmp.trino_python_issue (
id int,
c_name varchar,
file_date varchar
)
WITH (
   format = 'ORC',
      external_location='hdfs://datalakeprod/data/dl/test/trino_python_issue/',
       partitioned_by = ARRAY['file_date'] 
)
```

4) Enabled Overwrite
`set session dl_prod.insert_existing_partitions_behavior = 'overwrite'`

5) Insert Data into Final Table
```
INSERT INTO  dl_prod.prod_tmp.trino_python_issue
SELECT * FROM  dl_prod.prod_tmp.trino_python_issue_stg 
```

6)Drop Final Table
`DROP TABLE prod_tmp.trino_python_issue`

7) Re-Create the final Table
```
CREATE TABLE dl_prod.prod_tmp.trino_python_issue (
id int,
c_name varchar,
file_date varchar
)
WITH (
   format = 'ORC',
      external_location='hdfs://datalakeprod/data/dl/test/trino_python_issue/',
       partitioned_by = ARRAY['file_date'] 
)
```

8) Run the below Insert statement. It will fail with external Error
```
INSERT INTO  dl_prod.prod_tmp.trino_python_issue
SELECT * FROM  dl_prod.prod_tmp.trino_python_issue_stg 
```
Error in DBeaver:
[...]

PR #96, opened 2021-08-16
Minor improvement to README

ebyhr:


hashhar (approved):
LGTM.

We should fix #95 eventually.

PR #97, opened 2021-08-17
Add type annotations and tests for transaction isolation level

arturdryomov:


trino/transaction.py
```diff
@@ -33,15 +34,15 @@ class IsolationLevel(object):
     SERIALIZABLE = 4
 
     @classmethod
-    def levels(cls):
-        return {k for k in cls.__dict__.keys() if not k.startswith("_")}
+    def levels(cls) -> Iterable[str]:
+        return {k for k, v in cls.__dict__.items() if not k.startswith("_") and isinstance(v, int)}
```
arturdryomov, line 38: Without this check `levels`, `values` and `check` are returned as well.

tests/test_transaction.py
```diff
@@ -0,0 +1,31 @@
+from trino.transaction import IsolationLevel
```
ebyhr, line 1: Add Apache license header.

ebyhr (approved):
Looks good except for a minor comment.

tests/test_transaction.py
arturdryomov: Done, thanks!

ebyhr:
Merged, thanks!

PR #98, opened 2021-08-17
Use Trino JSON response as test example

ebyhr:
Used simple SELECT query to generate the stacktrace.
```
trino> select * from orders;
Query 20210817_140827_00000_arvdv failed: line 1:15: Schema must be specified when session schema is not set
io.trino.spi.TrinoException: line 1:15: Schema must be specified when session schema is not set
	at io.trino.sql.analyzer.SemanticExceptions.semanticException(SemanticExceptions.java:48)
	at io.trino.sql.analyzer.SemanticExceptions.semanticException(SemanticExceptions.java:43)
	at io.trino.metadata.MetadataUtil.lambda$createQualifiedObjectName$3(MetadataUtil.java:152)
...
```

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -44,10 +44,10 @@ To get some HTTP response, set logging level to DEBUG with
 
 """
 RESP_DATA_POST_0 = {
-    "nextUri": "coordinator:8080/v1/statement/20161115_222658_00040_xtnym/1",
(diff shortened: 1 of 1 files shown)


hashhar (approved):
LGTM.

This is a good reason to invest in some integration tests instead of mocks eventually cc: @aalbu

PR #100, opened 2021-08-19
Add type annotations and tests for query error

arturdryomov:


trino/client.py
```diff
@@ -501,7 +501,7 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
         call fetch() until finished is true.
         """
         if self.cancelled:
-            raise exceptions.TrinoUserError("Query has been cancelled", self.query_id)
+            raise exceptions.TrinoUserError({"message": "Query has been cancelled"}, self.query_id)
```
arturdryomov, line 504: `TrinoUserError` inherits `TrinoQueryError` which expects `dict`.

trino/exceptions.py
```diff
@@ -42,44 +42,53 @@ class TimeoutError(Exception):
...
-    def error_location(self):
-        location = self._error["errorLocation"]
-        return (location["lineNumber"], location["columnNumber"])
+    def error_location(self) -> typing.Optional[typing.Tuple[int, int]]:
+        location = self._error.get("errorLocation", None)
+
+        if location is None:
+            return None
+        else:
+            return location["lineNumber"], location["columnNumber"]
```
arturdryomov, line 85: Without this change the call raises an exception when there is no error, attempting to resolve `errorLocation` from a blank `dict`.

tests/test_exceptions.py
```diff
@@ -72,3 +74,51 @@ def __call__(self):
     with_retry(FailerUntil(2).__call__)()
     with pytest.raises(SomeException):
         with_retry(FailerUntil(3).__call__)()
+
+
+def test_trino_query_error_available():
```
ebyhr, line 79: Are you aware that test_client.py has similar test (`test_trino_fetch_error`)?

tests/test_exceptions.py
```diff
@@ -72,3 +74,51 @@ def __call__(self):
...
+    assert error.query_id == "fake_ID"
+
+    assert repr(error) == """TrinoQueryError(type=fake_type, name=fake_name, message="fake_message", query_id=fake_ID)"""  # noqa: E501
+    assert str(error) == repr(error)
+
+
+def test_trino_query_error_unavailable():
+    error = TrinoQueryError(
+        error=dict(),
+        query_id="fake_ID",
```
ebyhr, line 112: nit: I would change the value to `test_query_id`.

tests/test_exceptions.py
arturdryomov: It kind of is but it doesn’t check all fields and serves as a semi-integration one. Should I reuse JSON from there or leave this change as is?

ebyhr: Could you rename this test method? Maybe, `test_trino_query_syntax_error` or something.

ebyhr (commented):
Sorry for my late response, I forgot to submit a button.

hashhar (commented):
Looks good % existing comments.

If possible can you split into two commits - one that introduces tests and the next one that adds annotations?

PR #105, opened 2021-08-26
Fix bug when the query result is returned in first HTTP call

buremba:
If Trino returns the query result in the first HTTP call, the `columns` property is not set so the library throws the` Cursor must be loaded before call TrinoResultMetaData` exception.

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -507,20 +507,23 @@ class TrinoQuery(object):
         status = self._request.process(response)
         self.query_id = status.id
         self._stats.update({"queryId": self.query_id})
-        self._stats.update(status.stats)
+        self._update_state(status)
         self._warnings = getattr(status, "warnings", [])
         if status.next_uri is None:
             self._finished = True
         self._result = TrinoResult(self, status.rows)
         return self._result
 
+    def _update_state(self, status):
+        self._stats.update(status.stats)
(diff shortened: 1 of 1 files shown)


ebyhr:
@cla-bot check

hashhar:
When does Trino return the result in the first request itself? I can't think of any case except where the query planning/analysis/scheduling fails. Just want to understand what was the user-facing behaviour.

hashhar (commented):
The fix looks correct but I don't understand how the original error was being hit.

trino/client.py
```diff
@@ -507,20 +507,23 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
...
         self._stats.update({"queryId": self.query_id})
-        self._stats.update(status.stats)
+        self._update_stats(status)
         self._warnings = getattr(status, "warnings", [])
         if status.next_uri is None:
             self._finished = True
         self._result = TrinoResult(self, status.rows)
         return self._result
+    
+    def _update_stats(self, status):
```
findepi, line 517: rename function, since it does more than just update stats

hashhar:
@buremba Can you please squash all the commits (since it's a single logical change)?

hashhar (approved):
Thanks. Merged.

PR #109, opened 2021-09-13
Enable HTTPS when using port 443

### LINKED ISSUE #94: Enable HTTPS automatically if using port 443
Wwe should make the Python client default to HTTPS if using port 443. While it's technically a backwards incompatible change, it seems unlikely that anyone is using HTTP over the standard HTTPS port. If they are, they can specify `http_scheme='http'` explicitly.

This will make the Python client consistent with the JDBC driver in this regard. Similar issue for the CLI: https://github.com/trinodb/trino/issues/8798

shawnzhu:
Fixes #94


tests/test_client.py
```diff
@@ -402,6 +403,25 @@ def test_request_invalid_http_headers():
     assert str(value_error.value).startswith("cannot override reserved HTTP header")
 
 
+def test_enabling_https(monkeypatch):
```
hashhar, line 406: Let's test all of the following forms:

1. `server:443` => https.
2. `https://server:443` => https
3. `http://server:443` => http (since scheme is specified in the URL)

shawnzhu: added in https://github.com/trinodb/trino-python-client/pull/109/commits/306ac9b675d5283f5d9288fe98b47dc3e808b5e0

hashhar:
Looks good to me. Please squash the commits @shawnzhu with a message like (so that it has all context about the change and why it was done):

```
Update client to default protocol to HTTPS if port is 443

If a user passes in a port that is 443, but they do not pass in
a protocol, we can assume that the user is using the HTTPS protocol.

This change makes the Python client consistent with the behaviour of the JDBC driver.
```

@joshthoward PTAL if you can. (I don't like the boilerplate in the tests but don't know a better way to do it).

tests/test_client.py
```diff
@@ -402,6 +403,67 @@ def test_request_invalid_http_headers():
...
+    post_recorder = ArgumentsRecorder()
+    monkeypatch.setattr(TrinoRequest.http.Session, "post", post_recorder)
+
+    req = TrinoRequest(
+        host="coordinator",
+        port=constants.DEFAULT_TLS_PORT,
+        user="test",
+        catalog="test",
+        schema="test",
+        session_properties={},
```
ebyhr, line 416: Let's remove unnecessary arguments and use literal because actual users won't use the constant. 
```python
    req = TrinoRequest(
        host="coordinator",
        port=443,
        user="test",
    )
```

tests/test_client.py
```diff
@@ -402,6 +403,67 @@ def test_request_invalid_http_headers():
...
+
+    req = TrinoRequest(
+        host="coordinator",
+        port=constants.DEFAULT_TLS_PORT,
+        user="test",
+        catalog="test",
+        schema="test",
+        http_scheme=constants.HTTPS,
+        session_properties={},
+    )
```
ebyhr, line 437:
```python
    req = TrinoRequest(
        host="coordinator",
        port=443,
        user="test",
        http_scheme='https',
    )
```

ebyhr (approved):

PR #114, opened 2021-10-10
Add precision datetime support

### LINKED ISSUE #42: Support variable precision time, timestamp, time with time zone, timestamp with time zone types
Over couple of recent releases (up to including Presto 341), Presto gained support for variable precision `time`, `timestamp`, `time with time zone`, `timestamp with time zone` types.
For backward compatibility reasons, these are not rendered with actual precision to the client, unless the client sends proper `X-Presto-Client-Capabilities` header.

Add support for these, along with properly setting the header.

See more at https://github.com/prestosql/presto/issues/1284

andreyors:
closes #42


trino/constants.py
```diff
@@ -43,3 +43,6 @@
 HEADER_PREPARED_STATEMENT = 'X-Trino-Prepared-Statement'
 HEADER_ADDED_PREPARE = 'X-Trino-Added-Prepare'
 HEADER_DEALLOCATED_PREPARE = 'X-Trino-Deallocated-Prepare'
+
+HEADER_TRINO_CLIENT_CAPABILITIES = 'X-Trino-Client-Capabilities'
```
andreyors, line 47: I've tried `X-Presto-Client-Capabilities`, it seems the field was renamed

integration_tests/test_dbapi.py
```diff
@@ -106,22 +106,6 @@ def test_string_query_param(trino_connection):
     assert rows[0][0] == "six'"
 
 
-def test_datetime_query_param(trino_connection):
```
andreyors, line 109: I've added extended test case scenario at the bottom - with and without precision

integration_tests/test_dbapi.py
```diff
@@ -394,3 +378,35 @@ def test_eager_loading_cursor_description(trino_connection):
     assert description_after is not None
     assert len(description_after) == len(description_expected)
     assert all([a == e] for a, e in zip(description_after, description_expected))
+
+
+@pytest.mark.parametrize(
+    'parametric_datetime, datetime, expected_time, description',
+    (
+        (False, datetime(2020, 1, 1, 0, 0, 0, 0), "2020-01-01 00:00:00.000", "timestamp with time zone"),
```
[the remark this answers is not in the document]
andreyors: it is the same as False [here](https://github.com/trinodb/trino-python-client/pull/114/files#diff-cf9b945b8cb541d38680c39caa2de6568793a6d53644d6f2659de6a77ff96127R291), do we want to check if python recognizes False as None?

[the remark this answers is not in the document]
ebyhr: The purpose is verifying the default behavior, not there.

integration_tests/test_dbapi.py
```diff
@@ -106,20 +106,136 @@ def test_string_query_param(trino_connection):
...
+
+    cur.execute(f"SELECT TIME '{time}'")
+    rows = cur.fetchall()
+
+    assert rows[0][0] == time
+
+    cur.execute(f"SELECT TIME '{time} +00:00'")
+    rows = cur.fetchall()
+
+    assert rows[0][0] == f"{time}+00:00"
```
ebyhr, line 171: It would be better to move to test parameter. 
```python
@pytest.mark.parametrize(
    'parametric_datetime, time, description',
    (
        (True, "00:00:00.100001", "time(6)"),
        (False, "00:00:00.000", "time"),
        (None, "00:00:00.000", "time"),
        (True, "00:00:00.100001+01:00", "time(6) with time zone"),
        (False, "00:00:00.000+02:00", "timestamp with time zone"),
        (None, "00:00:00.000+03:00", "timestamp with time zone"),
    )
)
def test_time_query_param(run_trino, parametric_datetime, time, description):
    _, host, port = run_trino

    connection = trino.dbapi.Connection(
        host=host,
        port=port,
        user="test",
        source="test",
        parametric_datetime=parametric_datetime
    )

    cur = connection.cursor()

    cur.execute(f"SELECT TIME '{time}'")
    rows = cur.fetchall()

    assert rows[0][0] == time
    assert cur.description[0][1] == description
```

Subsequent `test_timestamp_query_param` and `test_interval_query_param` also.

PR #124, opened 2021-11-15
Reuse the same connection when getting a token

lukasz-walkiewicz:
The benefits of this approach are:
* there's no need for retrying 401 (Unauthorized) as the same connection
  is being reused,
* the whole authentication uses the same context (PROXIES, SSL verify
  etc.) so no need for client's setup duplication.
Additionally, OAuth2 client's tests were rewritten using HTTPretty which
stubs/mocks the http layer and makes changing implementation details
easier.

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -9,20 +9,25 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import json
+import re
+import threading
+import time
+import uuid
+from collections import namedtuple
+from unittest import mock
+from urllib.parse import urlparse
 
 import httpretty
 import pytest
 import requests
-import time
-from unittest import mock
-from urllib.parse import urlparse
-
-
+from httpretty import httprettified
 from requests_kerberos.exceptions import KerberosExchangeError
-from trino.client import TrinoQuery, TrinoRequest, TrinoResult
-from trino.auth import KerberosAuthentication, _OAuth2TokenBearer
-from trino import constants
+
 import trino.exceptions
+from trino import constants
+from trino.auth import KerberosAuthentication, _OAuth2TokenBearer
+from trino.client import TrinoQuery, TrinoRequest, TrinoResult
 
 
 @mock.patch("trino.client.TrinoRequest.http")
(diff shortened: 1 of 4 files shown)


trino/auth.py
```diff
@@ -142,101 +143,77 @@ class _OAuth2TokenBearer(AuthBase):
...
 
-    class _AuthStep(Enum):
-        GET_REDIRECT_SERVER = 0
-        GET_TOKEN = 1
-
     def __init__(self, http_session, redirect_auth_url_handler=handle_redirect_auth_url):
         self._redirect_auth_url = redirect_auth_url_handler
-        self.http_session = http_session
         self._thread_local = threading.local()
+        self._pattern = re.compile(r'Bearer x_redirect_server="(.+)", x_token_server="(.+)"')
```
hashhar, line 149: What if order is different (or lowercase/uppercase)? I think the older approach of split was more robust.

trino/auth.py
```diff
@@ -142,101 +143,77 @@ class _OAuth2TokenBearer(AuthBase):
...
-                    self._thread_local.auth_step = self._AuthStep.GET_REDIRECT_SERVER
-                    return r
-                error = token_response.get('error')
-                if error:
-                    raise exceptions.TrinoAuthError(f"Error while getting the token : {error}")
+        self._redirect_auth_url(redirect_server)
+
+        # Consume content and release the original connection
+        # to allow our new request to reuse the same one.
+        response.content
```
hashhar, line 181: Is consuming the content required? The docs don't mention it explicitly.

hashhar (commented):
Skimmed. Will take another look.

trino/auth.py
lukasz-walkiewicz: The previous approach didn't handle not matching headers well. I've reverted to the old split method but added an additional check.

trino/auth.py
lukasz-walkiewicz: Right... that's a good point. I wasn't sure about it either but on the other hand [`HTTPDigestAuth`](https://github.com/psf/requests/blob/b0e025ade7ed30ed53ab61f542779af7e024932e/requests/auth.py#L261) handles reusing connections this way and this implementation is somewhat influenced by it.

trino/auth.py
```diff
@@ -142,55 +143,40 @@ class _OAuth2TokenBearer(AuthBase):
...
 
         # we have to handle the authentication, may be token the token expired or it wasn't there at all
-        auth_info = r.headers.get('WWW-Authenticate')
+        auth_info = response.headers.get('WWW-Authenticate')
         if not auth_info:
             raise exceptions.TrinoAuthError("Error: header WWW-Authenticate not available in the response.")
 
-        matches = dict((s.strip().split('=', 1) for s in auth_info.split(",")))
-        if matches is None:
+        parts = list((s.strip().split('=', 1) for s in auth_info.split(",")))
```
hashhar, line 169: Does this solve https://trinodb.slack.com/archives/CFPVDCDHV/p1637095049019600?

Can we test for it?

trino/auth.py
```diff
@@ -142,55 +143,40 @@ class _OAuth2TokenBearer(AuthBase):
...
             raise exceptions.TrinoAuthError(f"Error: header info didn't match {auth_info}")
 
+        matches = dict(parts)
+
         auth_server = matches.get('Bearer x_redirect_server')
         if auth_server is None:
             raise exceptions.TrinoAuthError("Error: header info didn't have Bearer x_redirect_server")
 
-        token_server = matches.get('x_token_server')
+        token_server = matches.get('x_token_server').strip('"')
```
hashhar, line 179: Does this need to be done to `auth_server` too?

hashhar (commented):
Looks good % comments.

trino/auth.py
lukasz-walkiewicz: Sure, I'll add a test case.

trino/auth.py
lukasz-walkiewicz: Good catch! It was redundant

hashhar (approved):
Looks good to me.

I'll appreciate someone else skimming this too though. cc: @ebyhr @joshthoward ?

lukasz-walkiewicz:
Thanks, commits squashed.

trino/auth.py
```diff
@@ -141,102 +143,86 @@ class _OAuth2TokenBearer(AuthBase):
...
         if auth_server is None:
-            raise exceptions.TrinoAuthError("Error: header info didn't have Bearer x_redirect_server")
+            raise exceptions.TrinoAuthError("Error: header info didn't have x_redirect_server")
 
         token_server = matches.get('x_token_server')
         if token_server is None:
             raise exceptions.TrinoAuthError("Error: header info didn't have x_token_server")
 
-        self._thread_local.token_server = token_server.strip('"')
+        self._thread_local.token_server = token_server
```
losipiuk, line 183: General comment for the future. A single comment is changing logic and doing cleanup refactors. Would be nice to split the two, so review is easier.

trino/auth.py
```diff
@@ -141,102 +143,86 @@ class _OAuth2TokenBearer(AuthBase):
...
+
+        request = response.request.copy()
+        extract_cookies_to_jar(request._cookies, response.request, response.raw)
+        request.prepare_cookies(request._cookies)
+
+        request.headers['Authorization'] = "Bearer " + self._thread_local.token
+        retry_response = response.connection.send(request, **kwargs)
+        retry_response.history.append(response)
+        retry_response.request = request
+        return retry_response
```
losipiuk, line 203: nit: could extract `retry_request` method

losipiuk (approved):
I skimmed through the production code - LGTM

tests/unit/test_client.py
```diff
@@ -254,225 +259,360 @@ def long_call(request, uri, headers):
...
+def test_oauth2_authentication_flow(attempts, sample_post_response_data):
+    token = str(uuid.uuid4())
+    challenge_id = str(uuid.uuid4())
 
-    with pytest.raises(trino.exceptions.TrinoAuthError) as exp:
-        oauth_test = OAuthTestReq("get", OAUTH_SERVER_URL_FORCE_FAIL)
-        token_req = oauth_test.get_token_request()
-        oauth(token_req)
+    redirect_server = f"https://coordinator/oauth2/initiate/{challenge_id}"
+    token_server = f"https://coordinator/oauth2/token/{challenge_id}"
```
hovaesco, line 308: extract `redirect_server` and `token_server` to constants? those are repeated in a couple places.

hovaesco (approved):
LGTM, all comments are nitpicks

lukasz-walkiewicz:
Comments addressed, please take yet another look.

hovaesco (approved):

hashhar (approved):

lukasz-walkiewicz:
Thanks!

PR #126, opened 2021-11-18
Expose info_uri from query status in Cursor

### LINKED ISSUE #118: Info URI should be expose as part of response from query processing so that It can be tracked on UI directly.
Mostly Helpful for long-running queries where the user wants to keep track of the overall query performance and where it's stuck exactly.

chyzzqo2:
Fixes #118

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -393,3 +393,14 @@ def test_eager_loading_cursor_description(trino_connection):
     assert description_after is not None
     assert len(description_after) == len(description_expected)
     assert all([a == e] for a, e in zip(description_after, description_expected))
+
+
+def test_info_uri(trino_connection):
+    cur = trino_connection.cursor()
+    assert cur.info_uri is None
+    cur.execute('SELECT * FROM system.runtime.nodes')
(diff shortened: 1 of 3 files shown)


ebyhr (approved):

tests/integration/test_dbapi_integration.py
```diff
@@ -393,3 +393,11 @@ def test_eager_loading_cursor_description(trino_connection):
...
     assert len(description_after) == len(description_expected)
     assert all([a == e] for a, e in zip(description_after, description_expected))
+
+
+def test_info_uri(trino_connection):
+    cur = trino_connection.cursor()
+    assert cur.info_uri is None
+    cur.execute('SELECT * FROM system.runtime.nodes')
+    assert cur.info_uri is not None
+    assert cur._query.query_id in cur.info_uri
```
hashhar, line 403: Can you also add something like below too at the end to verify that the info_uri is available throughout the lifetime of a query?

```python
cur.fetchall()
assert cur.info_uri is not None
assert cur._query.query_id in cur.info_uri
```

hashhar (approved):
LGTM % comment

hashhar (approved):

PR #132, opened 2021-12-03
Fix broken module import

hashhar:
This reverts commit f9e68da9bcf532f50aff544a13ebb54af365b329.
This change broke the README code snippets with an error like "Cannot find reference 'dbapi' in '__init__.py'.

### DIFF
--- a/trino/__init__.py
+++ b/trino/__init__.py
@@ -10,4 +10,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from . import auth
+from . import dbapi
+from . import client
+from . import constants
+from . import exceptions
+from . import logging
+
+__all__ = ['auth', 'dbapi', 'client', 'constants', 'exceptions', 'logging']
+
 __version__ = "0.307.0"


hashhar:
Nope, still doesn't work. Fails with a different error since it tries importing sqlalchemy but it's not installed by default.

hashhar:
@ebyhr PTAL now. The README works for me now.

It still doesn't import sqlalchemy by default now so people using it still need to import the module manually (expected since it requires an additional dependency).

ebyhr (approved):

hashhar:
The alternative is updating the instructions to `import trino.dbapi` but that would break lots of user code.

I verified this works by doing `pip install 'git+https://github.com/hashhar/trino-python-client.git@2982da9a39d50d784e73bb1a11c340847a4e9058'` followed by running the README steps against Trino in docker.

hashhar:
And the integration tests didn't catch it because of 3ce46fe317da1e01ad7d41970415bdab5db025fe

I'll revert that too to make sure current tests test the actual scenario.

findepi (approved):

PR #134, opened 2021-12-03
Revert "Use more specific module imports in tests"

hashhar:
This reverts commit 3ce46fe317da1e01ad7d41970415bdab5db025fe.
The user-facing API (and documentation) is to import the trino package
instead of the specific trino.dbapi module. Integration tests should
test using the import that is documented and part of API.

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -15,7 +15,7 @@ from datetime import datetime
 import pytest
 import pytz
 
-import trino.dbapi
+import trino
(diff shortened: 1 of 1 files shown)


dungdm93:
@hashhar to keep backward compatible with example in README.
You only need `from . import auth` in `trino/__init__.py`

Because this line can break/waning when `import trino.sqlalchemy`
```
__all__ = ['auth', 'dbapi', 'client', 'constants', 'exceptions', 'logging']
```

ebyhr (approved):

PR #146, opened 2021-12-21
Add release pipeline

### LINKED ISSUE #145: Automate release process
GitHub Actions used in dbt-trino looks useful.
https://github.com/starburstdata/dbt-trino/blob/master/.github/workflows/release.yml

hovaesco:
Resolves #145

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,57 @@
+name: trino-python-client release
+
+on:
+  workflow_dispatch:
+
+env:
+  PYTHON_VERSION: "3.9.5"
+
+jobs:
+  release:
+    name: trino-python-client release
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{env.PYTHON_VERSION}}
+
+      - name: Configure Git user
+        run: |
+          git config user.email "releases@trinodb.io"
+          git config user.name "Trino release automation"
+
+      - name: Set release version
+        run: |
+          echo "VERSION_NUMBER=$(./setup.py --version)" >> $GITHUB_ENV
+          echo "NEXT_VERSION_NUMBER=$(echo ${{ env.VERSION_NUMBER }} | awk -F. -v OFS=. '{$2 += 1 ; print}')" >> $GITHUB_ENV
+          echo Next version number: ${{ env.NEXT_VERSION_NUMBER }}
+          sed -i 's/${{ env.VERSION_NUMBER }}/${{ env.NEXT_VERSION_NUMBER }}/' trino/__init__.py
+
+      - name: Prepare release
+        run: |
+          git diff
+          git commit -am "Bump version to ${{ env.NEXT_VERSION_NUMBER }}"
+          git tag -m "" "${{ env.NEXT_VERSION_NUMBER }}"
+
+      - name: Test release
+        run: |
+          python3 -m venv .venv
+          . .venv/bin/activate
+          pip install twine wheel setuptools
+          ./setup.py sdist bdist_wheel
+          pip install dist/trino-${{env.NEXT_VERSION_NUMBER}}.tar.gz
+          pip install dist/trino-${{env.NEXT_VERSION_NUMBER}}-py3-none-any.whl
+          twine check dist/trino-${{env.NEXT_VERSION_NUMBER}}-py3-none-any.whl dist/trino-${{env.NEXT_VERSION_NUMBER}}.tar.gz
+
+      - name: Release to PyPI
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+        run: |
+          twine upload --non-interactive dist/trino-${{env.NEXT_VERSION_NUMBER}}-py3-none-any.whl dist/trino-${{env.NEXT_VERSION_NUMBER}}.tar.gz
+
+      - name: Finalize release
+        run: |  
+          git push origin master "${{ env.NEXT_VERSION_NUMBER }}""


.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Get trino-python-client version
+        run: |
+          pip install setuptools
+          echo "VERSION_NUMBER=$(./setup.py --version)" >> $GITHUB_ENV
+
+      - name: Release to pypi
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
```
hovaesco, line 93: `PYPI_TOKEN` needs to be generated from the PyPi account which has access to publish artifacts.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Create GitHub release
+        uses: actions/create-release@v1
+        env:
+          # This token is provided by Actions, you do not need to create your own token
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{env.VERSION_NUMBER}}
+          release_name: ${{env.VERSION_NUMBER}}
+          prerelease: ${{ steps.release_type.outputs.isPrerelease }}
+          body: |
```
hovaesco, line 65: Not sure whether it is acceptable body of the release, currently there are no releases in GitHub.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Get trino-python-client version
+        run: |
+          pip install setuptools
+          echo "VERSION_NUMBER=$(./setup.py --version)" >> $GITHUB_ENV
+
+      - name: Find release type
+        id: release_type
+        env:
+          IS_PRERELEASE: ${{ contains(env.VERSION_NUMBER, 'rc') ||  contains(env.VERSION_NUMBER, 'b') }}
```
hovaesco, line 53: Don't know if we want to do release candidates in the future.

.github/workflows/release.yml
hovaesco: More info can be added here as: commits between releases etc.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+          python3 -m venv .venv
+          . .venv/bin/activate
+          pip install .
+          pip install twine wheel setuptools
+          ./setup.py sdist bdist_wheel
+          pip install dist/trino-*.tar.gz
+          pip install dist/trino-*-py3-none-any.whl
+          twine check dist/trino-*-py3-none-any.whl dist/trino-*.tar.gz
+
+  github-release:
```
hashhar, line 32: None of the other Trino projects do releases on GitHub since there are no artifacts we'd want to provide via GitHub.

In this case the "GitHub release" will only contain the tarball of the source code, no actual artifacts and is hence redundant in my opinion. Maybe @electrum thinks otherwise?

.github/workflows/release.yml
hashhar: cc: @findepi / @martint to add token to GitHub secrets.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Test release
+        run: |
+          python3 -m venv .venv
+          . .venv/bin/activate
+          pip install .
+          pip install twine wheel setuptools
+          ./setup.py sdist bdist_wheel
+          pip install dist/trino-*.tar.gz
+          pip install dist/trino-*-py3-none-any.whl
+          twine check dist/trino-*-py3-none-any.whl dist/trino-*.tar.gz
```
hashhar, line 30: This is a very surface level test and doesn't verify if the package works.

We should consider some form of "product-tests" here too (a follow up is fine).

hashhar (commented):
Looks good % some dry-runs and GitHub release being redundant/not useful.

I believe that PyPi does allow staging releases - we should use it for the `test` job and then install from there and verify if the package works.

.github/workflows/release.yml
hovaesco: We could run all tests here by executing `tox` instead of doing package checks (or run tests and then do package checks).

hovaesco: > I believe that PyPi does allow staging releases - we should use it for the test job and then install from there and verify if the package works.

That would be some form of `product-test` which you mentioned. 

Proposed flow for `test` job:
1. Package checks
2. Stage release to PyPi
3. Install from PyPi
4. Run all tests by running `tox`

findepi: > I believe that PyPi does allow staging releases -

you mean https://test.pypi.org/ ?

.github/workflows/release.yml
hovaesco: Since none of the other Trino projects do releases on GitHub, I dropped a whole job which was doing that.

hovaesco:
Bumping a version is one thing which is non automated and requires manual work. Can it still be done manually, or do you prefer to have it automated? If yes, we would need to have a CI user which will create bump commits.

hovaesco:
Discussed offline with @hashhar and I added an automation in release workflow which increments a version of trino-python-client and push bump commit before doing a release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,95 @@
+name: trino-python-client release
+
+on:
+  workflow_dispatch:
+
+env:
+  PYTHON_VERSION: "3.9.5"
```
hashhar, line 7: oddly specific version - any specific reason?

.github/workflows/release.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Configure Git user
+        run: |
+          git config user.email "releases@trinodb.io"
+          git config user.name "Trino release automation"
+
+      - name: Set release version
+        run: |
+          echo "VERSION_NUMBER=$(./setup.py --version)" >> $GITHUB_ENV
+          echo "NEXT_VERSION_NUMBER=$(echo ${{ env.VERSION_NUMBER }} | awk -F. -v OFS=. '{$2 += 1 ; print}')" >> $GITHUB_ENV
```
hashhar, line 33: If changing version numbering scheme to something like `317` instead of `0.317.0` helps with this we can do that as well probably. It seems Trino moved from 0.x numbers to plain x but the Python client never followed.

cc: @martint @findepi

.github/workflows/release.yml
```diff
@@ -0,0 +1,95 @@
...
+          echo Next version number: ${{ env.NEXT_VERSION_NUMBER }}
+          sed -i 's/${{ env.VERSION_NUMBER }}/${{ env.NEXT_VERSION_NUMBER }}/' trino/__init__.py
+
+      - name: Push commit
+        run: |
+          git diff
+          git commit -am "Bump version to ${{ env.NEXT_VERSION_NUMBER }}"
+          git tag -m "" "${{ env.NEXT_VERSION_NUMBER }}"
+          git push origin master "${{ env.NEXT_VERSION_NUMBER }}""
```
hashhar, line 42: the commit shouldn't be pushed until we actually finish the release IMO.

hashhar (commented):
some comments.

combining the jobs as steps in a single job might actually reduce some repetition and allow making the process atomic (and hence retryable).

.github/workflows/release.yml
findepi: Note that Python client never used Trino version numbering, only something similar.
I vaguely remember PyPI could be requiring "semver"-like, 3-part versions.
I may be wrong though, and also things could have changed.
From "product" perspective, I am fine with versioning Python client  with just one number (the major version).

Or, we can append ".0.0", if that's needed.
We had "0." at the front, because that's what Trino was using (back when it was Presto)

hovaesco: Based on https://peps.python.org/pep-0440/#final-releases it looks like only 2 and 3 parts semver version is supported. As an alternative, trino-python-client keeps existing versioning and switch to semantic versioning and release 1.0.0 once we agree that a client is well-supported, mature and feature rich.

.github/workflows/release.yml
hovaesco: Just wanted to hardcode some version to avoid breaking release workflow in the future.

hovaesco:
Workflow structure changed, I converted it to a single job workflow with multiple steps.

hovaesco:
@hashhar PTAL

.github/workflows/release.yml
hashhar: It supports one or more parts - single part versions also work.

hovaesco: So what are the next steps here? I don't really like singe part version number, it doesn't follow best practises in Python packages versioning.

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          git diff
+          git commit -am "Bump version to ${{ env.NEXT_VERSION_NUMBER }}"
+          git tag -m "" "${{ env.NEXT_VERSION_NUMBER }}"
+
+      - name: Test release
+        run: |
+          python3 -m venv .venv
+          . .venv/bin/activate
+          pip install .
```
hashhar, line 45: why install `.`? Don't we want to install from the wheel instead?

hashhar (approved):
LGTM % comments.

hashhar (approved):
@findepi can you please remind me if the PYPI_TOKEN secret is set in GHA already or not.

LGTM.

findepi:
> @findepi can you please remind me if the PYPI_TOKEN secret is set in GHA already or not.

it is

Issue #129, opened 2021-11-24
possible bug in start_transaction (dbapi.py)


bachng2017:
## Issue: 
`start_transaction()` (dbapi.py) might need to set the isolation level properly.

## Details: 
Currently the start_transaction does not change the isolation level (#b), while cursor() only compare the isolation level with AUTO_COMMIT(#a). The result is that the transaction and the query uses different `request instance` and run `independently` (perhaps)

- a. https://github.com/trinodb/trino-python-client/blob/master/trino/dbapi.py#L173
- b. https://github.com/trinodb/trino-python-client/blob/c8144dec2d6d44937a7aa1ba295d82b286b292c6/trino/dbapi.py#L130

For some application (likes dbt),  the following sequence is heavily used:
```
1. start_transaction()
2. query
3. commit() or rollback()
```
If the query finishes within the transaction-timeout, everything looks fine. If the query succeed but took a longer time than transaction-timeout (default is 5 min), error will happen at 3 likes below. Note that in this case, the query succeed and the result is recorded (because it seems to run with auto_commit enabled sessions)
```
trino.exceptions.DatabaseError: failed to commit transaction 73dc7443-ec53-4911-95be-a372792743b4: TrinoUserError(type=USER_ERROR, name=UNKNOWN_TRANSACTION, message="Unknown transaction ID: 73dc7443-ec53-4911-95be-a372792743b4. Possibly expired? Commands ignored until end of transaction block", query_id=20211121_145311_00013_7rttf)
```

## Expectation:
It might depend on the spec or definition of transaction-timeout, but expectation is that there is no error in the above mentioned case.

## Suggested solution:
Maybe `start_transaction` should have the option to set isolation level (with REPEATABLE_READ as default). And commit()/rollback() need to set the isolation level back to the value it is set when connection is initialized.

hovaesco:
> Maybe start_transaction should have the option to set isolation level (with REPEATABLE_READ as default). And commit()/rollback() need to set the isolation level back to the value it is set when connection is initialized.

Thanks for rising the issue. 

Would the suggested solution solve the issue when a transaction is executed > 5 min? I wonder how setting separate isolation level for `start_transaction()` help here. 
IIUC, It's more about reusing the same transaction request instance in `start_transaction()` and `commit/rollback()`.
 
 @hashhar / @ebyhr / @findinpath any ideas?

hashhar:
[...]

PR #128, opened 2021-11-19
Add latest Trino version to test matrix

hashhar:
Currently we only test the client against first Trino version i.e. 351.

This means that any changes in future versions that cause the client to break aren't caught by CI.
See https://github.com/trinodb/trino-go-client/pull/28 for motivation.

PR #152, opened 2021-12-31
Fix transaction handling in auto-commit mode

hovaesco:
When an explicit transaction was started in auto-commit mode, the transaction request was not correctly attached to the cursor. That caused query and transaction were executed separately. Query was executed in separate transaction and request. 

Issue #129 

Design consideration

Starting a separate transaction in auto-commit may be disabled at all. It wasn't working, until now. If user wants to manage transaction separately can use connection in other isolation level than auto-commit.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -153,9 +153,8 @@ def run_trino():
         stop_trino(container_id, proc)
 
 
-@pytest.fixture(scope="module")
 def trino_version():
-    yield TRINO_VERSION
+    return TRINO_VERSION
 
 
 @click.group()
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -16,7 +16,8 @@ import pytest
 import pytz
 
 import trino
-from trino.exceptions import TrinoQueryError
+from tests.integration.conftest import trino_version
+from trino.exceptions import TrinoQueryError, TrinoUserError
 from trino.transaction import IsolationLevel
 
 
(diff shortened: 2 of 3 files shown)


tests/integration/conftest.py
```diff
@@ -28,7 +28,7 @@
 logger = trino.logging.get_logger(__name__)
 
 
-TRINO_VERSION = os.environ.get("TRINO_VERSION") or "351"
+TRINO_VERSION = os.environ.get("TRINO_VERSION") or "latest"
```
hovaesco, line 31: @hashhar that was copied from your PR: https://github.com/trinodb/trino-python-client/pull/128 

Looks like your PR is good to be merged. 
I need to test my change on newer Trino version, since there were some changes in auto-commit.

ebyhr: @hashhar Could you share your conclusion about testing version matrix?

hashhar: Thanks for reminder. I've merged the linked PR.

I however wanted to do something similar to how we test the older JDBC drivers against current server - testing a sliding window of versions (which can be a follow-up).

hashhar:
@hovaesco The version matrix PR was merged. Can you rebase and drop the first two commits?

hovaesco:
@hashhar done!

tests/integration/test_dbapi_integration.py
```diff
@@ -362,6 +376,22 @@ def test_transaction_multiple(trino_connection_with_transaction):
...
+        cur.execute(
+            """
+            CREATE TABLE memory.default.nation
+            AS SELECT * from tpch.tiny.nation
+            """)
+
+        with pytest.raises(TrinoUserError) as transaction_error:
+            cur.fetchall()
+        assert "Catalog only supports writes using autocommit: memory" \
+               in str(transaction_error.value)
```
hashhar, line 392: Would this need to be conditional based on Trino version being tested?

Should we add a fixture called `honors_autocommit` that uses the Trino version being tested to return a boolean that we can use to skip tests that don't apply to certain versions. Similar to JDBC in https://github.com/trinodb/trino/blob/83456f0ee6aeed817d4d0cab477f91f1e5821557/testing/trino-test-jdbc-compatibility-old-driver/src/test/java/io/trino/JdbcDriverCapabilities.java#L79

hovaesco: yes, I would need to check when autocommit mechanics was changed. 

I will add a conditional check and skip this test depending on the version.

hovaesco: done!

tests/integration/conftest.py
```diff
@@ -153,9 +153,8 @@ def run_trino():
         stop_trino(container_id, proc)
 
 
-@pytest.fixture(scope="module")
```
hashhar, line 156: Hah, I had recently learnt about pytest fixtures and ended up overusing them. 😛 

Thanks for the cleanup.

hashhar (approved):
Squash the commits (first won't pass without the 2nd).

LGTM.

tests/integration/conftest.py
hovaesco: Yeah, fixtures cannot be used in `@pytest.mark.skipif`, so that's why I converted it. 

Squash done.

PR #154, opened 2022-01-30
Add support for client tags in Python API

abrisan:
Signed-off-by: Alex Brisan <abrisan1@bloomberg.net>

### DIFF
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -107,7 +107,8 @@ class Connection(object):
         request_timeout=constants.DEFAULT_REQUEST_TIMEOUT,
         isolation_level=IsolationLevel.AUTOCOMMIT,
         verify=True,
-        http_session=None
+        http_session=None,
+        client_tags=None
     ):
         self.host = host
         self.port = port
@@ -128,6 +129,7 @@ class Connection(object):
         self.redirect_handler = redirect_handler
         self.max_attempts = max_attempts
         self.request_timeout = request_timeout
+        self.client_tags = client_tags
 
         self._isolation_level = isolation_level
         self._request = None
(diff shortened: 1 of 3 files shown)


trino/dbapi.py
```diff
@@ -401,7 +401,11 @@ def _deallocate_prepare_statement(self, added_prepare_header, statement_name):
     def _generate_unique_statement_name(self):
         return 'st_' + uuid.uuid4().hex.replace('-', '')
 
-    def execute(self, operation, params=None):
+    def execute(self, operation, params=None, *, tags=None):
```
ebyhr, line 404: Could you move `tags` to connection from query? For consistency with JDBC driver and CLI.

abrisan: @ebyhr thanks for the review. I can certainly add it to connection as well. Since the change to execute is non-destructive, would you mind making this a superset of the other API's and leaving it in execute as well? We have use-cases where this would make sense on a per-query basis.

ebyhr: Can you share the details of your use-case? Since it's "client"-tags, per-query basis might be confusing to users. (=client won't change per query)

abrisan: Ok, I will make the change. The use cases I had in mind can also be handled if the tags are passed in the connection.

hovaesco (approved):

trino/dbapi.py
```diff
@@ -402,6 +404,10 @@ def _generate_unique_statement_name(self):
         return 'st_' + uuid.uuid4().hex.replace('-', '')
 
     def execute(self, operation, params=None):
+        additional_http_headers = {}
+        if self._connection.client_tags:
+            additional_http_headers[constants.HEADER_CLIENT_TAGS] = ",".join(self._connection.client_tags)
```
ebyhr, line 409: Please set the header in `client.py` as other headers.

mdesmet:
We would also like to use this feature. @abrisan Are you still on this?

PR #157, opened 2022-02-01
Trim CI matrix and prepare for explicit backward compatibility tests

hashhar:
Existing matrix became 14 jobs long.

We don't need to test older Trino versions with every single Python version - just the latest is enough. Matrix is now 7 jobs long.


hovaesco (approved):
LGTM

ebyhr (approved):
Could you remove 3.6 from `setup.py` & `tox.ini` as well?

PR #158, opened 2022-02-02
Add JWT token support via connection string in sqlalchemy

ms32035:


trino/sqlalchemy/dialect.py
```diff
@@ -84,6 +84,12 @@ def create_connect_args(self, url: URL) -> Tuple[Sequence[Any], Mapping[str, Any
             kwargs["http_scheme"] = "https"
             kwargs["auth"] = BasicAuthentication(url.username, url.password)
 
+        if "accessToken" in url.query:
+            kwargs["auth"] = JWTAuthentication(url.query["accessToken"])
+
+        if "http_scheme" in url.query:
```
dungdm93, line 90: `http_scheme="https"` is required when used with any authentication.

trino/sqlalchemy/dialect.py
```diff
@@ -66,7 +66,7 @@ def create_connect_args(self, url: URL) -> Tuple[Sequence[Any], Mapping[str, Any
         if url.port:
             kwargs["port"] = url.port
 
-        db_parts = (url.database or "system").split("/")
+        db_parts = (url.database or url.query.get("database") or "system").split("/")
```
dungdm93, line 69: Pass `database` via query make no sense to me

dungdm93 (changes requested):

trino/sqlalchemy/dialect.py
ms32035: This behaviour was taken from the existing sqlalchemy-trino library https://github.com/dungdm93/sqlalchemy-trino/blob/master/sqlalchemy_trino/dialect.py#L60
Can remove if makes no sense

dungdm93: This is because `super(TrinoDialect, self).create_connect_args(url)` spit & merge everything into `kwargs` (and `args`).
`database` is expected in URL path, not URL query.

trino/sqlalchemy/dialect.py
dungdm93: Yes, but I set schema to HTTPS when any authentication, not just JWT.
```diff
         if url.password:
             if not url.username:
                 raise ValueError("Username is required when specify password in connection URL")
-            kwargs["http_scheme"] = "https"
             kwargs["auth"] = BasicAuthentication(url.username, url.password)
 
+        if "accessToken" in url.query:
+            kwargs["auth"] = JWTAuthentication(url.query["accessToken"])
+
+        if "http_scheme" in url.query:
+            kwargs["http_scheme"] = url.query["http_scheme"]
+        if kwargs.get("auth"):
+            kwargs["http_scheme"] = "https"
```

ms32035: @dungdm93 can you please advise how to proceed after my latest changes

dungdm93: LGTM. But seem you need to sign the CLA first.
@ebyhr could you please help him.

dungdm93 (approved):

dungdm93:
@ms32035 could you rebase from master and add an example of how to use it into `README.md`
https://github.com/trinodb/trino-python-client#jwt-authentication

ebyhr:
@cla-bot check

trino/sqlalchemy/dialect.py
```diff
@@ -87,6 +87,10 @@ def create_connect_args(self, url: URL) -> Tuple[Sequence[Any], Mapping[str, Any
             kwargs["http_scheme"] = "https"
             kwargs["auth"] = BasicAuthentication(url.username, url.password)
 
+        if "accessToken" in url.query:
+            kwargs["http_scheme"] = "https"
+            kwargs["auth"] = JWTAuthentication(url.query["accessToken"])
```
ebyhr, line 92: I basically disagree to guess the authentication type from query parameter. Could you tell me why the below code is insufficient in your use case?
```python
from trino.auth import JWTAuthentication

engine = create_engine(
    "trino://<username>@<host>:<port>/<catalog>",
    connect_args={
        "auth": JWTAuthentication("<jwt_token>"),
        "http_scheme": "https",
    }
)
```

ebyhr (commented):
Please squash commits into one and edit the commit message to 
`Add JWT token support via connection string in sqlalchemy`

trino/sqlalchemy/dialect.py
ms32035: because there are several apps using the client, where the connection details are provided as a sqlalchemy connection string in a configuration file, command line parameter, etc, and not as a python object

dungdm93: I think query key should be snake_case to be the same convention as `http_scheme`.
So, @ms32035 would you mind to change it to `access_token`

ms32035: @dungdm93 changed

ebyhr (approved):

ebyhr:
Merged, thanks!

PR #159, opened 2022-02-07
Return None when failed to get server version in SQLAlchemy

ebyhr:
This is the same approach as pymssql. This commit adds SQLAlchemy support for version < 352. The another way to get server version is getting from `system.runtime.nodes` table, but the environment may disallow accessing the catalogs.

### DIFF
--- a/trino/sqlalchemy/dialect.py
+++ b/trino/sqlalchemy/dialect.py
@@ -17,10 +17,13 @@ from sqlalchemy.engine.base import Connection
 from sqlalchemy.engine.default import DefaultDialect, DefaultExecutionContext
 from sqlalchemy.engine.url import URL
 
-from trino import dbapi as trino_dbapi
+from trino import dbapi as trino_dbapi, logging
 from trino.auth import BasicAuthentication
 from trino.dbapi import Cursor
+from trino.exceptions import TrinoUserError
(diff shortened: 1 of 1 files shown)


trino/sqlalchemy/dialect.py
```diff
@@ -266,11 +269,15 @@ def has_sequence(self, connection: Connection, sequence_name: str, schema: str =
...
         query = "SELECT version()"
-        res = connection.execute(sql.text(query))
-        version = res.scalar()
-        return tuple([version])
+        try:
+            res = connection.execute(sql.text(query))
+            version = res.scalar()
+            return tuple([version])
+        except TrinoUserError as e:
+            logger.debug(f"Failed to get server version: {e.message}")
```
hashhar, line 279: Should this be a user-facing warning instead?
I'm not sure what SQLAlchemy uses this version for so maybe it's fine to ignore.

hashhar (approved):

trino/sqlalchemy/dialect.py
ebyhr: This variable is used in each dialect and SQLAlchemy core code doesn't use it as far as I confirmed.
If the library developers didn't implement the method, SQLAlchemy sets None by default, so ignoring or debug log is fine to me.

Issue #32, opened 2020-06-03
trino-python-client returns dtype string for datetime/timestamp fields


RJ3:
presto-client returns data type of string for a timestamp field.
Setup:  
 * PrestoSQL 333  
 * PostgreSQL 12.1  
 * presto-client 0.300.0

Tested:  
1. Using the PrestoSQL JDBC driver, the field is returned as a Java timestamp.
2. Using `psycopg2` driver with SQLalchemy directly to Postgres, the field is returned as a python datetime object or pandas timestamp object.  
3. **Using presto-client 0.300, the field is returned as a string.**

Expectation:   
The field should be returned as a python datetime object or pandas timestamp object, akin to the PostgreSQL driver.

Note:   
The issue has also been reported for pyhive (no longer under current development). I have not tested PrestoDB's version to see if the other team has also worked on the topic.

findepi:
The biggest question is how to avoid breaking users. 
For now we should implement napping to datetime behing a toggle (init param).
At some point, we probably should consdier switching the default, as we do not want to maintain inferior representation for ever.
cc @electrum

alexbrisan:
@findepi I would like to contribute a fix for this. Would you like me to detail a fix first, or just submit a Pull Request?

findepi:
@alexbrisan awesome!

it's worthwhile to have a conversation before doing work, if the work would take a lot of time.
Perhaps this won't be the case here, will it?

Things to consider

-  what do other drivers do? eg. postgresql's or oracle's?
- do we have a toggle to switch between old & new behavior?
- i am not expert on date/time/timezones handling in Python, but it seems the standard library doesn't cover all the use-cases (hence `pytz` exists). 
   - this is why checking with oracle's driver can be interesting. PostgreSQL doesn't have SQL standard-conforming `timestamp with time zone` data type and both Trino and Oracle do.

alexbrisan:
I will take a look at the oracle one and come up with a plan!

john-bodley:
Thanks for authoring the PR.

I was running into an issue with Apache Superset related to inconsistencies in how the frontend handled timestamps. We leverage Pandas for reading a SQL query into a dataframe and serializing it to JSON.

For context here's difference between MySQL and Trino, 

```python
import pandas as pd 
from sqlalchemy import create_engine

pd.read_sql_query(
    sql="SELECT TIMESTAMP('2022-01-01 00:00:00') AS ts", 
[...]

PR #160, opened 2022-02-13
Support decimal, date, time, timestamp with time zone and timestamp python types in prepared statements and result sets

### LINKED ISSUE #150: Decimal data type returned as strings
## Issue:
I am using python to pull data from a Trino database, and columns formatted as DECIMAL in the database are returned as strings rather than numbers.

**For Instance:**

```
import trino
conn = trino.dbapi.connect(
host='host'
port=port
user='user'
catalog='catalog'
schema='schema'
)
cur = conn.cursor()
cur.execute("SELECT DECIMAL_COL FROM TABLE LIMIT 5")
rows = cur.fetchall()
print(rows)
```
**This is an example of the output I get:**
```
[['123456.78'],['234567.89'],['3456.78'],['45678.90'],['67.89']]
```
Is the output above what is expected? I would expect this to be returned as numbers rather than strings. I do not encounter this issue with integers or double data types.

mdesmet:
Fixes #150 , #32

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,9 +317,39 @@ The transaction is created when the first SQL statement is executed.
 exits the *with* context and the queries succeed, otherwise
 `trino.dbapi.Connection.rollback()` will be called.
 
-## Development
+# Improved Python types
 
-### Getting Started With Development
+If you enable the flag `experimental_python_types`, the client will convert the results of the query to the 
+corresponding Python types. For example, if the query returns a `DECIMAL` column, the result will be a `Decimal` object.
+
+Limitations of the Python types are described in the 
+[Python types documentation](https://docs.python.org/3/library/datatypes.html). These limitations will generate an 
+exception `trino.exceptions.DataError` if the query returns a value that cannot be converted to the corresponding Python 
+type.
+
+```python
+import trino
+import pytz
+from datetime import datetime
+
+conn = trino.dbapi.connect(
+    ...
+)
+
+cur = conn.cursor(experimental_python_types=True)
+
+params = datetime(2020, 1, 1, 16, 43, 22, 320000, tzinfo=pytz.timezone('America/Los_Angeles'))
+
+cur.execute("SELECT ?", params=(params,))
+rows = cur.fetchall()
+
+assert rows[0][0] == params
+assert cur.description[0][1] == "timestamp with time zone"
+```
+
+# Development
+
+## Getting Started With Development
 
 Start by forking the repository and then modify the code in your fork.
 
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ with open("trino/__init__.py", "rb") as f:
 kerberos_require = ["requests_kerberos"]
 sqlalchemy_require = ["sqlalchemy~=1.3"]
 
-all_require = kerberos_require + sqlalchemy_require
+all_require = ["pytz"] + kerberos_require + sqlalchemy_require
 
 tests_require = all_require + [
     # httpretty >= 1.1 duplicates requests in `httpretty.latest_requests`
@@ -36,7 +36,6 @@ tests_require = all_require + [
     "httpretty < 1.1",
     "pytest",
     "pytest-runner",
-    "pytz",
     "click",
 ]
 
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -10,7 +10,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import math
-from datetime import datetime
+from datetime import datetime, time, date, timezone, timedelta
+from decimal import Decimal
 
 import pytest
 import pytz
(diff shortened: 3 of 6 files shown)


tests/integration/test_dbapi_integration.py
```diff
@@ -158,6 +168,38 @@ def test_array_query_param(trino_connection):
     assert rows[0][0] == "array(integer)"
 
 
+def test_array_timestamp_query_param(trino_connection):
```
hovaesco, line 171: Could you please add tests also for date and time?

mdesmet:
As mentioned in #32, this change will impact the behavior on existing codebases. Should we provide a feature flag for this improved type detection?

tests/integration/test_dbapi_integration.py
hashhar: And in the tests let's ensure following scenarios are covered:

- negative year (-0001-01-01)
- zero year (0000-01-01)
- gregorian-julian switch (1582-10-04 to 1582-10-14)
- epoch (1970-01-01)
- timestamps that lie in a DST gap (i.e. a timestamp that never occurred in a timezone)
- timestamps that are doubled due to DST

cc: @findepi since timestamps

mdesmet: Here are the min max values for python, now I think still it makes sense to support native types as they will handle 99% of normal use cases. We need to decide on how to handle the impossible cases.

```
>>> from datetime import date, time, datetime
>>> date.min
datetime.date(1, 1, 1)
>>> date.max
datetime.date(9999, 12, 31)
>>> time.min
datetime.time(0, 0)
>>> time.max
datetime.time(23, 59, 59, 999999)
>>> datetime.min
datetime.datetime(1, 1, 1, 0, 0)
>>> datetime.max
datetime.datetime(9999, 12, 31, 23, 59, 59, 999999)
```

It seems also supported to use non-existing dates, both in Python as in Trino cfr. https://libguides.ctstatelibrary.org/hg/colonialresearch/calendar

See also See also https://github.com/trinodb/trino/issues/5781

Doubled timestamps are not yet tested.

mdesmet:
Time with time zone (utc offset) doesn't really exist as a Python type. It's possible to have time with a timezone, but however without an actual moment in time, it's impossible to know the UTC offset. So I don't think this is something we can support for now.

tests/integration/test_dbapi_integration.py
mdesmet: FYI: in snowflake invalid dates just generate exceptions eg snowflake.connector.errors.InterfaceError: 252005: Failed to convert: field DATE '0000-01-01': DATE::-719528, Error: date value out of range

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,114 @@ def test_string_query_param(trino_connection):
...
+    rows = cur.fetchall()
+
+    assert rows[0][0] == params
+
+
+def test_special_dates_query_param(trino_connection):
+    cur = trino_connection.cursor()
+
+    for params in (
+        # datetime(-1, 1, 1, 0, 0, 0).date(),
```
nineinchnick, line 199: If these are not supported, add a comment explaining why - is it because of some limitations in Python or other issues. If it's not relevant, remove these.

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,114 @@ def test_string_query_param(trino_connection):
...
+
+    params = time(12, 3, 44, 333000)
+
+    cur.execute("SELECT ?", params=(params,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == params
+
+
+@pytest.mark.skip(reason="time with time zone currently not supported")
```
nineinchnick, line 221: What's required for it to be supported?

nineinchnick (approved):

tests/integration/test_dbapi_integration.py
mdesmet: We don't want to wrap these type of errors into a custom exception? Now the code will give a standard python exception.

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,114 @@ def test_string_query_param(trino_connection):
     assert rows[0][0] == "six'"
 
 
+def test_decimal_query_param(trino_connection):
+    cur = trino_connection.cursor()
+
+    cur.execute("SELECT ?", params=(Decimal('0.142857'),))
```
hashhar, line 130: we should test with:

- null
- decimal(38, 0)
- largest supported Trino decimal(38, 38)
- smallest supported Trino decimal(38, 38)

Also are we losing the scale and precision information? We should test whether the scale and precision of the input is being preserved.

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,114 @@ def test_string_query_param(trino_connection):
...
 
-    assert rows[0][0] == "2020-01-01 00:00:00.000 UTC"
+    assert rows[0][0] == params
     assert cur.description[0][1] == "timestamp with time zone"
 
 
+def test_datetime_with_time_zone_numeric_offset(trino_connection):
+    cur = trino_connection.cursor()
+
+    cur.execute("SELECT TIMESTAMP '2001-08-22 03:04:05.321 -08:00'")
```
hashhar, line 164: Can we test with values fractional seconds with trailing zeros?

tests/integration/test_dbapi_integration.py
hashhar: add inline comment beside each date with it's significance. See https://github.com/trinodb/trino/blob/7e99818f0194e6d792e3d67fc235511af279cf94/plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/BaseSqlServerTypeMapping.java#L320 for inspiration.

tests/integration/test_dbapi_integration.py
hashhar: Instead of removing can you add a test that fails? It helps document and also helps ensure tests get updated if behaviour changes.

tests/integration/test_dbapi_integration.py
hashhar: We need to provide a flag to control behaviour since this is a regression in client behaviour. Before this change it was possible to use the client to query negative dates or values lying in DST gaps/doubling. Now it's not.

hashhar (commented):
This introduces regression regarding range of values that can be represented and hence needs a toggle somewhere - maybe add an alternative Cursor implementation?

Left some other comments regarding tests.

mdesmet:
To convert named timezones i have added pytz as dependency. As we support older python versions, it seems easier to just use pytz, which was already a test dependency. Maybe we should make this an optional dependency as it's only required for when the flag is activated. Please advice.

Starting from python 3.9 we could replace this with the standard zoneinfo (however seems still need an additional library on Windows: https://tzdata.readthedocs.io/en/latest/).

setup.py
```diff
@@ -28,15 +28,14 @@
 kerberos_require = ["requests_kerberos"]
 sqlalchemy_require = ["sqlalchemy~=1.3"]
 
-all_require = kerberos_require + sqlalchemy_require
+all_require = ["pytz"] + kerberos_require + sqlalchemy_require
```
hovaesco, line 31: Move above as `timezone_require = ["pytz"]` and add here `timezone_require` to be consistent.

trino/client.py
```diff
@@ -472,10 +475,11 @@ class TrinoResult(object):
     https://docs.python.org/3/library/stdtypes.html#generator-types
     """
 
-    def __init__(self, query, rows=None):
+    def __init__(self, query, rows=None, experimental_python_types: bool = False):
```
hovaesco, line 478: Please add an entry in README on how to use `experimental_python_types` flag.

hovaesco (commented):
>  Maybe we should make this an optional dependency as it's only required for when the flag is activated.
Let's leave it as a required dependency since e.g I want to continue to use old python types and then switch to the new one.

setup.py
mdesmet: kerberos and sqlalchemy are optional dependencies. I would say we need then another list called default_require, any other default dependency would then be added to that list.

trino/client.py
```diff
@@ -36,6 +36,9 @@
 import copy
 import os
 import re
+from decimal import Decimal
```
ppalucha, line 39: nit: ordering of import - non-standard modules should go at the end + alphabetical sorting

ppalucha (approved):

setup.py
hashhar: Let's keep as it is. The idea with the lists was to use them for optional deps.

If we find this list becoming large we can extract a separate list. No need to do it now.

README.md
```diff
@@ -317,9 +317,39 @@ The transaction is created when the first SQL statement is executed.
...
+```python
+import trino
+import pytz
+from datetime import datetime
+
+conn = trino.dbapi.connect(
+    ...
+)
+
+cur = conn.cursor()
```
hovaesco, line 339: Shouldn't `experimental_python_types=True` be added here?

hovaesco (approved):

trino/client.py
```diff
@@ -494,20 +498,63 @@ def __iter__(self):
...
+                return datetime.strptime(value, "%Y-%m-%d").date()
+            elif raw_type == "timestamp with time zone":
+                dt, tz = value.rsplit(' ', 1)
+                if tz.startswith('+') or tz.startswith('-'):
+                    return datetime.strptime(value, "%Y-%m-%d %H:%M:%S.%f %z")
+                return datetime.strptime(dt, "%Y-%m-%d %H:%M:%S.%f").replace(tzinfo=pytz.timezone(tz))
+            elif "timestamp" in raw_type:
+                return datetime.strptime(value, "%Y-%m-%d %H:%M:%S.%f")
+            elif "time with time zone" in raw_type:
+                return datetime.strptime(value, "%H:%M:%S.%f").time()
```
john-bodley, line 537: Shouldn't this include timezone information? 

Note the `dateutil.parser.parse` method could be useful and replace the logic in lines 527–539, somewhat ignoring the specific `raw_type`, i.e., `value` should contain the timezone information if the `raw_type` is `timestamp with time zone` or `time with time zone`, i.e., 


```python
from dateutil.parser import parse

if raw_type in ("time",  "time with time zone", "timestamp", "timestamp with time zone"):
    result = parse(value)
    
    if raw_type in ("time",  "time with time zone"):
        result = result.time().replace(tzinfo=result.tzinfo)
        
    return result
```

john-bodley, line 537: Shouldn't this include timezone information? 

Note the `dateutil.parser.parse` method could be useful and replace the logic in lines 527–539, somewhat ignoring the specific `raw_type`, i.e., `value` should contain the timezone information if the `raw_type` is `timestamp with time zone` or `time with time zone`, i.e., 


```python
from dateutil.parser import parse

if raw_type in ("time",  "time with time zone", "timestamp", "timestamp with time zone"):
    result = parse(value)
    
    if raw_type in ("time",  "time with time zone"):
        result = result.time().replace(tzinfo=result.tzinfo)
        
    return result
```

john-bodley:
Thanks for authoring the PR.

I was running into an issue with Apache Superset related to inconsistencies in how the frontend handled timestamps. We leverage Pandas for reading a SQL query into a dataframe and serializing it to JSON.

For context here's difference between MySQL and Trino, 

```python
import pandas as pd 

pd.read_sql_query(
    sql="SELECT TIMESTAMP('2022-01-01 00:00:00') AS ts", 
    con=create_engine("mysql://root@localhost:3360"),
).to_json()

pd.read_sql_query(
    sql="SELECT TIMESTAMP '2022-01-01 00:00:00' AS ts", 
    con=create_engine("trino://localhost: 8080"),
).to_json()
```

which outputs `{"ts":{"0":1640995200000}}` (which infers UTC per the EPOCH definition) and `{"ts":{"0":"2022-01-01 00:00:00.000"}}` (which is in the local timezone) respectively and thus are treated differently in JavaScript:

```javascript
new Date(1640995200000)
> Sat Jan 01 2022 13:00:00 GMT+1300 (New Zealand Daylight Time)

new Date("2022-01-01 00:00:00.000")
> Sat Jan 01 2022 00:00:00 GMT+1300 (New Zealand Daylight Time)
```

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,267 @@ def test_string_query_param(trino_connection):
...
+        TIME '00:00:00.000'
+    """)
+    rows = cur.fetchall()
+
+    assert rows[0][0] == '0.142857'
+    assert rows[0][1] == '2018-01-01'
+    assert rows[0][2] == '2019-01-01 00:00:00.000 +01:00'
+    assert rows[0][3] == '2019-01-01 00:00:00.000 UTC'
+    assert rows[0][4] == '2019-01-01 00:00:00.000'
+    assert rows[0][5] == '00:00:00.000'
```
hashhar, line 146: Do these also implicitly assert that the data type is same? Or is it too smart and calls str/repr on the values and compares? For safety it would be nice to have explicit assertion on types too.

tests/integration/test_dbapi_integration.py
hashhar: This is now added as `test_null_decimal`, `test_biggest_decimal`, `test_smallest_decimal` and `test_highest_precision_decimal`.

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,267 @@ def test_string_query_param(trino_connection):
...
+    params = datetime(2020, 1, 1, 16, 43, 22, 320000)
+
+    cur.execute("SELECT ?", params=(params,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == params
+    assert cur.description[0][1] == "timestamp"
+
+
+def test_datetime_with_trailing_zeros(trino_connection):
```
hashhar, line 209: nit: Use timestamp in method names since that's the datatype being tested (makes searching for relevant tests easier).

hashhar, line 209: This test differs from others around it. Maybe move it before `test_datetime_query_param` to keep all datetime param tests together. I was confused for a while. 😅

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,267 @@ def test_string_query_param(trino_connection):
...
+    cur.execute("SELECT TIMESTAMP '2001-08-22 03:04:05.321 -08:00'")
+    rows = cur.fetchall()
+
+    assert rows[0][0] == datetime.strptime("2001-08-22 03:04:05.321 -08:00", "%Y-%m-%d %H:%M:%S.%f %z")
+
+
+def test_unexisting_datetimes_with_time_zone_query_param(trino_connection):
+    cur = trino_connection.cursor(experimental_python_types=True)
+
+    params = datetime(2021, 3, 28, 2, 30, 0, tzinfo=pytz.timezone('Europe/Brussels'))
```
hashhar, line 277: add a comment about significance of this value - might not be obvious to future readers.

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,267 @@ def test_string_query_param(trino_connection):
...
+def test_datetime_with_time_zone_numeric_offset(trino_connection):
+    cur = trino_connection.cursor(experimental_python_types=True)
+
+    cur.execute("SELECT TIMESTAMP '2001-08-22 03:04:05.321 -08:00'")
+    rows = cur.fetchall()
+
+    assert rows[0][0] == datetime.strptime("2001-08-22 03:04:05.321 -08:00", "%Y-%m-%d %H:%M:%S.%f %z")
+
+
+def test_unexisting_datetimes_with_time_zone_query_param(trino_connection):
```
hashhar, line 274:
```suggestion
def test_datetimes_with_time_zone_in_dst_gap_query_param(trino_connection):
```

OR 

```suggestion
def test_nonexistent_datetimes_with_time_zone_query_param(trino_connection):
```

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,267 @@ def test_string_query_param(trino_connection):
...
+
+    params = pytz.timezone('US/Eastern').localize(datetime(2002, 10, 27, 1, 30, 0), is_dst=True)
+
+    cur.execute("SELECT ?", params=(params,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == datetime(2002, 10, 27, 1, 30, 0, tzinfo=pytz.timezone('US/Eastern'))
+
+
+def test_doubled_datetimes_second_query_param(trino_connection):
```
hashhar, line 294: Maybe merge this case into previous test method itself?

tests/integration/test_dbapi_integration.py
hashhar: Oh, or maybe it should go before `test_null_datetime_with_time_zone` since that's where the read queries are tested without query params.

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,267 @@ def test_string_query_param(trino_connection):
...
+    cur.execute("SELECT CAST(NULL AS DATE)")
+    rows = cur.fetchall()
+
+    assert rows[0][0] is None
+
+
+def test_unsupported_python_dates(trino_connection):
+    cur = trino_connection.cursor(experimental_python_types=True)
+
+    # dates not supported in Python date type
```
hashhar, line 328: this gives impression that everything else is supported. Is that true?

Might also be useful to add test with Trino's min and max dates + Python's min and max dates (if they differ).

tests/integration/test_dbapi_integration.py
```diff
@@ -123,22 +124,267 @@ def test_string_query_param(trino_connection):
...
+
+    params = time(12, 3, 44, 333000)
+
+    cur.execute("SELECT ?", params=(params,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == params
+
+
+def test_time_with_time_zone_query_param(trino_connection):
```
hashhar, line 379: Does time also accept timedelta?

Let's also test `null` for times as well.

tests/integration/test_dbapi_integration.py
```diff
@@ -158,6 +404,38 @@ def test_array_query_param(trino_connection):
...
+
+    assert rows[0][0] == params
+
+    cur.execute("SELECT TYPEOF(?)", params=(params,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == "array(timestamp(6))"
+
+
+def test_array_timestamp_with_timezone_query_param(trino_connection):
```
hashhar, line 423: add test with arrays containing None (null) values too.

trino/dbapi.py
```diff
@@ -456,6 +478,7 @@ def fetchone(self) -> Optional[List[Any]]:
         """
 
         try:
+            assert self._iterator is not None
```
hashhar, line 481: Did we hit this somehow? What causes it?

trino/dbapi.py
```diff
@@ -359,12 +361,27 @@ def _format_prepared_param(self, param):
...
+            # named timezones
+            if hasattr(param.tzinfo, 'zone'):
+                return "TIMESTAMP '%s %s'" % (datetime_str, param.tzinfo.zone)
+            # offset-based timezones
+            return "TIMESTAMP '%s %s'" % (datetime_str, param.tzinfo.tzname(param))
+
+        # We can't calculate the offset for a time without a point in time
+        if isinstance(param, datetime.time) and param.tzinfo is None:
+            time_str = param.strftime("%H:%M:%S.%f")
+            return "TIME '%s'" % time_str
```
hashhar, line 379: What about time with timezone?

hashhar (commented):
A few remaining comments. Thanks for the work on this @mdesmet.

Looks almost good to go.

tests/integration/test_dbapi_integration.py
hashhar: @mdesmet Was this addressed? Can you point me to the method which verifies this? I think this is also the source of my confusion regarding https://github.com/trinodb/trino-python-client/pull/160#discussion_r821385194.

Maybe add a comment there too indicating that time with time zone is not supported.

trino/dbapi.py
mdesmet: All methods from tzinfo require a valid datetime to return a numeric offset. As far as I see we can't do anything more. I have added tests for both cases, but both of them are raising a NotSupportedError for now.

tests/integration/test_dbapi_integration.py
hashhar: Sorry I wasn't precise enough. What I meant is that does time also accept offsets for time zone (in addition to no timezone + named offsets).

mdesmet: I have added test_time_with_time_zone_negative_offset and test_time_with_time_zone_positive_offset now. Named timezones are not supported in Trino for time with time zone type.

tests/integration/test_dbapi_integration.py
```diff
@@ -158,6 +471,54 @@ def test_array_query_param(trino_connection):
     assert rows[0][0] == "array(integer)"
 
 
+def test_array_none_query_param(trino_connection):
+    cur = trino_connection.cursor(experimental_python_types=True)
+
+    params = [None, None]
```
hashhar, line 477: It'd be useful to have a mix of None and non-None value in the array as well (which I what I originally meant when asking to add this test - but thanks for covering NULL arrays as well - I indeed missed that). What I was trying to test was that whether we correctly infer the type of the array.

mdesmet: Python doesn't have typed arrays. the type is determined for every value. I guess if we start mixing types, Trino will error. Combining None and another type should just work. Will add a test for that.

hashhar: None + another type should work in Trino provided we are CASTing the NULLs according to type of other elements in array.

hashhar: If such NULL + non-NULL arrays don't work correctly when translating to the query we should block them within the client maybe.

trino/client.py
```diff
@@ -494,20 +498,69 @@ def __iter__(self):
...
+            elif "timestamp" in raw_type:
+                return datetime.strptime(value, "%Y-%m-%d %H:%M:%S.%f")
+            elif "time with time zone" in raw_type:
+                matches = re.match(r'^(.*)([\+\-])(\d{2}):(\d{2})$', value)
+                assert matches is not None
+                if matches.group(2) == '+':
+                    tz = timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
+                else:
+                    tz = -timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
+                return datetime.strptime(matches.group(1), "%H:%M:%S.%f").time().replace(tzinfo=timezone(tz))
```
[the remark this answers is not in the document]
hovaesco: LGTM, we could avoid checking `if matches.group(2) == '+'` and just add +/- and don't have a conditional here. But I don't have a strong opinion about either way.

tests/integration/test_dbapi_integration.py
hashhar: bump @mdesmet

tests/integration/test_dbapi_integration.py
hashhar: Resolved by adding `test_array_none_and_another_type_query_param`

It does work correctly.

trino/client.py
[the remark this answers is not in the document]
hashhar: Good point. It seems for values like `2022-01-01 01:01:01 05:30` we'd convert this to negative offset which seems wrong.

@mdesmet can you passthrough the sign here? Or explicitly compare to `-` as well. the default should be positive offset AFAIK.

hashhar (approved):
LGTM % a few more comments @mdesmet

tests/integration/test_dbapi_integration.py
mdesmet: It is done just above

```
for value in rows[0]:
        assert isinstance(value, str)
```

trino/client.py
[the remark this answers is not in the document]
mdesmet: Good point about the default. Will make that change now.

hashhar (approved):
Thank you.

PR #162, opened 2022-02-27
Support redirection through webbrowser for OAuth authentication

### LINKED ISSUE #161: Improve Oauth authentication flow to launch webbrowser and cache token
The Trino cli client and JDBC driver allow the Oauth2 token to be cached, this prevents prompting the user to go through authentication flow for each separate query. 

**Current behavior**

Currently the python trino client prints out urls in the console for the user. 

image

**Improved behavior**

The first query per thread shall launch a webbrowser to go through the authentication flow with the external authenticator. On success it will cache the token in the user keychain (secure local storage) if available (or in a per connection cache), similar to the trino CLI and JDBC driver.

Ideally the token can be stored to be reused over multiple threads (eg. dbt launches multiple threads with each their own designated database connection).

mdesmet:
Fixes #161

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.h
 
 A callback to handle the redirect url can be provided via param `redirect_auth_url_handler` of the `trino.auth.OAuth2Authentication` class. By default, it will try to launch a web browser (`trino.auth.WebBrowserRedirectHandler`) to go through the authentication flow and output the redirect url to stdout (`trino.auth.ConsoleRedirectHandler`). Multiple redirect handlers are combined using the `trino.auth.CompositeRedirectHandler` class.
 
-The OAuth2 token will be cached either per `trino.auth.OAuth2Authentication` instance.
+The OAuth2 token will be cached either per `trino.auth.OAuth2Authentication` instance or, when keyring is installed, it will be cached within a secure backend (MacOS keychain, Windows credential locker, etc) under a key including host of the Trino connection. Keyring can be installed using `pip install 'trino[external-authentication-token-cache]'`.
 
 - DBAPI
 
--- a/README.md
+++ b/README.md
@@ -167,11 +167,14 @@ the [`JWT` authentication type](https://trino.io/docs/current/security/jwt.html)
 
 ### OAuth2 Authentication
 
-- `OAuth2Authentication` class can be used to connect to a Trino cluster configured with
+The `OAuth2Authentication` class can be used to connect to a Trino cluster configured with
 the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.html).
-- A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it just outputs the redirect url to stdout.
 
-* DBAPI
+A callback to handle the redirect url can be provided via param `redirect_auth_url_handler` of the `trino.auth.OAuth2Authentication` class. By default, it will try to launch a web browser (`trino.auth.WebBrowserRedirectHandler`) to go through the authentication flow and output the redirect url to stdout (`trino.auth.ConsoleRedirectHandler`). Multiple redirect handlers are combined using the `trino.auth.CompositeRedirectHandler` class.
+
+The OAuth2 token will be cached either per `trino.auth.OAuth2Authentication` instance.
+
+- DBAPI
 
     ```python
     from trino.dbapi import connect
@@ -185,7 +188,7 @@ the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.h
     )
     ```
 
-* SQLAlchemy
+- SQLAlchemy
 
     ```python
     from sqlalchemy import create_engine
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,9 @@ with open("trino/__init__.py", "rb") as f:
 
 kerberos_require = ["requests_kerberos"]
 sqlalchemy_require = ["sqlalchemy~=1.3"]
+external_authentication_token_cache_require = ["keyring"]
 
+# We don't add localstorage_require to all_require as users must explicitly opt in to use keyring.
 all_require = kerberos_require + sqlalchemy_require
 
 tests_require = all_require + [
@@ -80,6 +82,7 @@ setup(
         "kerberos": kerberos_require,
         "sqlalchemy": sqlalchemy_require,
         "tests": tests_require,
+        "external-authentication-token-cache": external_authentication_token_cache_require,
     },
     entry_points={
         "sqlalchemy.dialects": [
--- a/tests/unit/test_dbapi.py
+++ b/tests/unit/test_dbapi.py
@@ -9,9 +9,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-from requests import Session
+import threading
+import uuid
 from unittest.mock import patch
+
+import httpretty
+from httpretty import httprettified
+from requests import Session
+
+from tests.unit.oauth_test_utils import _post_statement_requests, _get_token_requests, RedirectHandler, \
+    GetTokenCallback, REDIRECT_RESOURCE, TOKEN_RESOURCE, PostStatementCallback, SERVER_ADDRESS
+from trino import constants
+from trino.auth import OAuth2Authentication
 from trino.dbapi import connect
 
 
@@ -39,3 +48,177 @@ def test_http_session_is_defaulted_when_not_specified(mock_client):
     # THEN
     request_args, _ = mock_client.TrinoRequest.call_args
     assert mock_client.TrinoRequest.http.Session.return_value in request_args
+
+
+@httprettified
+def test_token_retrieved_once_per_auth_instance(sample_post_response_data):
+    token = str(uuid.uuid4())
+    challenge_id = str(uuid.uuid4())
+
+    redirect_server = f"{REDIRECT_RESOURCE}/{challenge_id}"
+    token_server = f"{TOKEN_RESOURCE}/{challenge_id}"
+
+    post_statement_callback = PostStatementCallback(redirect_server, token_server, [token], sample_post_response_data)
+
+    # bind post statement
+    httpretty.register_uri(
+        method=httpretty.POST,
+        uri=f"{SERVER_ADDRESS}:8080{constants.URL_STATEMENT_PATH}",
+        body=post_statement_callback)
+
+    # bind get token
+    get_token_callback = GetTokenCallback(token_server, token)
+    httpretty.register_uri(
+        method=httpretty.GET,
+        uri=token_server,
+        body=get_token_callback)
+
+    redirect_handler = RedirectHandler()
+
+    with connect(
+            "coordinator",
+            user="test",
+            auth=OAuth2Authentication(redirect_auth_url_handler=redirect_handler),
+            http_scheme=constants.HTTPS
+    ) as conn:
+        conn.cursor().execute("SELECT 1")
+        conn.cursor().execute("SELECT 2")
+        conn.cursor().execute("SELECT 3")
+
+    # bind get token
+    get_token_callback = GetTokenCallback(token_server, token)
+    httpretty.register_uri(
+        method=httpretty.GET,
+        uri=token_server,
+        body=get_token_callback)
+
+    redirect_handler = RedirectHandler()
+
+    with connect(
+            "coordinator",
+            user="test",
+            auth=OAuth2Authentication(redirect_auth_url_handler=redirect_handler),
+            http_scheme=constants.HTTPS
+    ) as conn2:
+        conn2.cursor().execute("SELECT 1")
+        conn2.cursor().execute("SELECT 2")
+        conn2.cursor().execute("SELECT 3")
+
+    assert len(_get_token_requests(challenge_id)) == 2
+
+
+@httprettified
+def test_token_retrieved_once_when_authentication_instance_is_shared(sample_post_response_data):
+    token = str(uuid.uuid4())
+    challenge_id = str(uuid.uuid4())
+
+    redirect_server = f"{REDIRECT_RESOURCE}/{challenge_id}"
+    token_server = f"{TOKEN_RESOURCE}/{challenge_id}"
+
+    post_statement_callback = PostStatementCallback(redirect_server, token_server, [token], sample_post_response_data)
+
+    # bind post statement
+    httpretty.register_uri(
+        method=httpretty.POST,
+        uri=f"{SERVER_ADDRESS}:8080{constants.URL_STATEMENT_PATH}",
+        body=post_statement_callback)
+
+    # bind get token
+    get_token_callback = GetTokenCallback(token_server, token)
+    httpretty.register_uri(
+        method=httpretty.GET,
+        uri=token_server,
+        body=get_token_callback)
+
+    redirect_handler = RedirectHandler()
+
+    authentication = OAuth2Authentication(redirect_auth_url_handler=redirect_handler)
+
+    with connect(
+            "coordinator",
+            user="test",
+            auth=authentication,
+            http_scheme=constants.HTTPS
+    ) as conn:
+        conn.cursor().execute("SELECT 1")
+        conn.cursor().execute("SELECT 2")
+        conn.cursor().execute("SELECT 3")
+
+    # bind get token
+    get_token_callback = GetTokenCallback(token_server, token)
+    httpretty.register_uri(
+        method=httpretty.GET,
+        uri=token_server,
+        body=get_token_callback)
+
+    with connect(
+            "coordinator",
+            user="test",
+            auth=authentication,
+            http_scheme=constants.HTTPS
+    ) as conn2:
+        conn2.cursor().execute("SELECT 1")
+        conn2.cursor().execute("SELECT 2")
+        conn2.cursor().execute("SELECT 3")
+
+    assert len(_post_statement_requests()) == 7
+    assert len(_get_token_requests(challenge_id)) == 1
+
+
+@httprettified
+def test_token_retrieved_once_when_multithreaded(sample_post_response_data):
+    token = str(uuid.uuid4())
+    challenge_id = str(uuid.uuid4())
+
+    redirect_server = f"{REDIRECT_RESOURCE}/{challenge_id}"
+    token_server = f"{TOKEN_RESOURCE}/{challenge_id}"
+
+    post_statement_callback = PostStatementCallback(redirect_server, token_server, [token], sample_post_response_data)
+
+    # bind post statement
+    httpretty.register_uri(
+        method=httpretty.POST,
+        uri=f"{SERVER_ADDRESS}:8080{constants.URL_STATEMENT_PATH}",
+        body=post_statement_callback)
+
+    # bind get token
+    get_token_callback = GetTokenCallback(token_server, token)
+    httpretty.register_uri(
+        method=httpretty.GET,
+        uri=token_server,
+        body=get_token_callback)
+
+    redirect_handler = RedirectHandler()
+
+    authentication = OAuth2Authentication(redirect_auth_url_handler=redirect_handler)
+
+    conn = connect(
+        "coordinator",
+        user="test",
+        auth=authentication,
+        http_scheme=constants.HTTPS
+    )
+
+    class RunningThread(threading.Thread):
+        lock = threading.Lock()
+
+        def __init__(self):
+            super().__init__()
+
+        def run(self) -> None:
+            with RunningThread.lock:
+                conn.cursor().execute("SELECT 1")
+
+    threads = [
+        RunningThread(),
+        RunningThread(),
+        RunningThread()
+    ]
+
+    # run and join all threads
+    for thread in threads:
+        thread.start()
+    for thread in threads:
+        thread.join()
+
+    assert len(_get_token_requests(challenge_id)) == 1
--- a/trino/auth.py
+++ b/trino/auth.py
@@ -22,6 +22,7 @@ from urllib.parse import urlparse
 from requests import Request
 from requests.auth import AuthBase, extract_cookies_to_jar
 from requests.utils import parse_dict_header
+import importlib
 
 import trino.logging
 from trino.client import exceptions
@@ -230,6 +231,40 @@ class _OAuth2TokenInMemoryCache(_OAuth2TokenCache):
         self._cache[host] = token
 
 
+class _OAuth2KeyRingTokenCache(_OAuth2TokenCache):
+    """
+    Keyring Token Cache implementation
+    """
+
+    def __init__(self):
+        super().__init__()
+        try:
+            self._keyring = importlib.import_module("keyring")
+        except ImportError:
+            self._keyring = None
+            logger.info("keyring module not found. OAuth2 token will not be stored in keyring.")
+
+    def is_keyring_available(self) -> bool:
+        return self._keyring is not None
+
+    def get_token_from_cache(self, host: str) -> Optional[str]:
+        try:
+            return self._keyring.get_password(host, "token")
+        except self._keyring.errors.NoKeyringError as e:
+            raise trino.exceptions.NotSupportedError("Although keyring module is installed no backend has been "
+                                                     "detected, check https://pypi.org/project/keyring/ for more "
+                                                     "information.") from e
+
+    def store_token_to_cache(self, host: str, token: str) -> None:
+        try:
+            # keyring is installed, so we can store the token for reuse within multiple threads
+            self._keyring.set_password(host, "token", token)
+        except self._keyring.errors.NoKeyringError as e:
+            raise trino.exceptions.NotSupportedError("Although keyring module is installed no backend has been "
+                                                     "detected, check https://pypi.org/project/keyring/ for more "
+                                                     "information.") from e
+
+
 class _OAuth2TokenBearer(AuthBase):
     """
     Custom implementation of Trino Oauth2 based authorization to get the token
@@ -239,7 +274,8 @@ class _OAuth2TokenBearer(AuthBase):
 
     def __init__(self, redirect_auth_url_handler: Callable[[str], None]):
         self._redirect_auth_url = redirect_auth_url_handler
-        self._token_cache = _OAuth2TokenInMemoryCache()
+        keyring_cache = _OAuth2KeyRingTokenCache()
+        self._token_cache = keyring_cache if keyring_cache.is_keyring_available() else _OAuth2TokenInMemoryCache()
         self._token_lock = threading.Lock()
         self._inside_oauth_attempt_lock = threading.Lock()
         self._inside_oauth_attempt_blocker = threading.Event()
--- a/trino/auth.py
+++ b/trino/auth.py
@@ -15,7 +15,9 @@ import json
 import os
 import re
 import threading
-from typing import Optional
+import webbrowser
+from typing import Optional, List, Callable
+from urllib.parse import urlparse
 
 from requests import Request
 from requests.auth import AuthBase, extract_cookies_to_jar
@@ -130,6 +132,7 @@ class _BearerAuth(AuthBase):
     """
     Custom implementation of Authentication class for bearer token
     """
+
     def __init__(self, token):
         self.token = token
 
@@ -156,9 +159,75 @@ class JWTAuthentication(Authentication):
         return self.token == other.token
 
 
-def handle_redirect_auth_url(auth_url):
-    print("Open the following URL in browser for the external authentication:")
-    print(auth_url)
+class RedirectHandler(metaclass=abc.ABCMeta):
+    """
+    Abstract class for OAuth redirect handlers, inherit from this class to implement your own redirect handler.
+    """
+
+    @abc.abstractmethod
+    def __call__(self, url: str) -> None:
+        raise NotImplementedError()
+
+
+class ConsoleRedirectHandler(RedirectHandler):
+    """
+    Handler for OAuth redirections to log to console.
+    """
+
+    def __call__(self, url: str) -> None:
+        print("Open the following URL in browser for the external authentication:")
+        print(url)
+
+
+class WebBrowserRedirectHandler(RedirectHandler):
+    """
+    Handler for OAuth redirections to open in web browser.
+    """
+
+    def __call__(self, url: str) -> None:
+        webbrowser.open_new(url)
+
+
+class CompositeRedirectHandler(RedirectHandler):
+    """
+    Composite handler for OAuth redirect handlers.
+    """
+
+    def __init__(self, handlers: List[Callable[[str], None]]):
+        self.handlers = handlers
+
+    def __call__(self, url: str):
+        for handler in self.handlers:
+            handler(url)
+
+
+class _OAuth2TokenCache(metaclass=abc.ABCMeta):
+    """
+    Abstract class for OAuth token cache, inherit from this class to implement your own token cache.
+    """
+
+    @abc.abstractmethod
+    def get_token_from_cache(self, host: str) -> Optional[str]:
+        pass
+
+    @abc.abstractmethod
+    def store_token_to_cache(self, host: str, token: str) -> None:
+        pass
+
+
+class _OAuth2TokenInMemoryCache(_OAuth2TokenCache):
+    """
+    In-memory token cache implementation. The token is stored per host, so multiple clients can share the same cache.
+    """
+
+    def __init__(self):
+        self._cache = {}
+
+    def get_token_from_cache(self, host: str) -> Optional[str]:
+        return self._cache.get(host)
+
+    def store_token_to_cache(self, host: str, token: str) -> None:
+        self._cache[host] = token
 
 
 class _OAuth2TokenBearer(AuthBase):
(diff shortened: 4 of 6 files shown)


mdesmet:
Adding the keyring feature instead of the threadlocal made the test fail. IMHO the threadlocal seems not so useful, this works different from the Trino CLI and JDBC driver. However maybe from security perspective it's a good default to not cache the token over multiple threads.

In trino CLI:
See https://trino.io/docs/current/installation/cli.html#external-authentication-sso

Further queries in the CLI session do not require additional logins while the authentication token remains valid. Token expiration depends on the external authentication type configuration.

Expired tokens force you to log in again.

in trino JDBC:
See https://trino.io/docs/current/installation/jdbc.html#parameter-reference
externalAuthenticationTokenCache | Allows the sharing of external authentication tokens between different connections for the same authenticated user until the cache is invalidated, such as when a client is restarted or when the classloader reloads the JDBC driver. This is disabled by default, with a value of NONE. To enable, set the value to MEMORY. If the JDBC driver is used in a shared mode by different users, the first registered token is stored and authenticates all users.

@lukasz-walkiewicz : what do you think?

mdesmet:
Open items: 

- Thread safety impact: writing and reading token is not thread safe. Might need locking to ensure only one token is obtained simultaneously.
- Hardcoded name of token, should be based on url and username.

trino/auth.py
```diff
@@ -251,13 +262,34 @@ def _get_token(self, token_server, response, **kwargs):
 
         raise exceptions.TrinoAuthError("Exceeded max attempts while getting the token")
 
+    def _get_token_from_cache(self):
```
hovaesco, line 265: This needs to be thread locked, the same for storing a token.

trino/auth.py
```diff
@@ -213,15 +222,17 @@ def _authenticate(self, response, **kwargs):
         response.content
         response.close()
 
-        self._thread_local.token = self._get_token(token_server, response, **kwargs)
-        return self._retry_request(response, **kwargs)
+        token = self._get_token(token_server, response, **kwargs)
+        self._store_token_to_cache(token)
```
hovaesco, line 226: By default, let's not cache a token like in JDBC driver to not introduce a potential breaking change.

mdesmet: Current behaviour is printing the OIDC url in the console? Could you elaborate on how the current OIDC is used in production?

I actually agree that caching the token in the keychain shouldn't be the default from a security perspective as it will persist even outside of the process context. However within the same process, given the same connection (host and username), I think it can be cached. Woud you agree?

Actually the unit test `test_multithreaded_oauth2_authentication_flow` makes you think that it only retrieves the token once per thread. However that's not actually happening because the _OAuth2TokenBearer is recreated for each http_session (every time we create a cursor a separate TrinoRequest is being created). Which makes the threadlocal storage only work per cursor. That's why i opted to make the _OAuth2TokenBearer an instance property instead of being recreated for every http_session, IMHO it should be linked to the lifetime of the connection instead of a cursor.

```

def set_http_session(self, http_session):
        http_session.auth = _OAuth2TokenBearer(http_session, self._redirect_auth_url)
        return http_session
```

trino/auth.py
mdesmet: Yeah, i see what you mean, but thread locking just reads and writes will not prevent us from retrieving the token multiple times in parallel threads when we get 40X errors. Ideally we will only launch the webbrowser flow only once, even in a multithread case. Still need to test this scenario more in depth.

mdesmet: Thread locking has been implemented, covering reads/writes to cache and launching the OIDC flow only once per process and per Connection object. I think this is how it should work. The default is now class instance based caching, if keyring is installed within venv, use keyring caching per username and host. If webbrowser is not installed, for example in a headless environment (CLI), the url is printed in the console as before.

I think i can add another test to proof that the caching of the token is per connection and not per cursor as before.

README.md
```diff
@@ -169,7 +169,7 @@ the [`JWT` authentication type](https://trino.io/docs/current/security/jwt.html)
 
 - `OAuth2Authentication` class can be used to connect to a Trino cluster configured with
 the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.html).
-- A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it just outputs the redirect url to stdout.
+- A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it will try to launch a web browser to go through the authentication flow, if that fails it just outputs the redirect url to stdout.
```
hovaesco, line 172: Add info about caching to README.

trino/auth.py
hovaesco: If I understand correctly, are you suggesting caching a token by default?

hovaesco (commented):
> I think i can add another test to proof that the caching of the token is per connection and not per cursor as before.

That would be great.

trino/auth.py
mdesmet: The default before was already caching the token **per cursor** and per thread, which to me seems illogical. The newly default introduced in this PR is a thread-safe per connection cache, giving the user the choice to either create a connection once and share it over multiple threads or create a connection per thread. If you look at the test (test_multithreaded_oauth2_authentication_flow) you can see that before this change the token is already cached: the token is only retrieved once per thread, while there are many more requests. However the test didn't take into account that the Bearer class is instantiated for each cursor, so from a user perspective it seemed like no caching is performed.

I think for most users the keyring will provide the most seamless experience. they can either do `pip install keyring` or `pip install 'trino-python-client[local-secure-storage]'` to even get less browser popups as they will solely depend on the lifetime of the token.

lukasz-walkiewicz:
First of all, thanks for the contribution @mdesmet.
I've tried running this locally in a new virtual environment however it didn't work:
```
(keyring)  ✘  ~/dev/prestosql/trino-python-client   feature/webbrowser  python main.py                                                                                      ~/dev/prestosql/trino-python-client
error could be tests
Exception in thread 1: :
Traceback (most recent call last):
  File "/nix/store/lrw7ifwnpz7rskfkphq9mj9fwm5i9ldc-python3-3.9.6/lib/python3.9/threading.py", line 973, in _bootstrap_inner
Exception in thread 2: :
Traceback (most recent call last):
  File "/nix/store/lrw7ifwnpz7rskfkphq9mj9fwm5i9ldc-python3-3.9.6/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/Users/lukasz/dev/prestosql/trino-python-client/main.py", line 30, in run
    self.run()
  File "/Users/lukasz/dev/prestosql/trino-python-client/main.py", line 30, in run
    cur.execute('SELECT * FROM system.runtime.nodes')
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/dbapi.py", line 441, in execute
    cur.execute('SELECT * FROM system.runtime.nodes')
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/dbapi.py", line 441, in execute
    result = self._query.execute()
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/client.py", line 561, in execute
    result = self._query.execute()
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/client.py", line 561, in execute
    response = self._request.post(self._sql, additional_http_headers)
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/client.py", line 361, in post
    response = self._request.post(self._sql, additional_http_headers)
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/client.py", line 361, in post
    http_response = self._post(
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/exceptions.py", line 136, in decorated
    http_response = self._post(
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/exceptions.py", line 136, in decorated
    raise error
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/exceptions.py", line 123, in decorated
    raise error
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/exceptions.py", line 123, in decorated
    result = func(*args, **kwargs)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/sessions.py", line 577, in post
    result = func(*args, **kwargs)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/sessions.py", line 577, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/sessions.py", line 515, in request
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/sessions.py", line 515, in request
    prep = self.prepare_request(req)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/sessions.py", line 443, in prepare_request
    prep = self.prepare_request(req)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/sessions.py", line 443, in prepare_request
    p.prepare(
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/models.py", line 322, in prepare
    p.prepare(
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/models.py", line 322, in prepare
    self.prepare_auth(auth, url)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/models.py", line 558, in prepare_auth
    self.prepare_auth(auth, url)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/requests/models.py", line 558, in prepare_auth
    r = auth(self)
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/auth.py", line 192, in __call__
    r = auth(self)
    token = self._get_token_from_cache(*user_host)
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/auth.py", line 287, in _get_token_from_cache
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/auth.py", line 192, in __call__
    return self._keyring.get_password(
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/keyring/core.py", line 55, in get_password
    token = self._get_token_from_cache(*user_host)
  File "/Users/lukasz/dev/prestosql/trino-python-client/trino/auth.py", line 287, in _get_token_from_cache
    return get_keyring().get_password(service_name, username)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/keyring/backends/fail.py", line 25, in get_password
    return self._keyring.get_password(
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/keyring/core.py", line 55, in get_password
    raise NoKeyringError(msg)
keyring.errors.NoKeyringError: No recommended backend was available. Install a recommended 3rd party backend package; or, install the keyrings.alt package if you want to use the non-recommended backends. See https://pypi.org/project/keyring for details.
    return get_keyring().get_password(service_name, username)
  File "/Users/lukasz/venvs/keyring/lib/python3.9/site-packages/keyring/backends/fail.py", line 25, in get_password
    raise NoKeyringError(msg)
keyring.errors.NoKeyringError: No recommended backend was available. Install a recommended 3rd party backend package; or, install the keyrings.alt package if you want to use the non-recommended backends. See https://pypi.org/project/keyring for details.
```
Environment:
```
(keyring)  ~/dev/prestosql/trino-python-client   feature/webbrowser  python                                                                                                  ~/dev/prestosql/trino-python-client
Python 3.9.6 (default, Oct 31 2021, 15:37:00)
[Clang 7.1.0 (tags/RELEASE_710/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.platform()
'macOS-10.16-x86_64-i386-64bit'
```
Which platform you tested it on?

mdesmet:
Tested on following platform:
```
>>> platform.platform()
'macOS-12.2.1-arm64-arm-64bit'
```

Checking on https://pypi.org/project/keyring/

Compatibility - macOS
macOS keychain support **macOS 11** (Big Sur) and later requires Python 3.8.7 or later with the “universal2” binary. See [#525](https://github.com/jaraco/keyring/issues/525) for details.

mdesmet:
> First of all, thanks for the contribution @mdesmet. I've tried running this locally in a new virtual environment however it didn't work:
> 

So it's possible that even if you have installed keyring, it doesn't work as there is no suitable backend. I adapted the code that it will fallback to just store the token as if keyring is not installed.

mdesmet:
> > I think i can add another test to proof that the caching of the token is per connection and not per cursor as before.
> 
> That would be great.

I added the test: test_token_only_retrieved_once_per_connection

README.md
mdesmet: I've added more details about the caching in the README.

mdesmet:
@lukasz-walkiewicz, @hovaesco  : Any other remarks or concerns? 

FYI launching the webbrowser and storing credentials in keyring is also what Snowflake is doing in their python connector, I think it will greatly improve the user experience of dbt in combination with oauth authentication. Am I missing any usage patterns that this behavior for OAuth authentication won't be compatible with?

hovaesco (commented):
Looks good. One question, if keyring is installed then a token is cached by default? Basically, having keyring installed is like an enabler to cache a token?

mdesmet:
> Looks good. One question, if keyring is installed then a token is cached by default? Basically, having keyring installed is like an enabler to cache a token?

Even without keyring installed the token is always cached per connection (actually per instance of OAuth2Authentication), compared to the old code where the token was cached per cursor. When you install keyring, the token will be reused even when you restart or rerun your application (eg multiple dbt runs won't prompt authentication again if your token is still valid).

README.md
```diff
@@ -167,11 +167,14 @@ the [`JWT` authentication type](https://trino.io/docs/current/security/jwt.html)
...
 
-- `OAuth2Authentication` class can be used to connect to a Trino cluster configured with
+The `OAuth2Authentication` class can be used to connect to a Trino cluster configured with
 the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.html).
-- A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it just outputs the redirect url to stdout.
 
-* DBAPI
+A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it will try to launch a web browser to go through the authentication flow, if that fails it just outputs the redirect url to stdout.
+
+The OAuth2 token will be cached either per `trino.auth.OAuth2Authentication` instance or, when keyring is installed, it will be cached within a secure backend (macos keychain, windows credential locker, ...) under a key including host and user of the Trino connection. Keyring can be installed using `pip install trino-python-client['local-secure-storage']`.
```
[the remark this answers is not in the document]
mdesmet: > `pip install 'trino[secure-local-storage]'` or `pip install -e '.[secure-local-storage]'`

I would only add the first one as the second one is only for local development (might confuse normal users).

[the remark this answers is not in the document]
hovaesco: Yeah, that's ok.

trino/auth.py
```diff
@@ -251,13 +281,49 @@ def _get_token(self, token_server, response, **kwargs):
...
+                except self._keyring.errors.NoKeyringError:
+                    print("Warning: although keyring module is installed no backend has been detected, "
+                          "check https://pypi.org/project/keyring/ for more information.")
+
+            # keyring is not installed, so we store the token within the context of the same connection
+            self._token = token
+
+    @staticmethod
+    def _determine_user_host(url, headers) -> Tuple[str, str]:
+        user = headers.get("X-Trino-User")
```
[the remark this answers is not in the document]
hovaesco: Is it used to create a key for saving a password by keyring?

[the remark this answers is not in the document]
mdesmet: Yes, that's it. Now given that the lifetime of a token isn't generally that long, I would think it's probably OK to get rid of the user. Also the keyring is not installed by default, so it means the user opts in on this behavior.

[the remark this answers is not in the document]
hovaesco: What if we use a different user between connections? The token will be overwritten by the newest one?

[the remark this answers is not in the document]
hovaesco: What if we use a different user between connections? The token will be overwritten by the newest one?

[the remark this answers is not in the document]
mdesmet: When you don't use keyring, the token is stored as instance variable within the Authentication class. So it's not reused unless you would reuse that same instance over multiple connections.

With keyring it will just try to reuse the existing token, only after it expires it will prompt again. Now the earlier code was dependent on the user set in the `trino.dbapi.Connection` object, this user is straight from user input. It is not required to set it up for OAuth connections.

The whole point about using keyring is mostly about providing a default user-friendly experience for user ran CLI applications like dbt, I think it would be quite uncommon if they would have multiple users for the same host and also using OAuth. The workaround would be then to not use keyring in this case. Maybe we need to describe this better in the docs?

I also don't think a lot of server applications will use this webbrowser and keyring feature. They can support redirection by adding their own `handle_redirect_auth_url`.

I also would never install keyring by default, it's the users them self that opt in by installing keyring.

trino/auth.py
```diff
@@ -168,14 +176,24 @@ class _OAuth2TokenBearer(AuthBase):
...
         self._redirect_auth_url = redirect_auth_url_handler
-        self._thread_local = threading.local()
-        http_session.hooks['response'].append(self._authenticate)
+        try:
+            self._keyring = importlib.import_module("keyring")
+        except ImportError:
+            self._keyring = None
+            print("Warning: keyring module not found. OAuth2 token will not be stored in keyring.")
+        self._token = None
+        self._token_lock = threading.Lock()
```
lukasz-walkiewicz, line 187: Why there are 2 locks needed?

trino/auth.py
```diff
@@ -168,14 +176,24 @@ class _OAuth2TokenBearer(AuthBase):
...
-    def __init__(self, http_session, redirect_auth_url_handler=handle_redirect_auth_url):
+    def __init__(self, redirect_auth_url_handler=handle_redirect_auth_url):
         self._redirect_auth_url = redirect_auth_url_handler
-        self._thread_local = threading.local()
-        http_session.hooks['response'].append(self._authenticate)
+        try:
+            self._keyring = importlib.import_module("keyring")
+        except ImportError:
+            self._keyring = None
+            print("Warning: keyring module not found. OAuth2 token will not be stored in keyring.")
```
lukasz-walkiewicz, line 185: Will it be printed for all clients which are not using `keyring`?

trino/auth.py
```diff
@@ -251,13 +281,48 @@ def _get_token(self, token_server, response, **kwargs):
...
+            if self._keyring is not None:
+                # keyring is installed, so we can reuse the same token over multiple threads
+                try:
+                    return self._keyring.get_password(
+                        f"{host}", "token"
+                    )
+                except self._keyring.errors.NoKeyringError:
+                    pass
+
+            # keyring is not installed, so we fall back to a per connection cache
```
lukasz-walkiewicz, line 295: per instance of `OAuth2Authentication` right?

trino/auth.py
```diff
@@ -157,8 +161,12 @@ def __eq__(self, other):
 
 
 def handle_redirect_auth_url(auth_url):
-    print("Open the following URL in browser for the external authentication:")
-    print(auth_url)
+    print("Opening your browser for the external authentication:")
+    result = webbrowser.open_new(auth_url)
+    if not result:
```
lukasz-walkiewicz, line 166: What do you think about having console and browser handlers seperately? This would allow users to use either
```python
connect(
  "coordinator",
  auth=OAuth2Authentication(redirect_auth_url_handler=WebBrowserHandler())
```
or
```python
connect(
  "coordinator",
  auth=OAuth2Authentication(redirect_auth_url_handler=ConsoleHandler())
```

mdesmet: I have no objection against having both handlers, but what's the use of the `ConsoleHandler`? If needed, a user can provide their own handler and override the default behaviour of launching the webbrowser. They would probably do something with the url instead of just printing it to the console?

The behaviour is actually pretty similar to how the JDBC driver works. See https://trino.io/docs/current/installation/jdbc.html?highlight=externalAuthentication

trino/auth.py
mdesmet: That's indeed correct, also mentioned in earlier reviews. I will update the comment.

trino/auth.py
mdesmet: So as the `_OAuth2TokenBearer` is instantiated as part of the __init__ of the `OAuth2Authentication` class, it will be shown only once. I thought it's a good way to remind the user that they can reduce the amount of browser oauth popups by installing keyring. It will indeed be shown if you don't install keyring and use the `OAuth2Authentication` class.

Maybe we can remove this and just write good docs. :)

tests/unit/test_dbapi.py
```diff
@@ -39,3 +47,62 @@ def test_http_session_is_defaulted_when_not_specified(mock_client):
     # THEN
     request_args, _ = mock_client.TrinoRequest.call_args
     assert mock_client.TrinoRequest.http.Session.return_value in request_args
+
+
+@httprettified
+def test_token_only_retrieved_once_per_connection(sample_post_response_data):
```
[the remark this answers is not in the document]
mdesmet: It's like you said, we can add a test to proof this is actually working.

trino/auth.py
lukasz-walkiewicz: I understand the motiviation but it should use a logger and maybe not on warning. It's a hint, not an indication of a problem.

trino/auth.py
lukasz-walkiewicz: The jdbc driver actually defines an interface: https://github.com/trinodb/trino/blob/master/client/trino-client/src/main/java/io/trino/client/auth/external/RedirectHandler.java with multiple implementations:
```java
DESKTOP_OPEN(new DesktopBrowserRedirectHandler()),
SYSTEM_OPEN(new SystemOpenRedirectHandler()),
PRINT(new SystemOutPrintRedirectHandler()),
OPEN(new CompositeRedirectHandler(ImmutableList.of(SYSTEM_OPEN, DESKTOP_OPEN))),
ALL(new CompositeRedirectHandler(ImmutableList.of(OPEN, PRINT)))
```
and even though the default is `ALL` users can also use `PRINT` which does not try to open up the browser.

trino/auth.py
lukasz-walkiewicz: I'm not sure I understand. So there are two critical sections:
* get/set token from the cache,
* obtain and replace the token.
For the latter `_inside_oauth_attempt_lock` does guarantee that only one thread tries to get a new token at a given time but I believe other threads will still try to get a new token after obtaining this lock even though the token was just refreshed and is still valid:
```python
with self._inside_oauth_attempt_lock:
    if not self._inside_oauth_attempt:
        self._inside_oauth_attempt = True
        self._attempt_oauth(response, **kwargs)
        self._inside_oauth_attempt = False
```

mdesmet: You're indeed right, it's not correct. I will make the change and add a test for this behaviour.

trino/auth.py
mdesmet: Ok, I will adapt the implementation to provide a `ConsoleHandler` and `WebBrowserHandler`.

trino/auth.py
mdesmet: OK, will use a logger. Not sure if anyone runs this with logging set to INFO, but I agree to your point that it's not really a warning from a logging perspective. Anyway it's explained in the docs already.

hovaesco:
Awesome job 🎉 One general question. Have you fully tested it with dbt-trino?

mdesmet:
> Awesome job 🎉 One general question. Have you fully tested it with dbt-trino?

Haven't tried that. Will try it during next week. On first sight we might have to ensure that only one instance of `trino.auth.OAuth2Authentication()` is created.

trino/auth.py
```diff
@@ -40,16 +41,16 @@ def get_exceptions(self):
...
-        config: Optional[str] = None,
-        service_name: str = None,
-        mutual_authentication: bool = False,
-        force_preemptive: bool = False,
-        hostname_override: Optional[str] = None,
-        sanitize_mutual_error_response: bool = True,
-        principal: Optional[str] = None,
-        delegate: bool = False,
-        ca_bundle: Optional[str] = None,
+            self,
```
lukasz-walkiewicz, line 44: undo or a move to a different commit

lukasz-walkiewicz (approved):
Thanks, looks much much better now :+1:
One last comment, could you please change the name of the PR to something more descriptive?

setup.py
```diff
@@ -27,6 +27,7 @@
 
 kerberos_require = ["requests_kerberos"]
 sqlalchemy_require = ["sqlalchemy~=1.3"]
+localstorage_require = ["keyring"]
```
hashhar, line 30: add this to `all_require`. `all` must install everything (except test requirements).

hashhar: I'm not too sure about making this optional though.

tests/unit/oauthtest.py
```diff
@@ -0,0 +1,128 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
hashhar, line 1: nit: rename as `test_oauth.py`?

trino/auth.py
```diff
@@ -156,9 +160,112 @@ def __eq__(self, other):
...
+
+        # keyring not available, so we fallback to the in-memory cache
+        return super().get_token_from_cache(host)
+
+    def store_token_to_cache(self, host: str, token: str) -> None:
+        try:
+            # keyring is installed, so we can store the token for reuse within multiple threads
+            self._keyring.set_password(host, "token", token)
+        except self._keyring.errors.NoKeyringError:
+            logger.info("Although keyring module is installed no backend has been detected, "
```
hashhar, line 264: Since we have this handling it should be safe to make "keyring" a default dependency instead of an "extra". WDYT?

hashhar (commented):
Quick skim, still need to look at the `_OAuth2TokenBearer` changes for correctness.

trino/auth.py
```diff
@@ -156,9 +160,112 @@ def __eq__(self, other):
         return self.token == other.token
 
 
-def handle_redirect_auth_url(auth_url):
-    print("Open the following URL in browser for the external authentication:")
-    print(auth_url)
+class RedirectHandler(metaclass=abc.ABCMeta):
+    """
+    Abstract class for Oauth redirect handlers.
```
hovaesco, line 165: nit: improve docstrings here and for classes below, like `OAuth`, remove extra spaces, keep dots etc.

trino/auth.py
mdesmet: In most cases you don't have to install any additional packages (eg recent MacOS or Windows). Only older MacOS or Linux require installing additional packages. Also from a security perspective it may be better to let the user opt in on this behaviour. 

Now we could indeed install it as a default dependency and make this a configuration parameter. Using a configuration parameter is definitely more explicit. We could then also don't ignore this exception, and forward the exception to the user. Now actually in those cases we just do simple INFO log and rely on the standard caching. That may not be so transparent.

something like:

```
conn = connect(
    user="<username>",
    auth=OAuth2Authentication(useKeyring=True),
    http_scheme="https",
    ...
)
```

hashhar: Since it depends on things working on the OS side I think it's better to have it an "extra" then since it's then a concious choice by the user.

We can revisit using parameter to control this in future as per feedback after using this.

tests/unit/oauthtest.py
hashhar: `conftest` is special actually. It comes from `pytest` framework. `oauth_test_utils` is better - thanks.

setup.py
hashhar: Discussed in https://github.com/trinodb/trino-python-client/pull/162#discussion_r841429442

trino/auth.py
mdesmet: I think in any case, maybe we should probably not ignore/log.info the `NoKeyRingError` and just rethrow it with a `Trino.exceptions.NotSupportedException`. I think it will be more explicit instead of silently using standard caching. Do you agree?

hashhar: Yes, if the keyring module is available but for some reason we are not using it then it should result in error shown to user instead of kind-of misleading them by using in-memory caching.

mdesmet:
> Awesome job 🎉 One general question. Have you fully tested it with dbt-trino?

I have succesfully tested the feature with dbt and also added the changes in the [dbt PR](https://github.com/starburstdata/dbt-trino/pull/41). We still will need to update the trino-python-client version there, after release.

If you want to try it yourself, you can run following commands

```
pip install git+https://github.com/mdesmet/dbt-trino.git@feature/oidc 
pip install git+https://github.com/mdesmet/trino-python-client.git@feature/webbrowser
```

README.md
```diff
@@ -167,11 +167,14 @@ the [`JWT` authentication type](https://trino.io/docs/current/security/jwt.html)
...
+The `OAuth2Authentication` class can be used to connect to a Trino cluster configured with
 the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.html).
-- A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it just outputs the redirect url to stdout.
 
-* DBAPI
+A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it will try to launch a web browser to go through the authentication flow and output the redirect url to stdout.
+
+The OAuth2 token will be cached either per `trino.auth.OAuth2Authentication` instance or, when keyring is installed, it will be cached within a secure backend (macos keychain, windows credential locker, etc) under a key including host of the Trino connection. Keyring can be installed using `pip install trino['secure-local-storage']`.
+
+- DBAPI
```
hashhar, line 177:
```suggestion
* DBAPI
```

to match the `SQLAlchemy` bullet below.

README.md
```diff
@@ -167,11 +167,14 @@ the [`JWT` authentication type](https://trino.io/docs/current/security/jwt.html)
 
 ### OAuth2 Authentication
 
-- `OAuth2Authentication` class can be used to connect to a Trino cluster configured with
+The `OAuth2Authentication` class can be used to connect to a Trino cluster configured with
 the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.html).
-- A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it just outputs the redirect url to stdout.
 
-* DBAPI
+A callback to handle the redirect url can be provided via param `redirect_auth_url_handler`, by default it will try to launch a web browser to go through the authentication flow and output the redirect url to stdout.
```
hashhar, line 173: Might be useful to provide the names of the inbuilt ones here.

README.md
```diff
@@ -167,11 +167,14 @@ the [`JWT` authentication type](https://trino.io/docs/current/security/jwt.html)
...
+
+The OAuth2 token will be cached either per `trino.auth.OAuth2Authentication` instance or, when keyring is installed, it will be cached within a secure backend (macos keychain, windows credential locker, etc) under a key including host of the Trino connection. Keyring can be installed using `pip install trino['secure-local-storage']`.
```
hashhar, line 175: nit: MacOS and Windows (instead of all lowercase)

hashhar: `pip install 'trino[secure-local-storage]'` Existing syntax with quotes inside braces fails for me.

setup.py
```diff
@@ -27,6 +27,7 @@
...
 
 all_require = kerberos_require + sqlalchemy_require
```
hashhar, line 32: add `localstorage_require` to `all_require`

hashhar (approved):
Impl looks good to me. I'd love to be able to integration test these instead of doing manually but that's separate effort.

Some comments on README.

mdesmet:
> Impl looks good to me. I'd love to be able to integration test these instead of doing manually but that's separate effort.
> 

There was still some issue with the locking that i fixed yesterday. The integration tests actually lock on each query, which means that every request including retry is queued instead of executed in parallel. The locking is required to make httpretty threadsafe. So the existing tests are not completely covering the real nature of a multiple requests blocking until one finishes the oauth authentication process.

https://github.com/trinodb/trino-python-client/blob/0430df37d8074442daa22ec6c35a742efd75e363/tests/unit/test_client.py#L589-L590

hashhar:
I don't think `cursor` is expected to be thread-safe. What does DB-API say about it?

Or do you mean we lock per-instance of connection?

README.md
hovaesco: Yes it fails on zsh terminals, see https://stackoverflow.com/a/30539963

mdesmet:
> I don't think `cursor` is expected to be thread-safe. What does DB-API say about it?
> 
> Or do you mean we lock per-instance of connection?

It's not that we want to share a cursor over multiple threads. It's about the requests that happen within the cursor execution. The cursor is the first place where we actually talk to the coordinator, as the connection is just a cursor factory. So within a mulithreaded execution (each thread with their own cursor), multiple requests receive a 401 error and should wait until the first lock goes through the OAuth authentication flow. When that one thread completes and stores the token, the other threads should wake up and retry the request.

Within the integration tests, as we put a lock on every cursor execution, this effectively lets only one thread go through the request, 401, retrieve token and retry request flow. while all other threads are waiting and then execute with the token provided by the first thread. This doesn't match the real behaviour where actually multiple threads fail on 401 (as mentioned above).

I tried without the lock but then the httpretty request counts not match up anymore. So the issue lies within the test framework as mentioned in the comment in the code.

https://github.com/trinodb/trino-python-client/blob/0430df37d8074442daa22ec6c35a742efd75e363/tests/unit/test_client.py#L587-L591

hovaesco (approved):
I tested it with dbt-trino and it works flawlessly, thank you for great work!

README.md
mdesmet: It's the other way around if we want to match the other sections.

mdesmet:
` keyring.errors.NoKeyringError: No recommended backend was available. Install a recommended 3rd party backend package; or, install the keyrings.alt package if you want to use the non-recommended backends. See https://pypi.org/project/keyring for details.`

We can't add add `localstorage_require` to `all_require` as within the tests there is no keyring backend. I will revert that change. I also don't think we should install keyring without explicit approval from the user.

hashhar:
Thanks, please add that as a comment to `localstorage_require` about why it's missing from `all_require`.

setup.py
```diff
@@ -80,6 +82,7 @@
         "kerberos": kerberos_require,
         "sqlalchemy": sqlalchemy_require,
         "tests": tests_require,
+        "secure-local-storage": localstorage_require,
```
hashhar, line 85: does anyone have better names? `oauth-cache`/`oauth-keyring`?

hashhar (approved):
LGTM % better name if anyone has any

setup.py
mdesmet: What do you think of `external-authentication-token-cache`? 

Would be similar to the [JDBC driver settings](https://trino.io/docs/current/installation/jdbc.html#parameter-reference).  Always helps to keep things as consistent as possible over multiple drivers.

hashhar: agreed, that seems a bit long but is very clear about what it does. @ebyhr your input would be helpful.

ebyhr: Long name looks no problem because the name won't appear in the scripts frequently. I have no strong opinion though.

hashhar: @mdesmet Let's go with `external-authentication-token-cache` like the JDBC driver. Looks good to merge otherwise.

hashhar (approved):
Thank you for your patience @mdesmet and contributing this useful feature.

PR #163, opened 2022-03-17
Get table comment from system catalog

ms32035:
This is actually thrown by trino 360+ when reading Iceberg tables


trino/sqlalchemy/dialect.py
```diff
@@ -237,6 +237,7 @@ def get_table_comment(self, connection: Connection, table_name: str, schema: str
         except error.TrinoQueryError as e:
             if e.error_name in (
                 error.NOT_FOUND,
+                error.NOT_SUPPORTED,
```
ebyhr, line 240: I think the right solution is fixing query at `SELECT "comment" FROM {properties_table}`. 
It should be `SELECT comment FROM system.metadata.table_comments` in my opinion.

ebyhr (commented):
Please capitalize commit message. 
https://github.com/trinodb/trino-python-client#code-style

trino/sqlalchemy/dialect.py
ms32035: @ebyhr that would require the user to also have access to the `system` catalog, so it might not be the best idea from security/permission management point of view

ebyhr: @ms32035 First of all, `$properties` table is specific to Hive and Iceberg connector. We shouldn't use such tables.

ms32035: So what is the suggested approach?

hashhar: We also have `information_schema.tables.table_comment` I believe?

ebyhr: Yes, but it's a hidden column for MySQL compatibility. Since JDBC driver uses `system.jdbc.tables` to get table metadata, I would recommend using `system.metadata.table_comments`.

trino/sqlalchemy/dialect.py
```diff
@@ -229,16 +229,20 @@ def get_check_constraints(
         return []
 
     def get_table_comment(self, connection: Connection, table_name: str, schema: str = None, **kw) -> Dict[str, Any]:
-        properties_table = self._get_full_table(f"{table_name}$properties", schema)
-        query = f'SELECT "comment" FROM {properties_table}'
+        query = dedent(
```
ebyhr, line 232: It would be better to raise a error when `schema` is None in my understanding.

ebyhr (approved):

ebyhr:
Thanks!

Issue #165, opened 2022-03-29
ModuleNotFoundError: No module named 'pytz'


aksakalli:
For the version [0.311.0](https://github.com/trinodb/trino-python-client/compare/0.310.0...0.311.0), `pytz` was removed from dependencies but included only for tests with the PR https://github.com/trinodb/trino-python-client/pull/160. I am getting `ModuleNotFoundError: No module named 'pytz'` although I did not set `experimental_python_types=True` or anything. I am just importing `trino.dbapi.connect`:

```
>>> from trino.dbapi import connect

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/trino/__init__.py", line 13, in <module>
    from . import auth
  File "/usr/local/lib/python3.9/site-packages/trino/auth.py", line 25, in <module>
    from trino.client import exceptions
  File "/usr/local/lib/python3.9/site-packages/trino/client.py", line 44, in <module>
    import pytz
ModuleNotFoundError: No module named 'pytz'
```

PR #166, opened 2022-03-29
Fix pytz module not found error

aksakalli:
Fixing https://github.com/trinodb/trino-python-client/issues/165 issue.

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -509,8 +509,7 @@ class TrinoResult(object):
     def response_headers(self):
         return self._query.response_headers
 
-    @classmethod
-    def _map_to_python_type(cls, item: Tuple[Any, Dict]) -> Any:
+    def _map_to_python_type(self, cls, item: Tuple[Any, Dict]) -> Any:
         (value, data_type) = item
 
         if value is None:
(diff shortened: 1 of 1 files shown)


hashhar (commented):
I'd have expected that the setup.py would have ensured that pytz gets installed as a dependency?

aksakalli:
> I'd have expected that the setup.py would have ensured that pytz gets installed as a dependency?

I thought the intention of https://github.com/trinodb/trino-python-client/pull/160 is to depend on `pytz` only when `experimental_python_types=True`.

hashhar (dismissed):
This seems like the better solution to me.

However you'd still want to add an entry to "extras_require" to setup.py for experimental types support so that pytz gets installed automatically.

The reason why pytz doesn't get installed by default is because `install_requires` in setup.py only lists `requests`.

hashhar:
On second thought, the API allows people to switch between normal and python types on the fly. So ideally pytz must be a required dependency otherwise the code may fail at runtime depending on how the cursor is created.

hashhar:
Alternative solution implemented in #168.

PR #167, opened 2022-03-29
Always install pytz dependency

aksakalli:
An alternative to https://github.com/trinodb/trino-python-client/pull/166 that fixes https://github.com/trinodb/trino-python-client/issues/165


hashhar (commented):
You'd instead want to create another list called `install_requires` and add `requests` and `pytz` to it.

Issue #169, opened 2022-03-31
support `executemany`


erikerlandson:
https://github.com/trinodb/trino-python-client/blob/master/trino/dbapi.py#L467

https://github.com/dungdm93/sqlalchemy-trino#pandas-support

MichaelTiemannOSC:
Why cannot we borrow the executemany of this code into our Trino code?  The execute, fetch, and fetchmany look just like the bog-standard Trino code.

https://airflow.apache.org/docs/apache-airflow-providers-google/stable/_modules/airflow/providers/google/cloud/transfers/trino_to_gcs.html

MichaelTiemannOSC:
Indeed that is correct.  But I posted my question before reading https://github.com/trinodb/trino-python-client/pull/170

That PR actually provides the implementation I'm seeking, so I'm eagerly waiting to see whether it make it into the next release.  My use-case is important 100M rows of SEC DERA data, and the row-by-row insertion path converting pandas dataframe into a textual VALUES clause is SLOW.  Having a path that can work at the dataframe level could speed things up considerably.

PR #170, opened 2022-04-01
Add support for executemany in DBAPI

ebyhr:
Relates to #169

### DIFF
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -244,6 +244,12 @@ class Cursor(object):
         return None
 
     @property
+    def update_type(self):
+        if self._query is not None:
+            return self._query.update_type
+        return None
+
+    @property
     def description(self):
         if self._query.columns is None:
             return None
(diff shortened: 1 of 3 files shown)


trino/dbapi.py
```diff
@@ -465,7 +465,28 @@ def execute(self, operation, params=None):
...
+
+        Return values are not defined.
+        """
+        for parameters in seq_of_params[:-1]:
+            self.execute(operation, parameters)
+            self.fetchall()
+        if seq_of_params:
+            self.execute(operation, seq_of_params[-1])
+        else:
+            self.execute(operation)
```
hovaesco, line 489: Why this is needed?

MichaelTiemannOSC: I note that the question of what to do when the parameter list is empty has been a question before: https://github.com/winterland1989/mysql-haskell/issues/30

However, SQLlite3 expressly permits executemany with an empy parameter list.  So my presumption would be this implementation is designed to support calling contexts that either strictly require, or do not strictly require, a non-empty parameter list.

ebyhr: @hovaesco This line is called when the method was called without parameters. I'm fine to throw an error for starting conservatively.

hovaesco: I don't have a strong opinion about doing this in either way. Some DB-API clients supports that and some do not. DB-API spec doesn't say anything about supporting that. 

I'm fine with supporting it here, just please add a test case for it.

trino/dbapi.py
```diff
@@ -465,7 +465,28 @@ def execute(self, operation, params=None):
...
+        sets constitutes undefined behavior, and the implementation is permitted (but not required)
+        to raise an exception when it detects that a result set has been created by an invocation of the operation.
+
+        The same comments as for .execute() also apply accordingly to this method.
+
+        Return values are not defined.
+        """
+        for parameters in seq_of_params[:-1]:
+            self.execute(operation, parameters)
+            self.fetchall()
```
hashhar, line 485: Can we identify and throw if a result set was returned? To prevent people from using this with SELECT queries and then wondering where the results go.

hashhar (commented):
Looks good to me.

I'd prefer to throw for queries returning multiple result sets if possible - but if it's not trivial to do then it should be fine as well.

trino/dbapi.py
ebyhr: The test case already exists in test_execute_many_without_params.

trino/dbapi.py
ebyhr: I am going to use updateType field, so multiple INSERT should be fine.

hashhar: We can adjust this condition over time. updateType looks good to me.

hashhar (approved):

ebyhr:
(Squashed commits into one)

PR #172, opened 2022-04-05
Disable supports_statement_cache in sqlalchemy dialect

DeanLa:
on sqlalchemy>=1.4 a warning is raised if supports_statement_cache
is not set.
To avoid the warning, supports_statement_cache is set to False.
More testing is required to set this to True.


ebyhr (approved):
Looks good except for the commit title and body. Please avoid past tense and wrap the body at 72 characters. 

https://github.com/trinodb/trino-python-client#code-style

trino/sqlalchemy/dialect.py
```diff
@@ -55,6 +55,9 @@ class TrinoDialect(DefaultDialect):
     supports_multivalues_insert = True
     postfetch_lastrowid = False
 
+    # Caching
```
hashhar, line 58: can you add a comment saying why it's disabled instead?

Something like `warnings are generated if not explicitly set and needs tests before it can be enabled`?

hashhar (approved):
Do also remove the reference to the issue number from the commit message. All the relevant information is already present in your commit message already.

ebyhr:
@DeanLa Please squash commits into one and edit the commit title as "Disable supports_statement_cache in sqlalchemy dialect"

ebyhr:
Let us merge once your CLA is registered in https://github.com/trinodb/cla/blob/master/contributors

ebyhr:
@DeanLa It's manual process and it takes for a while. Please be patient, sorry.

ebyhr:
@cla-bot check

ebyhr:
Thanks!

PR #173, opened 2022-04-22
Enable prepared statements over http.client._MAXLINE limit

### LINKED ISSUE #148: Too large headers when inserting data
We're having an issue with too big headers being passed between the client and the server when inserting bigger chunks of data. Namely, the `X-Trino-Added-Prepare` header (this has already been mentioned in https://github.com/trinodb/trino/issues/8797). The way it's passed down and back to the server makes it very unwieldy to process queries with a large number of parameters (as is the case when inserting many values). It requires configuring all the proxies and clients (including [monkey-patching the python HTTP client](https://github.com/python/cpython/blob/v3.10.1/Lib/http/client.py#L111)) along the way to values way beyond sensible.

It seems that the issue we're running into is that the inserts shouldn't have been prepared statements in the first place (especially since currently you cannot reuse them for uploading different chunks of data as they get [deallocated immediately](https://github.com/trinodb/trino-python-client/blob/0.309.0/trino/dbapi.py#L432) after the query has been executed).

As such, I think it would be good to have the option to skip using prepared statements which generate really bulky `X-Trino-Added-Prepare` headers (introduced in #41), in particular in the insert scenario.

mdesmet:
Fixes #148

Headers can become much longer then what is currently suported in Python's `http.client`. This code patches the header parsing from `http.client` to be able to parse the headers coming back from the Trino server.

Test is included. when removing the `disable_header_limit()` in client.py, the test will fail with `requests.exceptions.ConnectionError: ('Connection aborted.', LineTooLong('got more than 65536 bytes when reading header line'))`.

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -16,9 +16,11 @@ from decimal import Decimal
 import pytest
 import pytz
 import requests
+from requests.exceptions import ConnectionError
 
 import trino
 from tests.integration.conftest import trino_version
+from trino.client import PythonHttpClientOverride
 from trino.exceptions import TrinoQueryError, TrinoUserError, NotSupportedError
 from trino.transaction import IsolationLevel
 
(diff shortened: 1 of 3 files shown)


trino/client.py
```diff
@@ -62,6 +62,28 @@
...
+            if line in (b'\r\n', b'\n', b''):
+                break
+        hstring = b''.join(headers).decode('iso-8859-1')
+        return email.parser.Parser(_class=_class).parsestr(hstring)
+
+    http.client.parse_headers = parse_headers
+
+
+# Disable header limit to avoid LineTooLong('got more than 65536 bytes when reading header line')
+disable_header_limit()
```
hovaesco, line 84: Could this be moved into e.g `ClientSession` or `TrinoRequest` class?

mdesmet: It feels a lot less pythonic to me. it works of course and hides this override in `TrinoRequest` instead of the module's namespace.

```
class TrinoRequest: 
    @staticmethod
    def _disable_header_limit():
        import http.client
        import email.parser
        from http.client import HTTPMessage, _MAXLINE

        def parse_headers(fp, _class=HTTPMessage):
            headers = []
            while True:
                line = fp.readline(_MAXLINE + 1)
                headers.append(line)
                if line in (b'\r\n', b'\n', b''):
                    break
            header_string = b''.join(headers).decode('iso-8859-1')
            return email.parser.Parser(_class=_class).parsestr(header_string)

        http.client.parse_headers = parse_headers

    # Disable header limit to avoid LineTooLong('got more than 65536 bytes when reading header line')
    _disable_header_limit.__func__()
```

tests/integration/test_dbapi_integration.py
```diff
@@ -9,6 +9,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import http
```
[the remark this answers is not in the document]
mdesmet: That actually doesn't comply with the pre-commit check as it doesn't support private attributes.

```
tests/integration/test_dbapi_integration.py:15: error: Module "http.client" has no attribute "_MAXLINE"
Found 1 error in 1 file (checked 32 source files)
```

I think we don't want to import http.client at all in the module namespace as we only rely on some private property. I moved the import to the test itself.

trino/client.py
mdesmet: I created a helper class for this override. Let me know WDYT.

PR #175, opened 2022-04-24
Support `USE catalog.schema` statements

### LINKED ISSUE #153: USE catalog.schema not honored
The connection can be created with a catalog and schema argument. However I would have expected that the SQL USE statement would also switch between catalogs and schemas. Here's an example
```
import trino
conn = trino.dbapi.connect(
                host="localhost",
                port=8080,
                auth=None,
                user='the-user',
                http_scheme="http")
cur = conn.cursor()

# this works fine since table is fully qualified
cur.execute("select * from tpch.tiny.orders")
results = cur.fetchmany(10)
print(results)

# The USE statement to switch schema
cur.execute("use tpch.tiny")

# Query will fail because the USE statement is not honored
cur.execute("select * from orders")
results = cur.fetchmany(10)
print(results)
```

mdesmet:
Fixes #153

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -949,3 +949,54 @@ def retrieve_client_tags_from_query(run_trino, client_tags):
 
     query_client_tags = query_info['session']['clientTags']
     return query_client_tags
+
+
+@pytest.mark.skipif(trino_version() == '351', reason="current_catalog not supported in older Trino versions")
+def test_use_catalog_schema(trino_connection):
+    cur = trino_connection.cursor()
+    cur.execute('SELECT current_catalog, current_schema')
+    result = cur.fetchall()
+    assert result[0][0] is None
+    assert result[0][1] is None
+
+    cur.execute('USE tpch.tiny')
+    cur.fetchall()
+    cur.execute('SELECT current_catalog, current_schema')
+    result = cur.fetchall()
+    assert result[0][0] == 'tpch'
+    assert result[0][1] == 'tiny'
+
+    cur.execute('USE tpcds.sf1')
+    cur.fetchall()
+    cur.execute('SELECT current_catalog, current_schema')
+    result = cur.fetchall()
+    assert result[0][0] == 'tpcds'
+    assert result[0][1] == 'sf1'
+
+
+@pytest.mark.skipif(trino_version() == '351', reason="current_catalog not supported in older Trino versions")
+def test_use_catalog(run_trino):
+    _, host, port = run_trino
+
+    trino_connection = trino.dbapi.Connection(
+        host=host, port=port, user="test", source="test", catalog="tpch", max_attempts=1
+    )
+    cur = trino_connection.cursor()
+    cur.execute('SELECT current_catalog, current_schema')
+    result = cur.fetchall()
+    assert result[0][0] == 'tpch'
+    assert result[0][1] is None
+
+    cur.execute('USE tiny')
+    cur.fetchall()
+    cur.execute('SELECT current_catalog, current_schema')
+    result = cur.fetchall()
+    assert result[0][0] == 'tpch'
+    assert result[0][1] == 'tiny'
+
+    cur.execute('USE sf1')
+    cur.fetchall()
+    cur.execute('SELECT current_catalog, current_schema')
+    result = cur.fetchall()
+    assert result[0][0] == 'tpch'
+    assert result[0][1] == 'sf1'
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -20,7 +20,7 @@ from uuid import uuid4
 import click
 import trino.logging
 import pytest
-from trino.client import TrinoQuery, TrinoRequest
+from trino.client import TrinoQuery, TrinoRequest, ClientSession
 from trino.constants import DEFAULT_PORT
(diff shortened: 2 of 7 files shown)


tests/integration/test_dbapi_integration.py
```diff
@@ -859,3 +859,13 @@ def test_info_uri(trino_connection):
     cur.fetchall()
     assert cur.info_uri is not None
     assert cur._query.query_id in cur.info_uri
+
+
+def test_use_schema(trino_connection):
+    cur = trino_connection.cursor()
+    cur.execute('use tpch.tiny')
```
ebyhr, line 866: nit: Uppercase `use`

tests/integration/test_dbapi_integration.py
```diff
@@ -859,3 +859,13 @@ def test_info_uri(trino_connection):
     cur.fetchall()
     assert cur.info_uri is not None
     assert cur._query.query_id in cur.info_uri
+
+
+def test_use_schema(trino_connection):
+    cur = trino_connection.cursor()
```
ebyhr, line 865: Please add assertions to verify the original catalog and schema before executing `USE` query.

hashhar: Should we also assert that new cursors opened after a `USE` still use the default catalog and schema specified on the connection?

mdesmet:
Not sure why the build failed in 3.10 on the following error:

```
trino.exceptions.TrinoUserError: TrinoUserError(type=USER_ERROR, name=COLUMN_NOT_FOUND, message="line 1:8: Column 'current_catalog' cannot be resolved", query_id=20220424_155121_00213_82y8r)

```

ebyhr:
`current_catalog` was added in version 357, so it fails on 351.
https://trino.io/docs/current/release/release-357.html

tests/integration/test_dbapi_integration.py
mdesmet: Currently it seems indeed that new cursors don't take over the session properties.

However I don't think this will match the expectation of the users:

> JDBC defines a system session and transaction model for database access. A session is the duration of one connection to the database and is handled by a JDBC Connection object.

See [oracle docs](https://docs.oracle.com/javadb/10.8.3.0/devguide/cdevconcepts38375.html#:~:text=JDBC%20defines%20a%20system%20session,by%20a%20JDBC%20Connection%20object.&text=A%20Connection%20object%20represents%20a%20connection%20with%20a%20database.)

WDYT?

hashhar: DBAPI != JDBC. We should do what DB-API says and if it doesn't say explicitly then follow what some other drivers do (psycopg for example doesn't maintain search path changes across cursors last time I checked).

mdesmet: I checked the DBAPI specs, nothing mentioned about this. IMHO it's more of a general expectation. See e.g. https://stackoverflow.com/questions/8797724/mysql-concepts-session-vs-connection

Snowflake for example does apply it to the connection:

https://github.com/snowflakedb/snowflake-connector-python/blob/107c3033cb2c7c6961d06aa613e7c3a8f20ae435/src/snowflake/connector/cursor.py#L753-L759

In a lot of cases this is also not really a client thing, but something stored within the database side connection/session.

mdesmet: @ebyhr , @hashhar : WDYT?

hovaesco (approved):
LGTM

tests/integration/test_dbapi_integration.py
mdesmet: @hovaesco : What's your view on this? I think in the PR's code we currently do some locking in `ClientSession`. IMHO this should be per connection instead of per cursor.

hovaesco: It would be more useful when doing locking on connection/sesion level not per cursor which is limited and not user-friendly.

tests/integration/test_dbapi_integration.py
```diff
@@ -859,3 +859,25 @@ def test_info_uri(trino_connection):
...
+
+@pytest.mark.skipif(trino_version() == '351', reason="current_catalog not supported in older Trino versions")
+def test_use_schema(trino_connection):
+    cur = trino_connection.cursor()
+    cur.execute('SELECT current_catalog, current_schema')
+    result = cur.fetchall()
+    assert result[0][0] is None
+    assert result[0][1] is None
+
+    cur.execute('USE tpch.tiny')
```
ebyhr, line 872: Let's add another test case to specify only schema name. e.g. `USE sf1`.

ebyhr (approved):
Could you please rebase on upstream to resolve conflicts?

mdesmet:
I have rebased the code and made following changes:

- `ClientSession` is now thread-safe. it is initiated by the `dbapi.Connection` and some of its properties can be modified based on the headers coming back from the Trino server. It can later on be extended with more modifiable properties. 
- Added tests for both `USE catalog.schema` and `USE schema`
- Improved mypy typing for `ClientSession` class

hovaesco (commented):
Could you please (if possible) split the changes to two commits: refactor of `ClientSession` and add `USE` functionality?

hovaesco (approved):

ebyhr (approved):

ebyhr:
Merged, thanks!

PR #176, opened 2022-04-24
Support Infinity and nan in experimental_python_types

### LINKED ISSUE #54: Infinity and nan functions return strings instead of floats

mdesmet:
Fixes #54 

The issue also mentions support for map data type, which is implemented in a separate commit.

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -528,6 +528,21 @@ class TrinoResult(object):
                     "typeSignature": data_type["typeSignature"]["arguments"][0]["value"]
                 }
                 return [cls._map_to_python_type((array_item, raw_type)) for array_item in value]
+            if isinstance(value, dict):
+                if raw_type == "map":
+                    raw_key_type = {
+                        "typeSignature": data_type["typeSignature"]["arguments"][0]["value"]
+                    }
+                    raw_value_type = {
+                        "typeSignature": data_type["typeSignature"]["arguments"][1]["value"]
+                    }
+                    return {
+                        cls._map_to_python_type((key, raw_key_type)):
+                            cls._map_to_python_type((value[key], raw_value_type))
+                        for key in value
+                    }
+                # TODO: support row type
+                return value
             elif "decimal" in raw_type:
                 return Decimal(value)
             elif raw_type == "double":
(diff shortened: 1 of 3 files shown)


trino/client.py
```diff
@@ -524,8 +524,27 @@ def _map_to_python_type(cls, item: Tuple[Any, Dict]) -> Any:
...
+                }
             elif "decimal" in raw_type:
                 return Decimal(value)
+            elif "double" in raw_type:
+                if value == 'Infinity':
+                    return float("inf")
+                elif value == '-Infinity':
+                    return float("-inf")
+                elif value == 'NaN':
+                    return float("nan")
```
ebyhr, line 546: We could use `math.inf` & `math.nan` since the minimum supported version is 3.7 now. Please define constants if you use `float(x)` style.

mdesmet:
Within the existing code the conversion to MAP is assuming that types are equal over keys and values. Which is not necessarely the case. I think a dict can be represented by a MAP or a ROW. 

Currently the ROW is not implemented: neither in the prepared query parameter passing (currently giving a Trino error), neither in the resultset conversion (returned as is). The resultset conversion should probably be easy to be added as types come back from Trino. The prepared query parameter conversion might be a little trickier.

https://github.com/trinodb/trino-python-client/blob/939cab00ca5c95353999b63352b9658edfdecc99/trino/dbapi.py#L394-L400

trino/client.py
```diff
@@ -524,8 +528,31 @@ def _map_to_python_type(cls, item: Tuple[Any, Dict]) -> Any:
...
+                    return {
+                        cls._map_to_python_type((key, raw_key_type)):
+                            cls._map_to_python_type((value[key], raw_value_type))
+                        for key in value
+                    }
+                # TODO: support row type
+                return value
             elif "decimal" in raw_type:
                 return Decimal(value)
+            elif "double" in raw_type:
```
ebyhr, line 548: I think `==` is better than `in` considering custom types added by users.

ebyhr (approved):

trino/client.py
mdesmet: Good point, in this case it's not necessary as there is no precision anyway.

PR #177, opened 2022-04-25
Support Trino's `ROW` datatype

### LINKED ISSUE #34: Client should expose complex types like ROW in a usable format
Today the STRUCT/ROW fields are returned as plain strings by the Python client.

We should implement the equivalent of https://github.com/dropbox/PyHive/pull/328 here so that clients can properly parse and use the data returned for complex columns.

mdesmet:
Fixes #34

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -615,6 +615,32 @@ def test_boolean_query_param(trino_connection):
     assert rows[0][0] is False
 
 
+def test_row(trino_connection):
+    cur = trino_connection.cursor(experimental_python_types=True)
+    params = (1, Decimal("2.0"), datetime(2020, 1, 1, 0, 0, 0))
+    cur.execute("SELECT ?", (params,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == params
+
+
+def test_nested_row(trino_connection):
+    cur = trino_connection.cursor(experimental_python_types=True)
+    params = ((1, "test", Decimal("3.1")), Decimal("2.0"), datetime(2020, 1, 1, 0, 0, 0))
+    cur.execute("SELECT ?", (params,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == params
+
+
+def test_named_row(trino_connection):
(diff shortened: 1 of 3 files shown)


tests/integration/test_dbapi_integration.py
```diff
@@ -615,13 +615,22 @@ def test_boolean_query_param(trino_connection):
     assert rows[0][0] is False
 
 
-def test_float_query_param(trino_connection):
```
ebyhr, line 618: Why did you delete this test?

tests/integration/test_dbapi_integration.py
```diff
@@ -615,13 +615,22 @@ def test_boolean_query_param(trino_connection):
...
+    assert rows[0][0] == params
+
+
+def test_nested_row(trino_connection):
+    cur = trino_connection.cursor(experimental_python_types=True)
+    params = ((1, "test", Decimal("3.1")), Decimal("2.0"), datetime(2020, 1, 1, 0, 0, 0))
+    cur.execute("SELECT ?", (params,))
+    rows = cur.fetchall()
+
+    assert rows[0][0] == params
```
ebyhr, line 633: Let's add another test case having field names:
```sql
SELECT CAST(ROW(1, 2e0) AS ROW(x BIGINT, y DOUBLE))
```

mdesmet: I tested that yesterday, it has no impact on the data returned, if we want aliases of columns or fields we have to implement a DictCursor. Note that I also chose to use tuples instead of dicts for representing rows.

Not sure if we should revisit that decision to use tuples? From a Pyton API point of view It definitely feels more natural to pass down tuples as unnamed rows as query params compared to integer indexed dicts. WDYT?

tests/integration/test_dbapi_integration.py
mdesmet: I've put it back.

tests/integration/test_dbapi_integration.py
hashhar: I agree for anonymous rows tuples make more sense.

hashhar (approved):

PR #179, opened 2022-05-24
SqlAlchemy tweaks

hovaesco:
Resolves https://github.com/trinodb/trino-python-client/issues/106 and https://github.com/trinodb/trino-python-client/issues/108

PR #180, opened 2022-05-27
add trino sqlalchemy source

15374604475:
When we connected to trino through superset, the source could not be specified, so all queries could not be load-balanced through the presto gateway agent


ebyhr (commented):
Is this duplicate of https://github.com/trinodb/trino-python-client/pull/179/commits/dd011b6dc747d95ae0adb7da7bf7835ed096aac2?

PR #186, opened 2022-06-23
Add ability to specify catalog as a SQLALchemy table argument

laserkaplan:
This change allows for specifying a `trino_catalog` table argument to SQLALchemy Table objects, which is then checked when compiling statements and prepended to the proper objects. This allows for writing queries that talk to multiple catalogs at the same time.

This change was largely implemented by @AlexandreOuellet in the [PyHive](https://github.com/dropbox/PyHive) repository, with some minor edits by @VinceDPM and myself. However, since that repository is no longer well-maintained, and since Trino seems to have better support currently than Presto on the whole, I have switched my own focus to using Trino, and thus wanted to get this functionality working here as well.

As this is the first time I've attempted to contribute here, please let me know if I am missing anything in this PR! It would be great if this could be merged in an efficient manner.

### DIFF
--- a/tests/unit/sqlalchemy/test_compiler.py
+++ b/tests/unit/sqlalchemy/test_compiler.py
@@ -19,6 +19,7 @@ from sqlalchemy import (
     String,
     Table,
 )
+from sqlalchemy.schema import CreateTable
 
 from trino.sqlalchemy.dialect import TrinoDialect
 
@@ -26,8 +27,15 @@ metadata = MetaData()
 table = Table(
     'table',
     metadata,
-    Column('id', Integer, primary_key=True),
+    Column('id', Integer),
     Column('name', String),
+)
+table_with_catalog = Table(
+    'table',
+    metadata,
+    Column('id', Integer),
+    schema='default',
+    trino_catalog='other'
 )
 
 
@@ -64,3 +72,20 @@ def test_cte_insert_order(dialect):
         'FROM "table")\n'\
         ' SELECT cte.id, cte.name \n'\
         'FROM cte'
+
+
+def test_catalogs_argument(dialect):
+    statement = select(table_with_catalog)
+    query = statement.compile(dialect=dialect)
+    assert str(query) == 'SELECT default."table".id \nFROM "other".default."table"'
+
+
+def test_catalogs_create_table(dialect):
+    statement = CreateTable(table_with_catalog)
+    query = statement.compile(dialect=dialect)
+    assert str(query) == \
+        '\n'\
+        'CREATE TABLE "other".default."table" (\n'\
+        '\tid INTEGER\n'\
+        ')\n'\
+        '\n'
--- a/trino/sqlalchemy/compiler.py
+++ b/trino/sqlalchemy/compiler.py
@@ -10,6 +10,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from sqlalchemy.sql import compiler
+try:
+    from sqlalchemy.sql.expression import (
+        Alias,
+        CTE,
+        Subquery,
+    )
+except ImportError:
+    # For SQLAlchemy versions < 1.4, the CTE and Subquery classes did not explicitly exist
+    from sqlalchemy.sql.expression import Alias
+    CTE = type(None)
+    Subquery = type(None)
 
 # https://trino.io/docs/current/language/reserved.html
 RESERVED_WORDS = {
@@ -102,6 +113,31 @@ class TrinoSQLCompiler(compiler.SQLCompiler):
             text += "\nLIMIT " + self.process(select._limit_clause, **kw)
         return text
 
+    def visit_table(self, table, asfrom=False, iscrud=False, ashint=False,
+                    fromhints=None, use_schema=True, **kwargs):
+        sql = super(TrinoSQLCompiler, self).visit_table(
+            table, asfrom, iscrud, ashint, fromhints, use_schema, **kwargs
+        )
+        return self.add_catalog(sql, table)
+
+    @staticmethod
+    def add_catalog(sql, table):
+        if table is None:
+            return sql
+
+        if isinstance(table, (Alias, CTE, Subquery)):
+            return sql
+
+        if (
+                'trino' not in table.dialect_options
+                or 'catalog' not in table.dialect_options['trino']
+        ):
+            return sql
+
+        catalog = table.dialect_options['trino']['catalog']
+        sql = f'"{catalog}".{sql}'
+        return sql
+
 
 class TrinoDDLCompiler(compiler.DDLCompiler):
     pass
@@ -173,3 +209,7 @@ class TrinoTypeCompiler(compiler.GenericTypeCompiler):
 
 class TrinoIdentifierPreparer(compiler.IdentifierPreparer):
     reserved_words = RESERVED_WORDS
+
+    def format_table(self, table, use_schema=True, name=None):
+        result = super(TrinoIdentifierPreparer, self).format_table(table, use_schema, name)
+        return TrinoSQLCompiler.add_catalog(result, table)


tests/unit/sqlalchemy/test_compiler.py
```diff
@@ -44,3 +44,16 @@ def test_offset(dialect):
     statement = select(table).offset(0)
     query = statement.compile(dialect=dialect)
     assert str(query) == 'SELECT "table".id, "table".name \nFROM "table"\nOFFSET :param_1'
+
+
+def test_multiple_catalogs(dialect):
```
ebyhr, line 49: I'm not sure what "multiple catalogs" mean. Could you rename the test method or leave a comment?

trino/sqlalchemy/compiler.py
```diff
@@ -102,6 +112,38 @@ def limit_clause(self, select, **kw):
             text += "\nLIMIT " + self.process(select._limit_clause, **kw)
         return text
 
+    def visit_column(self, column, add_to_result_map=None, include_table=True, **kwargs):
```
ebyhr, line 115: Why do we need to use visitor pattern instead of just setting `X-Trino-Catalog` or `X-Trino-Schema` header?

ebyhr (commented):
Could you submit CLA if you haven't yet sent it?

tests/unit/sqlalchemy/test_compiler.py
laserkaplan: I have updated the test name to `test_catalog_argument`.

laserkaplan:
> Could you submit CLA if you haven't yet sent it?

I sent my signed CLA form yesterday. Please let me know if it hasn't been received!

trino/sqlalchemy/compiler.py
ebyhr: Thanks for your explanation. I understand `X-Trino-Catalog` doesn't fit such use cases.

ebyhr:
@cla-bot check

laserkaplan:
I have added an additional bit to this PR since it hadn't been approved/merged yet. The visitors worked when compiling SQL statements, but another method was needed to do the same functionality for DDL statements (like `CREATE TABLE`).

tests/unit/sqlalchemy/test_compiler.py
```diff
@@ -54,6 +62,24 @@ def test_offset(dialect):
...
+
+
+def test_catalogs_create_table(dialect):
+    statement = CreateTable(system_table)
+    query = statement.compile(dialect=dialect)
+    assert str(query) == \
+        '\n'\
+        'CREATE TABLE "system".information_schema."table" (\n'\
+        '\tid INTEGER NOT NULL, \n'\
+        '\tPRIMARY KEY (id)\n'\
```
ebyhr, line 78: This DDL looks invalid as Trino query.

trino/sqlalchemy/compiler.py
```diff
@@ -102,6 +112,38 @@ def limit_clause(self, select, **kw):
...
+        sql = super(TrinoSQLCompiler, self).visit_column(
+            column, add_to_result_map, include_table, **kwargs
+        )
+        table = column.table
+        return self.add_catalog(sql, table)
```
ebyhr, line 120: I understood the benefit for table, but not sure about column. Could you add a test case to show the usecase?

ebyhr (commented):
Please rebase on upstream.

tests/unit/sqlalchemy/test_compiler.py
laserkaplan: I changed the table to use a fake `other` catalog so that this test doesn't potentially interfere with the `system` catalog.

trino/sqlalchemy/compiler.py
laserkaplan: I think you're right that it's not needed. Removed the column visitor.

tests/unit/sqlalchemy/test_compiler.py
```diff
@@ -29,6 +30,13 @@
     Column('id', Integer, primary_key=True),
     Column('name', String),
 )
+table_with_catalog = Table(
+    'table',
+    metadata,
+    Column('id', Integer, primary_key=True),
```
hashhar, line 36: Is there some way to tell SQLAlchemy that our dialect doesn't support `primary_key`? The SQL it generates will lead to syntax errors when actually executed against Trino.

cc: @mdesmet @dungdm93 any ideas?

hashhar: Allowing to define but then silently ignoring them when creating feels misleading. Better to be explicit and fail fast in my opinion. I don't know if that's possible to implement though.

hashhar: Either way it'll probably end up being a separate PR so I won't block this PR for this reason but we should follow up on this.

mdesmet: Follow-up issue created: #205

laserkaplan: Thanks @hashhar @mdesmet ! I didn't realize originally that it was the `PRIMARY KEY` bit that was problematic.

laserkaplan:
Hi all, is there anything else blocking this PR from completion?

tests/unit/sqlalchemy/test_compiler.py
hashhar, line 36: Is the `primary_key=True` required for the test? I don't think so. If so please remove it since it distracts attention from what the test and this change is actually about.

hashhar: Same for the DDL which is used below - the `PRIMARY KEY` seems unrelated to the feature being added.

trino/sqlalchemy/compiler.py
```diff
@@ -10,6 +10,16 @@
...
+try:
+    from sqlalchemy.sql.expression import (
+        Alias,
+        CTE,
+        Subquery,
+    )
+except ImportError:
+    from sqlalchemy.sql.expression import Alias
+    CTE = type(None)
+    Subquery = type(None)
```
hashhar, line 22: Can you leave a comment here why this is ok to do?

tests/unit/sqlalchemy/test_compiler.py
hashhar: I see that the `primary_key=True` just above this one is pre-existing but that should also be removed. Unfortunate that it didn't get caught when it was introduced.

hashhar (approved):
Looks good to me. @mdesmet Can you please take a look as well?

@laserkaplan Please squash the commits now (since it's one logical change).

trino/sqlalchemy/compiler.py
```diff
@@ -102,6 +112,31 @@ def limit_clause(self, select, **kw):
...
+        if isinstance(table, (Alias, CTE, Subquery)):
+            return sql
+
+        if (
+                'trino' not in table.dialect_options
+                or 'catalog' not in table.dialect_options['trino']._non_defaults
+        ):
+            return sql
+
+        catalog = table.dialect_options['trino']._non_defaults['catalog']
```
mdesmet, line 136: We should avoid to directly call private members of sqlalchemy classes.

I think the following should also work (not tested). 

```suggestion
        catalog = table.dialect_options['trino']['catalog']
```

See https://github.com/sqlalchemy/sqlalchemy/blob/0585c66a20b8075252bb4bbdeef0448f224de96a/lib/sqlalchemy/sql/base.py#L438

trino/sqlalchemy/compiler.py
```diff
@@ -102,6 +112,31 @@ def limit_clause(self, select, **kw):
...
+    def add_catalog(sql, table):
+        if table is None:
+            return sql
+
+        if isinstance(table, (Alias, CTE, Subquery)):
+            return sql
+
+        if (
+                'trino' not in table.dialect_options
+                or 'catalog' not in table.dialect_options['trino']._non_defaults
```
mdesmet, line 132: See remark about private members

mdesmet (commented):
LGTM % comments

laserkaplan:
> Please squash the commits now (since it's one logical change).

Is "squash and merge" enabled on the repo for when this is ready to merge?

tests/unit/sqlalchemy/test_compiler.py
laserkaplan: I went ahead and removed this from both temp tables, as well as in this test.

trino/sqlalchemy/compiler.py
laserkaplan: Yep, this is because the `CTE` and `Subquery` classes weren't introduced until SQLAlchemy 1.4. Added a comment.

mdesmet:
> Is "squash and merge" enabled on the repo for when this is ready to merge?

AFAIK we don't do squash and merge, only rebase and merge.

mdesmet (approved):
LGTM % commit squash

tests/unit/sqlalchemy/test_compiler.py
```diff
@@ -64,3 +72,20 @@ def test_cte_insert_order(dialect):
         'FROM "table")\n'\
         ' SELECT cte.id, cte.name \n'\
         'FROM cte'
+
+
+def test_catalogs_argument(dialect):
+    statement = select(table_with_catalog)
+    query = statement.compile(dialect=dialect)
+    assert str(query) == 'SELECT default."table".id \nFROM "other".default."table"'
```
hashhar, line 80: Seems pre-existing but any idea why the schema name doesn't get quoted?

hashhar (approved):
I just reworded the commit message. LGTM.

tests/unit/sqlalchemy/test_compiler.py
laserkaplan: Looked into this a little. `visit_table` quotes table names by default, but it checks whether schema names "need" to be quoted (e.g. if they are a reserved word). The PyHive package got around this by considering everything to be a reserved word, which would trigger this quoting. I don't think it's necessary here, but it definitely is a little weird to see schema names not quoted while table names are.

hashhar: Thanks for looking into this and explaining. I think we can leave as is for now since from the few queries I tested with special schema names it was able to handle them correctly. I didn't try too hard to break it though to be honest. 😄

PR #195, opened 2022-07-07
Cancel current query on `cursor.close()`

### LINKED ISSUE #84: Closing a Cursor does not cancel the query
Description

When lazy-loading results using an `Cursor` object, a call of `Cursor.close()` does not cancel the query. This results in a (temporary) memory leak on the server because the server never gets informed that the cursor is closed/invalid/destroyed.

Expectation

Closing a cursor cancels queries on the server, releasing all a associated resources.

Steps to reproduce:

1. `execute()` a query using `trino.dbapi.connect(…).cursor()` cursor object (assuming a LOT of rows in result set)
2. fetching one row using `fetchone()` or `next(iter(result))` etc.
3. calling `cursor.close()` to clean up ressources
4. checking running queries in the Trino UI -> query is server-side stalled, waiting for the client to request more data

Considerations

- Could it be necessary to also deallocate prepared statements?
- a cancelled query/cursor is not `finished`. 
- Calling `Cursor.close()` must not `close()` its associated connection because the expectation (based on all other RDBMS clients) is that you may have long living (application lifetime) connections with short living (request lifetime) cursors. That means: connections to cursors are 1:n.
- When a `TrinoQuery` is canceled, `execute()` throws an exception. When its `finished`, `execute()` just (re-)send the query. This seems a bit inconsistent.

Sorry for not providing you with a full PR, but the issue may touch general design of the software. 

```
diff --git a/trino/client.py b/trino/client.py
index cc13967..5b53895 100644
--- a/trino/client.py
+++ b/trino/client.py
@@ -556,7 +556,7 @@ class TrinoQuery(object):
 
     def is_finished(self):
         # type: () -> bool
-        return self._finished
+        return self._finished or self._cancelled
 
     @property
     def response_headers(self):
diff --git a/trino/dbapi.py b/trino/dbapi.py
index 3ce5a63..abfe374 100644
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -467,7 +467,11 @@ class Cursor(object):
         self._query.cancel()
 
     def close(self):
-        self._connection.close()
+        # cancel query to maybe release server-side ressources
+        if self._query is not None:
+            self.cancel()
+        
+        #self._connection.close()
```

mdesmet:
Fixes #84

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -766,7 +766,7 @@ def test_select_tpch_1000(trino_connection):
 def test_cancel_query(trino_connection):
     cur = trino_connection.cursor()
     cur.execute("SELECT * FROM tpch.sf1.customer")
-    cur.fetchone()  # TODO (https://github.com/trinodb/trino/issues/2683) test with and without .fetchone
+    cur.fetchone()
     cur.cancel()  # would raise an exception if cancel fails
 
     cur = trino_connection.cursor()
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -775,6 +775,18 @@ def test_cancel_query(trino_connection):
     assert "Cancel query failed; no running query" in str(cancel_error.value)
 
 
+def test_close_cursor(trino_connection):
+    cur = trino_connection.cursor()
+    cur.execute("SELECT * FROM tpch.sf1.customer")
+    cur.fetchone()
+    cur.close()  # would raise an exception if cancel fails
+
+    cur = trino_connection.cursor()
+    with pytest.raises(Exception) as cancel_error:
+        cur.close()
+    assert "Cancel query failed; no running query" in str(cancel_error.value)
+
+
 def test_session_properties(run_trino):
     _, host, port = run_trino
 
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -567,7 +567,9 @@ class Cursor(object):
         self._query.cancel()
 
     def close(self):
-        self._connection.close()
+        self.cancel()
+        # TODO: Cancel not only the last query executed on this cursor
+        #  but also any other outstanding queries executed through this cursor.
 
 
 Date = datetime.date


tests/integration/test_dbapi_integration.py
```diff
@@ -775,6 +775,18 @@ def test_cancel_query(trino_connection):
     assert "Cancel query failed; no running query" in str(cancel_error.value)
 
 
+def test_close_cursor(trino_connection):
+    cur = trino_connection.cursor()
+    cur.execute("SELECT * FROM tpch.sf1.customer")
+    cur.fetchone()  # TODO (https://github.com/trinodb/trino/issues/2683) test with and without .fetchone
```
hovaesco, line 781: Is this TODO still relevant ?

mdesmet: That particular issue is closed. I have tested the same code when omitting the `cur.fetchone()` the query is still cancelled, however it is not showing up in the Trino UI. I will remove the comment in a preparatory commit.

tests/integration/test_dbapi_integration.py
```diff
@@ -885,6 +897,12 @@ def test_eager_loading_cursor_description(trino_connection):
     assert all([a == e] for a, e in zip(description_after, description_expected))
 
 
+def test_catalogs(trino_connection):
```
ebyhr, line 900: What's the purpose of this test? Can you add assertion?

tests/integration/test_dbapi_integration.py
```diff
@@ -885,6 +897,12 @@ def test_eager_loading_cursor_description(trino_connection):
...
+    cur = trino_connection.cursor()
+    cur.execute("show catalogs")
```
ebyhr, line 902: nit: Uppercase "show catalogs".

ebyhr (approved):

tests/integration/test_dbapi_integration.py
mdesmet: Sorry that was a trace of a test I did. I have removed it.

ebyhr:
Merged, thanks!

Va1-ha11a:
@ebyhr  started having this problem with Trino installed in docker today, first time ever, never happened since I started the trino container. What should I do to resolve?

Issue #199, opened 2022-07-15
DB API raises exceptions which are not derived from the PEP 249 database API specification


john-bodley:
The Trino DB API raises exceptions outside the [PEP 249](https://peps.python.org/pep-0249/) database API specification and thus are not correctly handled by downstream interfaces including SQLAlchemy, Apache Superset, etc.

For example the DB API cursor `fetchall` method is raising a `TrinoUserError` which is not derived from a PEP 249 exception.

```
  File "/usr/local/lib/python3.9/dist-packages/trino/dbapi.py", line 560, in fetchall
    return list(self.genall())
  File "/usr/local/lib/python3.9/dist-packages/trino/client.py", line 512, in __iter__
    rows = self._query.fetch()
  File "/usr/local/lib/python3.9/dist-packages/trino/client.py", line 685, in fetch
    status = self._request.process(response)
  File "/usr/local/lib/python3.9/dist-packages/trino/client.py", line 443, in process
    raise self._process_error(response["error"], response.get("id"))
trino.exceptions.TrinoUserError: TrinoUserError(type=USER_ERROR, name=NOT_SUPPORTED, message="line 1:1: Relation 'foo.bar' is a table, not a view", query_id=20220715_233855_30555_sz8ja)
```

The resolution is that the public DB API methods should only raise PEP 249 exceptions.

hashhar:
> The resolution is that the public DB API methods should only raise PEP 249 exceptions.

Or change existing exceptions to be subclasses of the predefined ones. Thanks for reporting this. Can you elaborate a bit on what you mean by "are not correctly handled by downstream interfaces including SQLAlchemy, Apache Superset, etc."? What do you observe?

john-bodley:
@hashhar see https://github.com/apache/superset/pull/20729 for details, specifically [this](https://github.com/apache/superset/pull/20729/files#diff-b7a7129a51cf2a1f5021fc5244dbdda8df6c6c7207dfe715eaf97b1e1c4a0655R954-R956) comment.

PR #201, opened 2022-07-18
[CI] Migrate to pyproject.toml and use poetry for deterministic builds

john-bodley:
Somewhat of a yak shaving exercise as I was initially going to look into https://github.com/trinodb/trino-python-client/issues/199 though realized that the `trino-python-client` package has no deterministic dependency management. 

There's a slew of frameworks for handling Python dependencies including Pipenv, pip-tools, etc. however it seems like Poetry is the flavor of the month and integrates with `pyproject.toml`—a replacement for the legacy `setup.py`. 

This PR:

1. Deprecates the `setup.py` in favor of `pyproject.toml`. 
2. Adds Poetry for ensuring deterministic dependency management which generates a completely exhaustive frozen set of dependencies in the `poetry.lock` file. 
3. Updates the Github workflows to invoke `tox`—ensuring a consistent local and remote CI workflow. 

to: @bitsondatadev @hashhar


.github/workflows/ci.yml
```diff
@@ -23,12 +23,6 @@ jobs:
       - name: "Install Python"
         uses: actions/setup-python@v2
 
-      - name: "Install pre-commit"
-        run: pip install pre-commit
-
-      - name: "Run pre-commit checks"
-        run: pre-commit run --all-files
```
hashhar, line 30: We generally keep linting and static analysis jobs separate from jobs which run tests. Makes it easier to see what failed.

It looks like now pre-commit runs as part of the test jobs themselves.

.github/workflows/ci.yml
```diff
@@ -59,7 +54,8 @@ jobs:
...
           sudo apt-get update
           sudo apt-get install libkrb5-dev
-          pip install .[tests]
-      - name: Run tests
+          sudo curl -sSL https://install.python-poetry.org | python3 - --preview
+          poetry install --only=dev
+      - name: Run tox
         run: |
-          pytest -s tests/
+          poetry run tox --parallel
```
hashhar, line 61: Why is `--parallel` needed?

README.md
```diff
@@ -175,9 +175,9 @@ the [`JWT` authentication type](https://trino.io/docs/current/security/jwt.html)
 
     ```python
-    
+
     engine = create_engine("trino://<username>@<host>:<port>/<catalog>/<schema>?access_token=<jwt_token>")
-  
+
```
hashhar, line 180: extract no-op changes to separate commit.

README.md
```diff
@@ -494,11 +488,7 @@ They pull a Docker image and then run a container with a Trino server:
   ```
 - Create release package and upload it to PyPI
   ```bash
-  . .venv/bin/activate &&
-  pip install twine &&
-  rm -rf dist/ &&
-  ./setup.py sdist bdist_wheel &&
-  twine upload dist/* &&
+  poetry publish --build &&
```
hashhar, line 491: any possibility of clearing old state/artifacts for safety?

.pre-commit-config.yaml
```diff
@@ -13,3 +13,8 @@ repos:
         additional_dependencies:
           - "types-pytz"
           - "types-requests"
+
+  - repo: "https://github.com/python-poetry/poetry"
+    rev: "1.2.0b3"
```
hovaesco, line 18: Let's switch to stable release before merging

README.md
```diff
@@ -417,29 +417,14 @@ assert cur.description[0][1] == "timestamp with time zone"
...
+Clone the repository and go inside the code directory.
 
-We recommend that you use Python3's `venv` for development:
+Python dependencies are managed using [Poetry](https://python-poetry.org/) which helps to ensure the project is managed in a deterministic way. Poetry [creates a virtual environment](https://python-poetry.org/docs/managing-environments/) to aid with the process. Currently this project uses [dependency groups](https://python-poetry.org/docs/master/managing-dependencies/) which are a pre-release feature (denoted by the `--preview` flag) and thus Poetry should be installed via:
 
 ```
-$ python3 -m venv .venv
-$ . .venv/bin/activate
-$ pip install .
+$ curl -sSL https://install.python-poetry.org | python3 - --preview
```
hovaesco, line 425: Let's wait for the official release before merging.

pyproject.toml
```diff
@@ -0,0 +1,66 @@
+[build-system]
+requires = ["poetry-core>=1.1.0b3"]
```
hovaesco, line 2: Should it be `1.2.0b3` as for now?

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry]
+name = "trino-python-client"
+version = "0.314.0"
```
hovaesco, line 7: We need some changes in versioning guides in README.md.

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+
+include = ["LICENSE"]
+
+[tool.poetry.dependencies]
+keyring = { version = "*", optional = true }
+python = "^3.7"
+pytz = "*"
+requests = "*"
+requests_kerberos = { version = "*", optional = true }
+sqlalchemy = { version = "~1.4", optional = true }
```
hovaesco, line 45: It was ~1.3 in setup.py previously, is there any specific reason for that change?

.github/workflows/ci.yml
john-bodley: @hashhar rather than multiple GitHub jobs this workflow is configured to have only one job and thus from a wall-clock perspective having the relevant tox environments run in parallel is more desirable.

README.md
john-bodley: @hashhar this is likely a default from my editor. I'll revert it. Adding the `prettier` pre-commit hook helps to ensure files are formatted in a consistent manner which long term would prevent these types of no-op changes.

README.md
```diff
@@ -450,32 +435,41 @@ When the code is ready, submit a Pull Request.
 
 ### Running Tests
 
-`trino-python-client` uses [pytest](https://pytest.org/) for its tests. To run
+`trino-python-client` uses [tox](https://tox.wiki/en/latest/)—a tool for standardizing testing in Python—which leverages the [pytest](https://pytest.org/) testing framework. To run
 only unit tests, type:
 
 ```
-$ pytest tests/unit
+$ poetry run tox -e <environment> -- tests/unit
```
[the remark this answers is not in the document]
john-bodley: @hashhar local development likely wants to run multiple environments, i.e., `py39`, `pre-commit`, etc. Personally having used `tox` for a number of years I have no issue with explicitly invoking the Python specific environment when running the unit/integration tests.

README.md
john-bodley: @hashhar I'm unsure. There's https://github.com/python-poetry/poetry/issues/1329 which has been open for some time.

.pre-commit-config.yaml
john-bodley: @hovaesco the reason for including a pre-release it it exposes [dependency groups](https://python-poetry.org/docs/master/managing-dependencies/). It's not clear from https://github.com/python-poetry/poetry/issues/5586 when the 1.2.0 release is scheduled.

Are there any major concerns with using a pre-release in the interim especially if it seems to fulfill our needs?

pyproject.toml
john-bodley: @hovaesco could you provide more context about what is required. I merely lifted this from [here](https://github.com/trinodb/trino-python-client/blob/master/trino/__init__.py#L22) which doesn't seem to be referenced elsewhere.

pyproject.toml
john-bodley: @hovaesco this is _precisely_ the reason for this change, i.e., to ensure deterministic builds. Specifically when this was set to `~1.3` running, 

```
poetry run tox -e py39 -- tests/unit
```

would result in the following error:

```
Traceback:
../../../.pyenv/versions/3.9.4/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/unit/sqlalchemy/test_dialect.py:5: in <module>
    from sqlalchemy.engine import make_url
E   ImportError: cannot import name 'make_url' from 'sqlalchemy.engine'
```

In SQLAlchemy v1.4 they exposed the `make_url` function to the to the [sqlachemy.engine](https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4/lib/sqlalchemy/engine/__init__.py#L59) module as previously in v1.3 one would need to import this via, 

```
from sqlalchemy.engine.url import make_url
```

i.e., v1.3.* is not viable given how the current code is written. This logic was added in https://github.com/trinodb/trino-python-client/commit/da1441f1083ef1e908f4a8d68d2ef80491157944 which defined the `sqlalechemy~=1.3` requirement (note the inclusion of the `=` which differs from the Poetry dependency), which—per PEP 440—implies,

> For a given release identifier V.N , the compatible release clause is approximately equivalent to the pair of comparison clauses:
>
> `>= V.N, == V.*`

thus the `sqlalechemy~=1.3` requirement is equivalent to `sqlachemy>=1.3, == 1.*`. If you look at a GitHub workflow CI run ([example](https://github.com/trinodb/trino-python-client/runs/7476040790?check_suite_focus=true)) you can see, 

```
Collecting sqlalchemy~=1.3
  Downloading SQLAlchemy-1.4.39-cp39-cp39-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl (1.6 MB)
 ```
 
that the CI is installing the latest version (v1.4.39) which adheres to the requirement specification.

Note in Poetry the [tilde requirements](https://python-poetry.org/docs/dependency-specification/#tilde-requirements)  (which don't include the `=`) differ slightly, so `~1.3` is equivalent to `>= 1.3.0, < 1.4.0`.

File: README.md (hunk 6 above)
Comment (john-bodley) [reply README.md:425]: See previous comment.

File: README.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (hashhar) [reply README.md:442]: We should reduce friction for new contributors. Generally people shouldn't need to care what env they are testing against unless the issue is env specific (then they can provide their own choice of env).

File: .github/workflows/ci.yml (hunk 2 above)
Comment (hashhar) [reply .github/workflows/ci.yml:61]: Having single job makes it harder to inspect which env is failing without having to go to the actions page, and check the logs. Any way we can preserve current matrix behaviour?

File: .github/workflows/ci.yml (hunk 1 above)
Comment (hashhar) [reply .github/workflows/ci.yml:30]: While that is a reasonable benefit it has it's downsides as well - namely that in case of failures you need to go to the CI run logs and see what exactly failed and unless you know what to search/grep for it's painful.

File: pyproject.toml (hunk 8 above)
Comment (hovaesco) [reply pyproject.toml]: Please take a look [here](https://github.com/trinodb/trino-python-client/blob/master/README.md?plain=1#L486)

File: pyproject.toml (hunk 9 above)
Comment (hovaesco) [reply pyproject.toml:45]: I see, thanks for the detailed explanation.

File: .github/workflows/ci.yml (hunk 1 above)
Comment (john-bodley) [reply .github/workflows/ci.yml:30]: There are pros and cons with everything. In general adhering to the KISS principle—simplifying logic and ensure consistency—likely out weights the cons. Note it's worth noting the errors are present in the logs, albeit maybe having to toggle an additional chevron.

File: .github/workflows/ci.yml (hunk 2 above)
Comment (john-bodley) [reply .github/workflows/ci.yml:61]: See https://github.com/trinodb/trino-python-client/pull/201#discussion_r931252809. I'm really not certain whether the hypothesized concerns are problematic in reality. There's very little friction in introspecting the GitHub workflow logs to determine why CI is failing.

File: README.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (john-bodley) [reply README.md:442]: I'm not really sure explicitly specifying an environment is friction. It's worth noting that local CI likely requires multiple environments, i.e., a common workflow could be to first run `tox -e pre-commit` to ensure linting, typing, etc. is correct before running the test via `tox -e py39` say.

One could argue having a these environments—a subset of which can be run via the `-e` argument—actually reduces friction as it obfuscates the actual commands from the user, i.e., rather than the user needing to know they need to run the `pytest`, `pre-commit`, etc. commands (along with the relevant flags/arguments) they simply invoke the respective environment through the `tox` CLI which standardizes testing workflows.

**Comment by john-bodley** [comment]
Thanks @hashhar and @hovaesco for the review. I’ve hopefully answered/addressed all your comments/questions.

I opted to downgrade the version of Poetry (from a pre-release to the latest stable version)—albeit it being less feature rich—to unblock the PR.

File: README.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (hovaesco) [reply README.md:442]: Perhaps adding some examples like `poetry run tox -e py39 -- tests/unit` could be a reasonable consensus? Contributors usually follow README first, or we can have a sample local testing workflow described like run pre-commit then running tests etc.

File: .github/workflows/ci.yml [hunk 11]
```diff
@@ -23,12 +23,6 @@ jobs:
       - name: "Install Python"
         uses: actions/setup-python@v2
 
-      - name: "Install pre-commit"
-        run: pip install pre-commit
-
-      - name: "Run pre-commit checks"
-        run: pre-commit run --all-files
-
```
hovaesco, line 31: If those steps are removed, then I assume `checks` jobs is redundant. Outstanding jobs step be moved to `build` job.

hovaesco: However, it would be nice to still execute pre-commit checks in `checks` job. Then `build` could be run only if `checks` job is successful. It will reduce runners usage.

hovaesco (commented):
One general comment, could you please squash commits (if possible) and rename commit messages?

README.md
hashhar: Let's keep the `rm -rf dist/` in the steps - I verified that Poetry indeed doesn't clean up old artifacts.

.github/workflows/ci.yml
john-bodley: @hovaesco is it necessary to split out the checks vs the build? I agree that one could argue that `pre-commit` performs pre-flight checks, however i) it does mean the DRY principle isn't really adhered to, and ii) you could argue (per the name) that these should actually be first run locally (alongside all the other local CI checks) at the time of the commit. This is actually the most efficient use of resources and results in faster iteration.

john-bodley:
@hovaesco thanks for the review. Hopefully I've addressed your comments.

.github/workflows/ci.yml
hashhar: You cannot control behaviour of people who submit PRs. It's still useful to let people submit PRs in an imperfect shape and guide them to the finish line so that their next contribution is easier.

Also keeping the checks split has two benefits:
- Lesser wall-time for each check.
- Easy to see at a glance what failed with zero additional clicks (that's also why a matrix was used instead of tox for the version based tests)

iamannalynleynes1206 (approved):

PR #202, opened 2022-07-18
Support for time, datetime, json, real, double types

lpoulain:
This PR adds support for time, datetime, json, read and double Trino types when querying for some rows.

The `time` and `datetime` required some data massaging:

- The Python precision for datetime's milliseconds is maximum 6 digits, so any extra digits (such as the last 3 digits from a  `datetime(9)`) are dropped
- The time in Python does not have any timezone, so the timezone is dropped


mdesmet:
The goal of this code is similar as what is implemented for `experimental_python_types`. The idea behind using this flag is that it would not impact any existing code relying on the string output of the Trino API (backwards compatibllity).

Could you review the existing mapping code and correct any shortcomings that you may have fixed.

https://github.com/trinodb/trino-python-client/blob/951ad82b7a7571c99dddf4e35f66fb403f052c11/trino/client.py#L583-L651

AFAIK we didn't do unit testing for the types. So I think that's probably something we should keep.

tests/unit/test_types.py
```diff
@@ -0,0 +1,1852 @@
...
+    None,
+    'AQgAAAACCwEAgANEAAAgAAABAAAASsQF+7cDRAABAA==',
+    None,
+    'AHsUrkfheoQ/AAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAQAAAAAAAAABAAAAAAAAAAAAAPA/AQAAAAAAAIA=',
+    None,
+    'AAAAAAAAAPA/AAAAAAAA8D8AAAAAAABZQAAAAAAAAPA/AQAAAAAAAAAAAPA/AAAAAAAA8D8=',
+    None
+]
+
+def test_types():
```
mdesmet, line 1836: Please split the tests by type, that will be easier to troubleshoot when things go wrong.

mdesmet: Also test using  the flag `experimental_python_types` set to `True` with the `TrinoResult` class.

mdesmet (changes requested):

trino/client.py
```diff
@@ -226,6 +227,66 @@ def __repr__(self):
             )
         )
 
+    def __process_rows(self, rows, columns):
```
mdesmet, line 230: Should probably just be a private function instead of the double underscore of a Python hook.

tests/unit/test_types.py
lpoulain: Done

trino/client.py
lpoulain: Fixed

mdesmet:
As a general remark, please rebase instead of merge changes from master, otherwise you will introduce already implemented changes in your PR's commits.

trino/client.py
```diff
@@ -588,100 +662,42 @@ def __iter__(self):
...
-                yield self._map_row(self._experimental_python_types, row, self._query.columns)
+                yield self._map_row(self._experimental_python_types, row, col_mapping_funcs)
 
     @property
     def response_headers(self):
         return self._query.response_headers
 
     @classmethod
-    def _map_row(cls, experimental_python_types, row, columns):
+    def _map_row(cls, experimental_python_types, row, col_mapping_func):
```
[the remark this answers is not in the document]
lpoulain: It could, but then the whole body of `_map_to_python_type()` would be repeated inside each mapping function.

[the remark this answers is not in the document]
mdesmet: Sorry I meant `_map_to_python_type`, that logic seems more well placed in the col_mapping_func. 

The actual mapping I would do in the `TrinoQuery.fetch()` method. That way the TrinoResult is just returning the data from the API and doesn't need to know about `col_mapping_funcs` or `experimental_python_types`.

tests/integration/test_dbapi_integration.py
```diff
@@ -294,7 +294,7 @@ def test_datetime_with_utc_time_zone_query_param(trino_connection):
     rows = cur.fetchall()
 
     assert rows[0][0] == params
-    assert cur.description[0][1] == "timestamp with time zone"
+    assert cur.description[0][1] == "timestamp(6) with time zone"
```
mdesmet, line 297: Note that only milliseconds are currently supported. If you would change the test case to more precise values over ms, eg. `params = datetime(2020, 1, 1, 16, 43, 22, 320258)`, you would see that 258 part is lost, because of #42.

IMHO providing a value with a precision that is not actually supported right now seems not so useful.

lpoulain: Actually the whole 6 digits are preserved now.

trino/client.py
```diff
@@ -229,6 +231,84 @@ def __repr__(self):
             )
         )
 
+    def _col_func(self, column):
```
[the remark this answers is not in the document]
lpoulain: Processing the `column` section (to get a list of mapping lambdas) and processing the rows can be done in multiple locations. I performed the former in the `TrinoStatus` class because this is where the data is first received. Processing the rows is done downstream, so left it there.

As far as the best candidate to perform both operations, I don't have any string opinion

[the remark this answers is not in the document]
mdesmet: By extracting it to its own class it would allow for separation of concern of each class and put it in the most appropriate place (composition).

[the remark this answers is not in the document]
lpoulain: Fair enough. But there is still the question about whether the location where mapping function generation and execution should take place.

trino/client.py
```diff
@@ -229,6 +231,86 @@ def __repr__(self):
...
+            return lambda values: tuple(elt_mapping_funcs[idx](value) for idx, value in enumerate(values))
+        elif col_type == 'map':
+            key_mapping_func = self._col_func(column['arguments'][0]['value'])
+            value_mapping_func = self._col_func(column['arguments'][1]['value'])
+            return lambda values: {key_mapping_func(key): value_mapping_func(value) for key, value in values.items()}
+        elif col_type.startswith('decimal'):
+            return lambda val: Decimal(val)
+        elif col_type.startswith('double') or col_type.startswith('real'):
+            return lambda val: float('inf') if val == 'Infinity' \
+                else -float('inf') if val == '-Infinity' \
```
mdesmet, line 251: Use the existing constants for performance.

```
INF = float("inf")
NEGATIVE_INF = float("-inf")
NAN = float("nan")
```

trino/client.py
```diff
@@ -339,6 +421,7 @@ def __init__(
             self._http_session = self.http.Session()
             self._http_session.verify = verify
         self._http_session.headers.update(self.http_headers)
+        self._http_session.headers['x-trino-client-capabilities'] = 'PARAMETRIC_DATETIME'
```
mdesmet, line 424: This change should probably go into another PR. See #114

trino/client.py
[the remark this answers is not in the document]
mdesmet: Sorry I meant `_map_to_python_type`. You could use the `experimental_python_types` in the col_mapping_funcs itself, that way the `TrinoResult` class just needs to know about the col_mapping_funcs.

lpoulain:
Closing to split the PR is smaller chunks

PR #206, opened 2022-07-21
Optimize the experimental python types flag

lpoulain:
Optimizes the `experimental_python_types` flag. Instead of checking the type for each row, check the type once for each `fetch()` call and compute a list of lambdas which are to be applied to the values from each row.

- A new `RowMapperFactory` class was created to wrap this behavior. The `columns` content from the status is analyzed and a list of lambda functions is created (one lambda per column)
- The `experimental_python_types` flag is now processed in the `TrinoQuery` class instead of the `TrinoResult` class

### DIFF
--- a/tests/integration/test_types_integration.py
+++ b/tests/integration/test_types_integration.py
@@ -1,0 +1,247 @@
+import math
+import pytest
+from decimal import Decimal
+import trino
+
+
+@pytest.fixture
+def trino_connection(run_trino):
+    _, host, port = run_trino
+
+    yield trino.dbapi.Connection(
+        host=host, port=port, user="test", source="test", max_attempts=1
+    )
+
+
+def test_boolean(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS BOOLEAN)", python=None) \
+        .add_field(sql="false", python=False) \
+        .add_field(sql="true", python=True) \
+        .execute()
+
+
+def test_tinyint(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS TINYINT)", python=None) \
+        .add_field(sql="CAST(-128 AS TINYINT)", python=-128) \
+        .add_field(sql="CAST(42 AS TINYINT)", python=42) \
+        .add_field(sql="CAST(127 AS TINYINT)", python=127) \
+        .execute()
+
+
+def test_smallint(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS SMALLINT)", python=None) \
+        .add_field(sql="CAST(-32768 AS SMALLINT)", python=-32768) \
+        .add_field(sql="CAST(42 AS SMALLINT)", python=42) \
+        .add_field(sql="CAST(32767 AS SMALLINT)", python=32767) \
+        .execute()
+
+
+def test_int(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS INTEGER)", python=None) \
+        .add_field(sql="CAST(-2147483648 AS INTEGER)", python=-2147483648) \
+        .add_field(sql="CAST(83648 AS INTEGER)", python=83648) \
+        .add_field(sql="CAST(2147483647 AS INTEGER)", python=2147483647) \
+        .execute()
+
+
+def test_bigint(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS BIGINT)", python=None) \
+        .add_field(sql="CAST(-9223372036854775808 AS BIGINT)", python=-9223372036854775808) \
+        .add_field(sql="CAST(9223 AS BIGINT)", python=9223) \
+        .add_field(sql="CAST(9223372036854775807 AS BIGINT)", python=9223372036854775807) \
+        .execute()
+
+
+def test_real(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS REAL)", python=None) \
+        .add_field(sql="CAST('NaN' AS REAL)", python=math.nan) \
+        .add_field(sql="CAST('-Infinity' AS REAL)", python=-math.inf) \
+        .add_field(sql="CAST(3.4028235E38 AS REAL)", python=3.4028235e+38) \
+        .add_field(sql="CAST(1.4E-45 AS REAL)", python=1.4e-45) \
+        .add_field(sql="CAST('Infinity' AS REAL)", python=math.inf) \
+        .execute()
+
+
+def test_double(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS DOUBLE)", python=None) \
+        .add_field(sql="CAST('NaN' AS DOUBLE)", python=math.nan) \
+        .add_field(sql="CAST('-Infinity' AS DOUBLE)", python=-math.inf) \
+        .add_field(sql="CAST(1.7976931348623157E308 AS DOUBLE)", python=1.7976931348623157e+308) \
+        .add_field(sql="CAST(4.9E-324 AS DOUBLE)", python=5e-324) \
+        .add_field(sql="CAST('Infinity' AS DOUBLE)", python=math.inf) \
+        .execute()
+
+
+def test_decimal(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS DECIMAL)", python=None) \
+        .add_field(sql="CAST(null AS DECIMAL(38,0))", python=None) \
+        .add_field(sql="DECIMAL '10.3'", python=Decimal('10.3')) \
+        .add_field(sql="CAST('0.123456789123456789' AS DECIMAL(18,18))", python=Decimal('0.123456789123456789')) \
+        .add_field(sql="CAST(null AS DECIMAL(18,18))", python=None) \
+        .add_field(sql="CAST('234.123456789123456789' AS DECIMAL(18,4))", python=Decimal('234.1235')) \
+        .add_field(sql="CAST('10.3' AS DECIMAL(38,1))", python=Decimal('10.3')) \
+        .add_field(sql="CAST('0.123456789123456789' AS DECIMAL(18,2))", python=Decimal('0.12')) \
+        .add_field(sql="CAST('0.3123' AS DECIMAL(38,38))", python=Decimal('0.3123')) \
+        .execute()
+
+
+def test_varchar(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="'aaa'", python='aaa') \
+        .add_field(sql="U&'Hello winter \2603 !'", python='Hello winter °3 !') \
+        .add_field(sql="CAST(null AS VARCHAR)", python=None) \
+        .add_field(sql="CAST('bbb' AS VARCHAR(1))", python='b') \
+        .add_field(sql="CAST(null AS VARCHAR(1))", python=None) \
+        .execute()
+
+
+def test_char(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST('ccc' AS CHAR)", python='c') \
+        .add_field(sql="CAST(null AS CHAR)", python=None) \
+        .add_field(sql="CAST('ddd' AS CHAR(1))", python='d') \
+        .add_field(sql="CAST('😂' AS CHAR(1))", python='😂') \
+        .add_field(sql="CAST(null AS CHAR(1))", python=None) \
+        .execute()
+
+
+def test_varbinary(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="X'65683F'", python='ZWg/') \
+        .add_field(sql="X''", python='') \
+        .add_field(sql="CAST('' AS VARBINARY)", python='') \
+        .add_field(sql="from_utf8(CAST('😂😂😂😂😂😂' AS VARBINARY))", python='😂😂😂😂😂😂') \
+        .add_field(sql="CAST(null AS VARBINARY)", python=None) \
+        .execute()
+
+
+def test_varbinary_failure(trino_connection):
+    SqlExpectFailureTest(trino_connection) \
+        .execute("CAST(42 AS VARBINARY)")
+
+
+def test_json(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST('{}' AS JSON)", python='"{}"') \
+        .add_field(sql="CAST('null' AS JSON)", python='"null"') \
+        .add_field(sql="CAST(null AS JSON)", python=None) \
+        .add_field(sql="CAST('3.14' AS JSON)", python='"3.14"') \
+        .add_field(sql="CAST('a string' AS JSON)", python='"a string"') \
+        .add_field(sql="CAST('a \" complex '' string :' AS JSON)", python='"a \\" complex \' string :"') \
+        .add_field(sql="CAST('[]' AS JSON)", python='"[]"') \
+        .execute()
+
+
+def test_interval(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS INTERVAL YEAR TO MONTH)", python=None) \
+        .add_field(sql="CAST(null AS INTERVAL DAY TO SECOND)", python=None) \
+        .add_field(sql="INTERVAL '3' MONTH", python='0-3') \
+        .add_field(sql="INTERVAL '2' DAY", python='2 00:00:00.000') \
+        .add_field(sql="INTERVAL '-2' DAY", python='-2 00:00:00.000') \
+        .execute()
+
+
+def test_array(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS ARRAY(VARCHAR))", python=None) \
+        .add_field(sql="ARRAY['a', 'b', null]", python=['a', 'b', None]) \
+        .execute()
+
+
+def test_map(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS MAP(VARCHAR, INTEGER))", python=None) \
+        .add_field(sql="MAP(ARRAY['a', 'b'], ARRAY[1, null])", python={'a': 1, 'b': None}) \
+        .execute()
+
+
+def test_row(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS ROW(x BIGINT, y DOUBLE))", python=None) \
+        .add_field(sql="CAST(ROW(1, 2e0) AS ROW(x BIGINT, y DOUBLE))", python=(1, 2.0)) \
+        .execute()
+
+
+def test_ipaddress(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS IPADDRESS)", python=None) \
+        .add_field(sql="IPADDRESS '2001:db8::1'", python='2001:db8::1') \
+        .execute()
+
+
+def test_uuid(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS UUID)", python=None) \
+        .add_field(sql="UUID '12151fd2-7586-11e9-8f9e-2a86e4085a59'", python='12151fd2-7586-11e9-8f9e-2a86e4085a59') \
+        .execute()
+
+
+def test_digest(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS HyperLogLog)", python=None) \
+        .add_field(sql="CAST(null AS P4HyperLogLog)", python=None) \
+        .add_field(sql="CAST(null AS SetDigest)", python=None) \
+        .add_field(sql="CAST(null AS QDigest(BIGINT))", python=None) \
+        .add_field(sql="CAST(null AS TDigest)", python=None) \
+        .add_field(sql="approx_set(1)", python='AgwBAIADRAA=') \
+        .add_field(sql="CAST(approx_set(1) AS P4HyperLogLog)", python='AwwAAAAg' + 'A' * 2730 + '==') \
+        .add_field(sql="make_set_digest(1)", python='AQgAAAACCwEAgANEAAAgAAABAAAASsQF+7cDRAABAA==') \
+        .add_field(sql="tdigest_agg(1)",
+                   python='AAAAAAAAAPA/AAAAAAAA8D8AAAAAAABZQAAAAAAAAPA/AQAAAAAAAAAAAPA/AAAAAAAA8D8=') \
+        .execute()
+
+
+class SqlTest:
+    def __init__(self, trino_connection):
+        self.cur = trino_connection.cursor(experimental_python_types=True)
+        self.sql_args = []
+        self.expected_result = []
+
+    def add_field(self, sql, python):
+        self.sql_args.append(sql)
+        self.expected_result.append(python)
+        return self
+
+    def execute(self):
+        sql = 'SELECT ' + ',\n'.join(self.sql_args)
+
+        self.cur.execute(sql)
+        actual_result = self.cur.fetchall()
+        self._compare_results(actual_result[0], self.expected_result)
+
+    def _compare_results(self, actual, expected):
+        assert len(actual) == len(expected)
+
+        for idx, actual_val in enumerate(actual):
+            expected_val = expected[idx]
+            if type(actual_val) == float and math.isnan(actual_val) \
+               and type(expected_val) == float and math.isnan(expected_val):
+                continue
+
+            assert actual_val == expected_val
+
+
+class SqlExpectFailureTest:
+    def __init__(self, trino_connection):
+        self.cur = trino_connection.cursor(experimental_python_types=True)
+
+    def execute(self, field):
+        sql = 'SELECT ' + field
+
+        try:
+            self.cur.execute(sql)
+            self.cur.fetchall()
+            success = True
+        except Exception:
+            success = False
+
+        assert not success, "Test not expected to succeed"
--- a/trino/client.py
+++ b/trino/client.py
@@ -574,11 +574,10 @@ class TrinoResult(object):
     https://docs.python.org/3/library/stdtypes.html#generator-types
     """
 
-    def __init__(self, query, rows=None, experimental_python_types: bool = False):
+    def __init__(self, query, rows=None):
         self._query = query
         self._rows = rows or []
         self._rownumber = 0
-        self._experimental_python_types = experimental_python_types
 
     @property
     def rownumber(self) -> int:
@@ -588,7 +587,7 @@ class TrinoResult(object):
         # Initial fetch from the first POST request
         for row in self._rows:
             self._rownumber += 1
-            yield self._map_row(self._experimental_python_types, row, self._query.columns)
+            yield row
         self._rows = None
 
         # Subsequent fetches from GET requests until next_uri is empty.
@@ -597,92 +596,11 @@ class TrinoResult(object):
             for row in rows:
                 self._rownumber += 1
                 logger.debug("row %s", row)
-                yield self._map_row(self._experimental_python_types, row, self._query.columns)
+                yield row
 
     @property
     def response_headers(self):
         return self._query.response_headers
-
-    @classmethod
-    def _map_row(cls, experimental_python_types, row, columns):
-        if not experimental_python_types:
-            return row
-        else:
-            return cls._map_to_python_types(cls, row, columns)
-
-    @classmethod
-    def _map_to_python_type(cls, item: Tuple[Any, Dict]) -> Any:
-        (value, data_type) = item
-
-        if value is None:
-            return None
-
-        raw_type = data_type["typeSignature"]["rawType"]
-
-        try:
-            if isinstance(value, list):
-                if raw_type == "array":
-                    raw_type = {
-                        "typeSignature": data_type["typeSignature"]["arguments"][0]["value"]
-                    }
-                    return [cls._map_to_python_type((array_item, raw_type)) for array_item in value]
-                if raw_type == "row":
-                    raw_types = map(lambda arg: arg["value"], data_type["typeSignature"]["arguments"])
-                    return tuple(
-                        cls._map_to_python_type((array_item, raw_type))
-                        for (array_item, raw_type) in zip(value, raw_types)
-                    )
-                return value
-            if isinstance(value, dict):
-                raw_key_type = {
-                    "typeSignature": data_type["typeSignature"]["arguments"][0]["value"]
-                }
-                raw_value_type = {
-                    "typeSignature": data_type["typeSignature"]["arguments"][1]["value"]
-                }
-                return {
-                    cls._map_to_python_type((key, raw_key_type)):
-                        cls._map_to_python_type((value[key], raw_value_type))
-                    for key in value
-                }
-            elif "decimal" in raw_type:
-                return Decimal(value)
-            elif raw_type == "double":
-                if value == 'Infinity':
-                    return INF
-                elif value == '-Infinity':
-                    return NEGATIVE_INF
-                elif value == 'NaN':
-                    return NAN
-                return value
-            elif raw_type == "date":
-                return datetime.strptime(value, "%Y-%m-%d").date()
-            elif raw_type == "timestamp with time zone":
-                dt, tz = value.rsplit(' ', 1)
-                if tz.startswith('+') or tz.startswith('-'):
-                    return datetime.strptime(value, "%Y-%m-%d %H:%M:%S.%f %z")
-                return datetime.strptime(dt, "%Y-%m-%d %H:%M:%S.%f").replace(tzinfo=pytz.timezone(tz))
-            elif "timestamp" in raw_type:
-                return datetime.strptime(value, "%Y-%m-%d %H:%M:%S.%f")
-            elif "time with time zone" in raw_type:
-                matches = re.match(r'^(.*)([\+\-])(\d{2}):(\d{2})$', value)
-                assert matches is not None
-                assert len(matches.groups()) == 4
-                if matches.group(2) == '-':
-                    tz = -timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
-                else:
-                    tz = timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
-                return datetime.strptime(matches.group(1), "%H:%M:%S.%f").time().replace(tzinfo=timezone(tz))
-            elif "time" in raw_type:
-                return datetime.strptime(value, "%H:%M:%S.%f").time()
-            else:
-                return value
-        except ValueError as e:
-            error_str = f"Could not convert '{value}' into the associated python type for '{raw_type}'"
-            raise trino.exceptions.TrinoDataError(error_str) from e
-
-    def _map_to_python_types(self, row: List[Any], columns: List[Dict[str, Any]]) -> List[Any]:
-        return list(map(self._map_to_python_type, zip(row, columns)))
 
 
 class TrinoQuery(object):
@@ -705,9 +623,10 @@ class TrinoQuery(object):
         self._request = request
         self._update_type = None
         self._sql = sql
-        self._result = TrinoResult(self, experimental_python_types=experimental_python_types)
+        self._result = TrinoResult(self)
         self._response_headers = None
         self._experimental_python_types = experimental_python_types
+        self._row_mapper: Optional[RowMapper] = None
 
     @property
     def columns(self):
@@ -758,12 +677,18 @@ class TrinoQuery(object):
         self._warnings = getattr(status, "warnings", [])
         if status.next_uri is None:
             self._finished = True
-        self._result = TrinoResult(self, status.rows, self._experimental_python_types)
+
+        rows = self._row_mapper.map(status.rows) if self._row_mapper else status.rows
+
+        self._result = TrinoResult(self, rows)
         return self._result
 
     def _update_state(self, status):
         self._stats.update(status.stats)
         self._update_type = status.update_type
+        if not self._row_mapper and status.columns:
+            self._row_mapper = RowMapperFactory().create(columns=status.columns,
+                                                         experimental_python_types=self._experimental_python_types)
         if status.columns:
             self._columns = status.columns
 
@@ -776,7 +701,11 @@ class TrinoQuery(object):
         self._response_headers = response.headers
         if status.next_uri is None:
             self._finished = True
-        return status.rows
+
+        if not self._row_mapper:
+            return []
+
+        return self._row_mapper.map(status.rows)
 
     def cancel(self) -> None:
         """Cancel the current query"""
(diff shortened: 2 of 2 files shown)


tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,314 @@
...
+        None,
+        32767,
+        -32768,
+        None,
+        2147483647,
+        -2147483648,
+        None,
+        9223372036854775807,
+        -9223372036854775808,
+        None
```
ebyhr, line 51: Can you introduce a helper class/method that is similar to [SqlDataTypeTest](https://github.com/trinodb/trino/blob/master/testing/trino-testing/src/main/java/io/trino/testing/datatype/SqlDataTypeTest.java) in Trino? It will improve the readability a lot.

lpoulain: Do you mean a class/method where you only pass the input SQL and expected output, or something different?

ebyhr: "only pass the input SQL and expected output" is right.

The below is Trino example using `SqlDataTypeTest`. There're 4 arguments in `addRoundTrip()`, but it will be 3 arguments in this Python test.
```
    @Test
    public void testBit()
    {
        SqlDataTypeTest.create()
                .addRoundTrip("bit", "b'1'", BOOLEAN, "true")
                .addRoundTrip("bit", "b'0'", BOOLEAN, "false")
                .addRoundTrip("bit", "NULL", BOOLEAN, "CAST(NULL AS BOOLEAN)")
                .execute(getQueryRunner(), mysqlCreateAndInsert("tpch.test_bit"));
    }
```

lpoulain: How about something like:

```
SqlTest()
    .add_field("false AS col_bool", False)
    .add_field("cast(null AS BOOLEAN) AS col_bool_null", None)
    ...
    .execute()
```

ebyhr: Personally, I would like to omit column names because it unrelates to the test purpose. When we want to clarify the value meaning, just adding code comment seems enough to me.

lpoulain: Fair enough. But aside from that is this the type of API you had in mind?

ebyhr: Yes :)

lpoulain: Done

trino/client.py
```diff
@@ -839,3 +769,110 @@ def decorated(*args, **kwargs):
...
 
     return wrapper
+
+
+class RowMapperFactory:
+    def __init__(self, columns):
+        if columns is None:
+            self.mapping_funcs = []
+        else:
+            self.mapping_funcs = [self._col_func(column['typeSignature']) for column in columns]
```
mdesmet, line 779: The class constructor is rather used to add dependencies of the factory.

```suggestion
    def create(self, columns, experimental_python_types):
        if columns is None:
            raise ValueError("columns is not defined")
        if experimental_python_types is None:
            raise ValueError("experimental_python_types is not defined")
        if experimental_python_types:
             return RowMapper([self._col_func(column['typeSignature']) for column in columns])
        return RowMapper()
```

trino/client.py
```diff
@@ -839,3 +769,110 @@ def decorated(*args, **kwargs):
...
+
+    def _map_value(self, value, col_mapping_func):
+        if value is None:
+            return None
+
+        try:
+            return col_mapping_func(value)
+        except ValueError as e:
+            error_str = f"Could not convert '{value}' into the associated python type"
+            raise trino.exceptions.TrinoDataError(error_str) from e
```
mdesmet, line 795: Move to separate class `RowMapper`. If no col_mapping_funcs are provided in constructor, return the row value as is.

trino/client.py
```diff
@@ -776,6 +701,11 @@ def fetch(self) -> List[List[Any]]:
         self._response_headers = response.headers
         if status.next_uri is None:
             self._finished = True
+
+        if self._experimental_python_types:
+            factory = RowMapperFactory(status.columns)
```
mdesmet, line 706: `RowMapperFactory().create(...)` should only be called once when the columns are known. Maybe move this logic to the `_update_state` method. Add the created `RowMapper` instance on an instance field of `TrinoQuery`.

trino/client.py
```diff
@@ -758,7 +676,14 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
...
             self._finished = True
-        self._result = TrinoResult(self, status.rows, self._experimental_python_types)
+
+        if self._experimental_python_types:
+            factory = RowMapperFactory(status.columns)
+            rows = factory.map_rows(status.rows)
+        else:
+            rows = status.rows
+
+        self._result = TrinoResult(self, rows)
```
mdesmet, line 686: Also applies to the `fetch` method.

```suggestion
        assert self._row_mapper is not None
        self._result = TrinoResult(self, self._row_mapper(rows))
```

trino/client.py
lpoulain: I can change the `__init__()` into `create()`, but the exceptions is a different story:

- `columns` can be legitimately be null where there is no result yet
- This class should be invoked only when `experimental_python_types` is set, so why passing it and checking it?

mdesmet: When there is no columns yet, we should not yet create the `RowMapper`. See my other remarks. 

If `experimental_python_types` is `False`, the `RowMapperFactory` can just return a no-op `RowMapper` that just returns values as is.

trino/client.py
lpoulain: Actually if we want to call `RowMapperFactory()` only once we know the column, then we shouldn't call this in `execute()`. At this point we have no guarantee we'll have columns at this point.

mdesmet: Correct, that's why I suggested to add this in the `_update_state` method to call it only when you actually have the columns.

trino/client.py

trino/client.py

trino/client.py
lpoulain: I modified the code as best as I could, because in `execute()` function expects some row (maybe) yet the columns may be not known yet.

trino/client.py

trino/client.py
lpoulain: Also, `experimental_python_types` can be set to `None` and that's a valid value.

trino/client.py
mdesmet: Yeah, the columns are only known when Trino is starting to send data, which can be in the first `execute` or in any of the subsequent `fetch` calls. We should delay the creation of the `RowMapper` until the columns are present and reuse that cached `RowMapper` on the fetched data to get the performance benefit.

hovaesco (commented):
Could you please add some docstrings/comments to the new classes?

trino/client.py
```diff
@@ -839,3 +766,118 @@ def decorated(*args, **kwargs):
...
+            return RowMapper()
+        if experimental_python_types:
+            return RowMapper([self._col_func(column['typeSignature']) for column in columns])
+        return RowMapper()
+
+    def _col_func(self, column):
+        col_type = column['rawType']
+
+        if col_type == 'array':
+            elt_mapping_func = self._col_func(column['arguments'][0]['value'])
```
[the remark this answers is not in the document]
lpoulain: That's the mapping function for each element of the array

trino/client.py
```diff
@@ -839,3 +766,118 @@ def decorated(*args, **kwargs):
...
+        if columns is None:
+            return RowMapper()
+        if experimental_python_types:
+            return RowMapper([self._col_func(column['typeSignature']) for column in columns])
+        return RowMapper()
+
+    def _col_func(self, column):
+        col_type = column['rawType']
+
+        if col_type == 'array':
```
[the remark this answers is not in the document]
mdesmet: That way we could do unit tests. especially on the date and time types, that would be beneficial.

trino/client.py
```diff
@@ -758,12 +677,17 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
...
+
+        rows = self._row_mapper.map(status.rows)
+
+        self._result = TrinoResult(self, rows)
         return self._result
 
     def _update_state(self, status):
         self._stats.update(status.stats)
         self._update_type = status.update_type
+        self._row_mapper = RowMapperFactory().create(columns=status.columns,
```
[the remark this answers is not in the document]

trino/client.py
```diff
@@ -839,3 +766,118 @@ def decorated(*args, **kwargs):
...
+                else NEGATIVE_INF if val == '-Infinity' \
+                else NAN if val == 'NaN' \
+                else float(val)
+        elif col_type.startswith('time'):
+            pattern = "%Y-%m-%d %H:%M:%S" if col_type.startswith('timestamp') else "%H:%M:%S"
+            ms_size, ms_to_trim = self._get_number_of_digits(column)
+            if ms_size > 0:
+                pattern += ".%f"
+
+            if col_type.startswith('timestamp'):
```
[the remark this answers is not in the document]

trino/client.py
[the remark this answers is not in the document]
lpoulain: Done

trino/client.py
```diff
@@ -776,7 +701,8 @@ def fetch(self) -> List[List[Any]]:
         self._response_headers = response.headers
         if status.next_uri is None:
             self._finished = True
-        return status.rows
+
+        return self._row_mapper.map(status.rows) if self._row_mapper else status.rows
```
mdesmet, line 705: At this stage the `self._row_mapper` should never be undefined. When rows are returned, the columns are also known.

trino/client.py
```diff
@@ -758,12 +677,18 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
...
+
+        rows = self._row_mapper.map(status.rows) if self._row_mapper else status.rows
+
+        self._result = TrinoResult(self, rows)
         return self._result
 
     def _update_state(self, status):
         self._stats.update(status.stats)
         self._update_type = status.update_type
+        if not self._row_mapper:
```
mdesmet, line 689: I think it's possible that the columns are not yet known at this point. The row mapper would not be recreated in case the columns arrive later.

trino/client.py
lpoulain: Actually `_row_mapper` may not be undefined at this point. This is when the query is queued and we're still waiting for results.

trino/client.py
lpoulain: That is correct, but then the `RowMapperFactory` will just return `None`

hovaesco (approved):

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+@pytest.fixture
+def trino_connection(run_trino):
+    _, host, port = run_trino
+
+    yield trino.dbapi.Connection(
+        host=host, port=port, user="test", source="test", max_attempts=1
+    )
+
+
+def test_int_types(trino_connection):
```
ebyhr, line 18: I would split into test methods per type. `test_boolean`, `test_boolean` and so on.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+        .add_field(sql="cast(null AS JSON)", python=None) \
+        .execute()
+
+
+def test_datetime_types(trino_connection):
+    the_tz = datetime.timezone(datetime.timedelta(days=-1, seconds=57600))
+
+    SqlTest(trino_connection) \
+        .add_field(sql="DATE '2001-08-22'", python=datetime.date(2001, 8, 22)) \
+        .add_field(sql="cast(null AS DATE)", python=None) \
```
ebyhr, line 84: It would be better to add more test cases for `DATE`, `TIMESTAMP` types, 4digit year, 5digit year, negative year and so on.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+        .add_field(sql="cast(null AS TIMESTAMP(12) WITH TIME ZONE)", python=None) \
+        .execute()
+
+
+def test_misc_types(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="INTERVAL '3' MONTH", python='0-3') \
+        .add_field(sql="cast(null AS INTERVAL YEAR TO MONTH)", python=None) \
+        .add_field(sql="INTERVAL '2' DAY", python='2 00:00:00.000') \
+        .add_field(sql="cast(null AS INTERVAL DAY TO SECOND)", python=None) \
```
ebyhr, line 157: Add a negative sign case for `INTERVAL` type.

ebyhr (commented):
Let me defer the final review to @hashhar.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+        host=host, port=port, user="test", source="test", max_attempts=1
+    )
+
+
+def test_int_types(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="false", python=False) \
+        .add_field(sql="cast(null AS BOOLEAN)", python=None) \
+        .add_field(sql="cast(127 AS TINYINT)", python=127) \
+        .add_field(sql="cast(-128 AS TINYINT)", python=-128) \
```
hashhar, line 23: Should also test some value not at the edge of supported range.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+        .add_field(sql="cast(4.9E-324 AS DOUBLE)", python=5e-324) \
+        .add_field(sql="cast('Infinity' AS DOUBLE)", python=math.inf) \
+        .add_field(sql="cast('-Infinity' AS DOUBLE)", python=-math.inf) \
+        .add_field(sql="cast('NaN' AS DOUBLE)", python=math.nan) \
+        .add_field(sql="cast(null AS DOUBLE)", python=None) \
+        .add_field(sql="DECIMAL '10.3'", python=Decimal('10.3')) \
+        .add_field(sql="cast(null AS DECIMAL)", python=None) \
+        .add_field(sql="cast('0.123456789123456789' AS DECIMAL(18,18))", python=Decimal('0.123456789123456789')) \
+        .add_field(sql="cast(null AS DECIMAL(18,18))", python=None) \
+        .add_field(sql="cast('10.3' AS DECIMAL(38,0))", python=10.0) \
```
hashhar, line 55: nit: in type-mapping tests like these it's generally a good idea to test values that round-trip properly. So instead of testing that `10.3` becomes `10` use a `DECIMAL(38, 1)` and test that value is retained.

Coercions should be a separate test altogether.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+        .add_field(sql="cast(null AS DECIMAL(38,0))", python=None) \
```
hashhar, line 56: Add tests for DECIMAL(38, 38) and something below DECIMAL(18, x) since above 18 everything is represented differently in Trino.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+    SqlTest(trino_connection) \
+        .add_field(sql="'aaa'", python='aaa') \
+        .add_field(sql="U&'Hello winter \2603 !'", python='Hello winter °3 !') \
+        .add_field(sql="cast(null AS VARCHAR)", python=None) \
+        .add_field(sql="cast('bbb' AS VARCHAR(1))", python='b') \
+        .add_field(sql="cast(null AS VARCHAR(1))", python=None) \
+        .add_field(sql="cast('ccc' AS CHAR)", python='c') \
+        .add_field(sql="cast(null AS CHAR)", python=None) \
+        .add_field(sql="cast('ddd' AS CHAR(1))", python='d') \
+        .add_field(sql="cast(null AS CHAR(1))", python=None) \
```
hashhar, line 70: test multibyte chars are preserved. e.g. `😂` as a `CHAR(1)` should be preserved.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+        .add_field(sql="X'65683F'", python='ZWg/') \
+        .add_field(sql="cast(null AS VARBINARY)", python=None) \
+        .add_field(sql="cast('{}' AS JSON)", python='"{}"') \
+        .add_field(sql="cast('null' AS JSON)", python='"null"') \
+        .add_field(sql="cast(null AS JSON)", python=None) \
+        .execute()
+
+
+def test_datetime_types(trino_connection):
+    the_tz = datetime.timezone(datetime.timedelta(days=-1, seconds=57600))
```
hashhar, line 80: `the_tz` is unclear what timezone it is.

Also it would be useful to parametrize this method on timezones.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
...
+        .add_field(sql="cast('2001-08-22 01:23:45.123456789 -08:00' AS TIMESTAMP(9) WITH TIME ZONE)",
+                   python=datetime.datetime(2001, 8, 22, 1, 23, 45, 123000, tzinfo=the_tz)) \
+        .add_field(sql="cast(null AS TIMESTAMP(9) WITH TIME ZONE)", python=None) \
+        .add_field(sql="cast('2001-08-22 01:23:45.123456789123 -08:00' AS TIMESTAMP(12) WITH TIME ZONE)",
+                   python=datetime.datetime(2001, 8, 22, 1, 23, 45, 123000, tzinfo=the_tz)) \
+        .add_field(sql="cast(null AS TIMESTAMP(12) WITH TIME ZONE)", python=None) \
+        .execute()
+
+
+def test_misc_types(trino_connection):
```
hashhar, line 152: Split into different ones for each tested data-type.

hashhar (commented):
A quick pass.

trino/client.py
hashhar: We should make that invalid in the connection constructor. nullable booleans is just asking for pain.

mdesmet: That might break backwards compatibility: code might not explicitely set `experimental_python_types` at both `Connection` or `Cursor` level. We could handle that in the respective constructors.

Agreed that within `RowMapperFactory` only True or False are possible values and it's driven by the `Cursor` where the logic has to take care of the flag set at `Connection` AND `Cursor` level..

hashhar: We can provide a non-None default for it in the constructors.

tests/integration/test_types_integration.py
lpoulain: DECIMAL(38,38) doesn't seem to be accepted (`DECIMAL scale must be in range [0, precision (18)]: 38`)

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,208 @@
+import math
```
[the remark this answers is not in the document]
lpoulain: When I run `pytest` locally it doesn't find it. And other tests do import `math`

tests/integration/test_types_integration.py
lpoulain: Python dates don't seem to like years which are 5-digits (`ValueError: year 10000 is out of range`) or negative (`ValueError: year -1 is out of range`)

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
hashhar: It means we're dropping the precision. Since DECIMAL(38, x) means precision = 38 but the error message says precision is 18.

So a bug to be fixed.

tests/integration/test_types_integration.py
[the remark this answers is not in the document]
hashhar: Sorry, the fact that I commented on the `math` import is just incidental.

I don't see any issues with existing tests.

```
(venv) hashhar@Ashhars-MacBook-Pro trino-python-client % tree tests/integration
tests/integration
├── __init__.py
├── conftest.py
├── test_dbapi.py
└── test_sqlalchemy_integration.py
(venv) hashhar@Ashhars-MacBook-Pro trino-python-client % pytest tests/integration
=========================================================== test session starts ===========================================================
platform darwin -- Python 3.10.6, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/hashhar/code/src/github.com/trinodb/trino-python-client
collected 95 items

tests/integration/test_dbapi.py ..................................................................................s                 [ 87%]
tests/integration/test_sqlalchemy_integration.py ............                                                                       [100%]

================================================ 94 passed, 1 skipped in 188.51s (0:03:08) ================================================
```

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+        .add_field(sql="cast(42 AS TINYINT)", python=42) \
+        .add_field(sql="cast(-42 AS TINYINT)", python=-42) \
+        .add_field(sql="cast(null AS TINYINT)", python=None) \
+        .execute()
+
+
+def test_smallint_types(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="cast(32767 AS SMALLINT)", python=32767) \
+        .add_field(sql="cast(-32768 AS SMALLINT)", python=-32768) \
```
hashhar, line 39: test a value not at the edge.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+        .add_field(sql="cast(null AS SMALLINT)", python=None) \
+        .execute()
+
+
+def test_int_types(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="cast(2147483647 AS INTEGER)", python=2147483647) \
```
hashhar, line 46: all the `add_field` calls which test min/max values could use a comment like `# min value in Trino`.

Also are there data types where the range supported by Trino is outside the range supported by the Python type? If so we should test them explicitly (also with comments like `# min value in Trino, not representable in Python` or something like that).

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+        .add_field(sql="cast(-2147483648 AS INTEGER)", python=-2147483648) \
+        .add_field(sql="cast(null AS INTEGER)", python=None) \
+        .add_field(sql="cast(9223372036854775807 AS BIGINT)", python=9223372036854775807) \
+        .add_field(sql="cast(-9223372036854775808 AS BIGINT)", python=-9223372036854775808) \
+        .add_field(sql="cast(null AS BIGINT)", python=None) \
```
hashhar, line 51: INTEGER and BIGINT are different types - split.

hashhar: Oh seems leftover/redundant since `test_bigint_types` exists.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+        .execute()
+
+
+def test_bigint_types(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="cast(9223372036854775807 AS BIGINT)", python=9223372036854775807) \
+        .add_field(sql="cast(-9223372036854775808 AS BIGINT)", python=-9223372036854775808) \
```
hashhar, line 58: test a value not at the edge.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+        .add_field(sql="cast('Infinity' AS DOUBLE)", python=math.inf) \
+        .add_field(sql="cast('-Infinity' AS DOUBLE)", python=-math.inf) \
+        .add_field(sql="cast('NaN' AS DOUBLE)", python=math.nan) \
+        .add_field(sql="cast(null AS DOUBLE)", python=None) \
+        .execute()
+
+
+def test_decimal_types(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="DECIMAL '10.3'", python=Decimal('10.3')) \
```
hashhar, line 87: In general it's easier to read such type based tests if they follow some consistent ordering.

e.g. `null` -> min value -> some random value -> max value.

Also for parametrized types like DECIMAL, TIMESTAMP it helps to keep types with same parameters together. i.e. all `DECIMAL(18,18)` are together, then others etc. etc.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+        .add_field(sql="cast(null AS DECIMAL)", python=None) \
+        .add_field(sql="cast('0.123456789123456789' AS DECIMAL(18,18))", python=Decimal('0.123456789123456789')) \
+        .add_field(sql="cast(null AS DECIMAL(18,18))", python=None) \
```
hashhar, line 90: Test decimals with precision lower than 18 as well since they are represented differently by Trino and may change how they end up serialized in future.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+
+def test_string_types(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="'aaa'", python='aaa') \
+        .add_field(sql="U&'Hello winter \2603 !'", python='Hello winter °3 !') \
+        .add_field(sql="cast(null AS VARCHAR)", python=None) \
+        .add_field(sql="cast('bbb' AS VARCHAR(1))", python='b') \
+        .add_field(sql="cast(null AS VARCHAR(1))", python=None) \
+        .add_field(sql="cast('ccc' AS CHAR)", python='c') \
+        .add_field(sql="cast(null AS CHAR)", python=None) \
```
hashhar, line 105: split chars into their own method. The behave differently from varchar regarding padding.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+        .add_field(sql="cast('ddd' AS CHAR(1))", python='d') \
+        .add_field(sql="cast('😂' AS CHAR(1))", python='😂') \
+        .add_field(sql="cast(null AS CHAR(1))", python=None) \
+        .add_field(sql="X'65683F'", python='ZWg/') \
+        .add_field(sql="cast(null AS VARBINARY)", python=None) \
+        .add_field(sql="cast('{}' AS JSON)", python='"{}"') \
+        .add_field(sql="cast('null' AS JSON)", python='"null"') \
+        .add_field(sql="cast(null AS JSON)", python=None) \
```
hashhar, line 113: varbinary and JSON into their own methods.

Also let's test some varbinary values which don't represent valid strings.

hashhar: Also empty varbinary (not null) is special as well.

hashhar: for JSON we should also test float values, strings, strings with special chars (", ', :), empty array as well.

tests/integration/test_types_integration.py
hashhar: Add that as a test which asserts the failure - it documents the limitations and also makes it easy to notice when they are no longer limitations.

hashhar (commented):
Some more comments.

I'd recommend focusing on everything except temporal types for now since temporal types are a rabbit hole and need to be reviewed in more depth.

tests/integration/test_types_integration.py
hashhar: It's useful to mark the comments (at-least ones which won't require discussion) as resolved by yourself. Makes reviewing large PRs with lots of comments easier.

tests/integration/test_types_integration.py
lpoulain: Note that this is a Trino error, not a Python error. `SELECT cast('10.3' AS DECIMAL(38,38))` blows up in Starburst.

hashhar: It's because the value is not 38,38. 38,38 means only 38 decimal places, so values < 0. Try to test with `CAST('0.3123' AS DECIMAL(38,38))`

tests/integration/test_types_integration.py
lpoulain: Actually right now we don't seem to perform anything special treatment for JSON

tests/integration/test_types_integration.py
lpoulain: The comments actually don't mix well with the `\`

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+
+def test_boolean_types(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="false", python=False) \
+        .add_field(sql="true", python=True) \
+        .add_field(sql="cast(null AS BOOLEAN)", python=None) \
+        .execute()
+
+
+def test_tinyint_types(trino_connection):
```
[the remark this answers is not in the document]
lpoulain: Done

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,278 @@
...
+    SqlTest(trino_connection) \
+        .add_field(sql="cast(127 AS TINYINT)", python=127) \
```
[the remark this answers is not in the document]
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done for splitting each type in its own test

tests/integration/test_types_integration.py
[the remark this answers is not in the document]
hashhar: Since other files already follow the `_integration` suffix I'm leaving this as is.

tests/integration/test_types_integration.py
```diff
@@ -0,0 +1,294 @@
...
+        .add_field(sql="CAST(null AS CHAR)", python=None) \
+        .add_field(sql="CAST('ddd' AS CHAR(1))", python='d') \
+        .add_field(sql="CAST('😂' AS CHAR(1))", python='😂') \
+        .add_field(sql="CAST(null AS CHAR(1))", python=None) \
+        .execute()
+
+
+def test_varbinary(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="X'65683F'", python='ZWg/') \
```
hashhar, line 119: also test some varbinary input which isn't a valid UTF-8 string.

tests/integration/test_types_integration.py
hashhar: Either add TODO to add the missing tests or add the tests for the other values and assert that they fail.

tests/integration/test_types_integration.py
hashhar: Let's remove temporal type tests from this PR and tackle it in separate PR.

trino/client.py
```diff
@@ -839,3 +768,147 @@ def decorated(*args, **kwargs):
...
+
+    def create(self, columns, experimental_python_types):
+        assert columns is not None
+
+        if experimental_python_types:
+            return RowMapper([self._col_func(column['typeSignature']) for column in columns])
+        return RowMapper()
+
+    def _col_func(self, column):
+        col_type = column['rawType']
```
hashhar, line 788: What other fields does the typeSignature have available? I'm asking because it's possible for startswith to fail in presence of user contributed types.

trino/client.py
```diff
@@ -839,3 +768,147 @@ def decorated(*args, **kwargs):
...
+        elif col_type.startswith('decimal'):
+            return lambda val: Decimal(val)
+        elif col_type.startswith('double') or col_type.startswith('real'):
+            return self._double_map_func()
+        elif col_type.startswith('timestamp'):
+            return self._timestamp_map_func(column, col_type)
+        elif col_type.startswith('time'):
+            return self._time_map_func(column, col_type)
+        elif col_type == 'date':
+            return lambda val: datetime.strptime(val, '%Y-%m-%d').date()
```
hashhar, line 805: Seems this won't be able to handle negative years.

trino/client.py
```diff
@@ -839,3 +768,147 @@ def decorated(*args, **kwargs):
...
+        else:
+            return lambda val: val
+
+    def _array_map_func(self, column):
+        elt_mapping_func = self._col_func(column['arguments'][0]['value'])
```
hashhar, line 810: `element_mapping_func` - avoid abbreviations where possible.

Also seems `_map_<type>` would be shorter and still as clear  (`_array_map_func` -> `_map_array`).

trino/client.py
```diff
@@ -839,3 +768,147 @@ def decorated(*args, **kwargs):
...
+            else float(val)
+
+    def _timestamp_map_func(self, column, col_type):
+        pattern = "%Y-%m-%d %H:%M:%S"
+        ms_size, ms_to_trim = self._get_number_of_digits(column)
+        if ms_size > 0:
+            pattern += ".%f"
+
+        dt_size = 20 + ms_size - ms_to_trim
+        dt_tz_offset = 20 + ms_size
```
hashhar, line 835: 20 is a magic number. Extract a constant.

trino/client.py
```diff
@@ -839,3 +768,147 @@ def decorated(*args, **kwargs):
...
+        if 'with time zone' in col_type:
+
+            if ms_to_trim > 0:
+                return lambda val: \
+                    [datetime.strptime(val[:dt_size] + val[dt_tz_offset:], pattern + ' %z')
+                        if tz.startswith('+') or tz.startswith('-')
+                        else datetime.strptime(dt[:dt_size] + dt[dt_tz_offset:], pattern)
+                                     .replace(tzinfo=pytz.timezone(tz))
+                        for dt, tz in [val.rsplit(' ', 1)]][0]
```
hashhar, line 844: while this is very concise a proper function would be more readable.

trino/client.py
```diff
@@ -839,3 +768,147 @@ def decorated(*args, **kwargs):
...
+    lambda functions (one for each column) which will process a data value
+    and returns a RowMapper instance which will process rows of data
+    """
+
+    def create(self, columns, experimental_python_types):
+        assert columns is not None
+
+        if experimental_python_types:
+            return RowMapper([self._col_func(column['typeSignature']) for column in columns])
+        return RowMapper()
```
hashhar, line 785: Create a well-named singleton for the `NoOpRowMapper`/`PassThroughRowMapper`.

trino/client.py
lpoulain: Python does not seem to handle such years.

trino/client.py
lpoulain: typeSignature has `rawType` and `arguments` fields (the latter is used for types such as `decimal(3, 1)`)

tests/integration/test_types_integration.py
lpoulain: You mentioned at one point is that the order of the tests should be null / min value / intermediate value / max value. So should we be testing the min value or not?

trino/client.py
hashhar: Then can the checks below use `rawType == 'decimal'` instead of existing `rawType.startswith('decimal')` etc.?

tests/integration/test_types_integration.py
hashhar: We should test min value. What the later comment was referencing is https://github.com/trinodb/trino-python-client/pull/206/commits/eba160bdf600d96d12d72c41cd77fdb9c6fe52fb#diff-da781ffd6f3bd2fa7b1f7d7055dac6ce9f04b9a776feb777e35247eb024719a7L30-L31 (which you've fixed now).

tests/integration/test_types_integration.py
lpoulain: Actually it doesn't seem to work. SELECT CAST('😂😂😂😂😂😂' AS VARBINARY) returns `8J+YgvCfmILwn5iC8J+YgvCfmILwn5iC` in Starburst itself.

trino/client.py
lpoulain: I've seen cases where the value is `decimal(18, 18)`

trino/client.py
lpoulain: The mapping function is technically not mapping the array but the elements inside the array. Hence the name. But I can rename it.

trino/client.py
hashhar: 👀 Thanks. I'll need to re-read some code on the server side to have better understanding. 🙂

tests/integration/test_types_integration.py
hashhar: It's expected. Varbinary returns raw bytes. However if you run from_utf8 on that varbinary output then it should give you back the string.

```
trino> SELECT from_utf8(CAST('😂😂😂😂😂😂' AS VARBINARY));
    _col0
--------------
 😂😂😂😂😂😂
(1 row)
```

hashhar: A test like `sql="X'0001020304050607080DF9367AA7000000'", python='<whatever bag of bytes we get back>'` should be fine.

tests/integration/test_types_integration.py
hashhar: Very unfortunate. 😢 Let's live with it.

hashhar: "Unresolved" for the second question. I think an example is `DATE` where Trino has wider range than Python?

trino/client.py
lpoulain: Would it? The function relies on multiple variables computed before the lambda (e.g. `dt_tz_offset`, `tz`, etc.)

tests/integration/test_types_integration.py
lpoulain: Yes, `DATE` indeed. But didn't you say to drop the temporal test types for now?

hashhar: Yes. Let's drop dates. But are there other types with such a case? DATE was the only example I know of.

tests/integration/test_types_integration.py
lpoulain: What test are you talking about here? JSON tests (even though we just convert them to strings), varbinary which represent invalid strings (e.g. non-UTF-8?) or something else?

trino/client.py

trino/client.py

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: It's my impression that Python barely has any limit when it comes to any type of number.

tests/integration/test_types_integration.py
hashhar: Both.

- Add test for varbinary which doesn't represent a string
- Add test for JSON with:
  - JSON '3.14'
  - JSON 'a string'
  - JSON 'a " complex \' string :' (I'm not sure if I escaped that `'` correctly)
  - JSON '[]'

Even if the test fails we should assert on that failure  so that when something like https://github.com/trinodb/trino-python-client/pull/214 changes behaviour we know to update the tests. It also serves as documentation of the "issues" we need to fix.

trino/client.py
```diff
@@ -770,19 +770,29 @@ def decorated(*args, **kwargs):
     return wrapper
 
 
+class NoOpRowMapper:
+    """
+    No-op RowMapper which does not perform any transformation
```
hashhar, line 775: nit: Might be useful to add - "Used when experimental_python_types is False." to the docstring.

trino/client.py
```diff
@@ -826,13 +836,14 @@ def _double_map_func(self):
             else float(val)
 
     def _timestamp_map_func(self, column, col_type):
+        datetime_default_size = 20
```
hashhar, line 839: Why 20. I see:

```python
>>> len(datetime.datetime.strftime(datetime.datetime.utcnow(), "%Y-%m-%d %H:%M:%S"))
19
```

I'm probably missing something or this deserves a comment.

tests/integration/test_types_integration.py
hashhar: I was thinking of the other direction. i.e. types in Trino which have wider range that Python. Seems there are none. We can revisit when we find any. Resolved.

tests/integration/test_types_integration.py
hashhar: Temporal type tests will be a follow-up PR, so marking as resolved.

They need more thorough and dedicated review.

tests/integration/test_types_integration.py
```diff
@@ -117,7 +116,9 @@ def test_char(trino_connection):
 def test_varbinary(trino_connection):
     SqlTest(trino_connection) \
         .add_field(sql="X'65683F'", python='ZWg/') \
+        .add_field(sql="X'0001020304050607080DF9367AA7000000'", python='AAECAwQFBgcIDfk2eqcAAAA=') \
         .add_field(sql="CAST('' AS VARBINARY)", python='') \
```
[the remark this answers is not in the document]
lpoulain: Done

trino/client.py
lpoulain: Done

trino/client.py

tests/integration/test_types_integration.py
lpoulain: I'm not entirely sure I know exactly what you have in mind but I've added such tests.

hashhar (approved):
LGTM.

@ebyhr FYI I've asked timestamp related tests to be added in separate PR to help with reviewing them.

hashhar:
@lpoulain Can you squash the "fixups" together?

lpoulain:
@hashhar done

hashhar:
@lpoulain "All commits" do not need to be squashed. Just the logically related ones.

For future PRs see https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#git-merge-strategy

hashhar:
@lpoulain Please reword the commit message as:

```Optimize experimental_python_types and add type-mapping tests

Instead of checking the type for each row, check the type once for each
fetch() call and compute a list of lambdas which are to be applied to
the values from each row. A new RowMapperFactory class is created to
wrap this behavior.
The experimental_python_types flag is now processed in the TrinoQuery
class instead of the TrinoResult class.

Type mapping tests for each lambda which maps rows to Python types is
added.
```

hashhar (approved):
Looks good but please reword commit message to have useful information.

lpoulain:
> Looks good but please reword commit message to have useful information.

@hashhar done

PR #208, opened 2022-07-27
[mypy] Enforce Mypy checks for a subset of modules

john-bodley:
Even though Mypy is enabled, per [this](https://github.com/trinodb/trino-python-client/blob/master/.pre-commit-config.yaml#L8-L15) pre-commit hook, the default options are fairly loose which allows for untyped methods and calls.

This PR aims to slowly start tightening the screws module by module (resulting in digestible PRs) —starting with `trino.exceptions`, `trino.logging`, and `trino.transactions`—by enforcing typed methods, calls, etc. which improves code clarity and correctness.

The hope is over time contributors can systematically go through module by module and ensure all methods are correctly typed and remove them from the ignore list.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,8 +4,20 @@ test=pytest
 [flake8]
 max-line-length = 120
 # W503 raises a warning when there is a line break before a binary operator.
-# This is best practice according to PEP 8 and the rule should be ignored. 
+# This is best practice according to PEP 8 and the rule should be ignored.
 #
 # https://www.flake8rules.com/rules/W503.html
 # https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
 ignore = W503
+
+[mypy]
+check_untyped_defs = true
+disallow_any_generics = true
+disallow_untyped_calls = true
+disallow_untyped_defs = true
+ignore_missing_imports = true
+no_implicit_optional = true
+warn_unused_ignores = true
+
+[mypy-tests.*,trino.auth,trino.client,trino.dbapi,trino.sqlalchemy.*]
+ignore_errors = true
--- a/trino/client.py
+++ b/trino/client.py
@@ -492,7 +492,7 @@ class TrinoRequest(object):
     def next_uri(self) -> Optional[str]:
         return self._next_uri
 
-    def post(self, sql, additional_http_headers=None):
+    def post(self, sql: str, additional_http_headers: Optional[Dict[str, Any]] = None):
         data = sql.encode("utf-8")
         # Deep copy of the http_headers dict since they may be modified for this
         # request by the provided additional_http_headers
@@ -524,7 +524,7 @@ class TrinoRequest(object):
                 )
         return http_response
 
-    def get(self, url):
+    def get(self, url: str):
         return self._get(
             url,
             headers=self.http_headers,
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,8 +11,7 @@ repos:
       - id: "mypy"
         name: "Python: types"
         additional_dependencies:
-          - "types-pytz"
-          - "types-requests"
+          - "types-all"
 
   - repo: https://github.com/pycqa/isort
     rev: 5.6.4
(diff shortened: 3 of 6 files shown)


setup.cfg
```diff
@@ -4,8 +4,20 @@ test=pytest
...
 max-line-length = 120
 # W503 raises a warning when there is a line break before a binary operator.
-# This is best practice according to PEP 8 and the rule should be ignored. 
+# This is best practice according to PEP 8 and the rule should be ignored.
 #
 # https://www.flake8rules.com/rules/W503.html
 # https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
 ignore = W503
+
+[mypy]
```
john-bodley, line 13: These are a fairly common set of flags to enable, i.e., they have been hugely valuable for projects like Apache Superset as configured [here](https://github.com/apache/superset/blob/f37fc1a7f0297ff3348221ed784a1034d2cf44e9/setup.cfg#L37-L44).

setup.cfg
```diff
@@ -4,8 +4,20 @@ test=pytest
...
+check_untyped_defs = true
+disallow_any_generics = true
+disallow_untyped_calls = true
+disallow_untyped_defs = true
+ignore_missing_imports = true
+no_implicit_optional = true
+warn_unused_ignores = true
+
+[mypy-tests.*,trino.auth,trino.client,trino.dbapi,trino.sqlalchemy.*]
+ignore_errors = true
```
hovaesco, line 23: Why mypy errors are ignored it these packages? For instance I see that you addressed some in `trino.client.py`, is it leftover?

trino/client.py
```diff
@@ -456,7 +456,7 @@ def statement_url(self) -> str:
     def next_uri(self) -> Optional[str]:
         return self._next_uri
 
-    def post(self, sql, additional_http_headers=None):
+    def post(self, sql: str, additional_http_headers: Optional[Dict[str, Any]] = None):
```
john-bodley, line 459: As of yet I'm unsure of the return type. This should become clearer when the module has type enforcement. This method (as well as `get`) are required to have their arguments typed given they're called in `trino/transaction.py` and we have `disallow_untyped_calls = true` set.

setup.cfg
john-bodley: @hovaesco I added [this](https://github.com/trinodb/trino-python-client/pull/208/files#r931114490) comment which hopefully explains why the change in `trino/client.py` is required.

mdesmet (approved):
We should add to the developers setup section in the README how to automatically execute the pre-commit hook when committing as in https://pre-commit.com/#usage

setup.cfg
hashhar: What is `mypy-tests.*` module? Some leftover?

hashhar:
@mdesmet @hovaesco I'm merging this. Can you send a follow-up PR with updated instructions in the README.md?

setup.cfg
john-bodley: @hashhar `mypy-tests` is not a module, see [here](https://mypy.readthedocs.io/en/stable/config_file.html#config-file-format) for specifics, but per,

> Additional sections named [mypy-PATTERN1,PATTERN2,...] may be present, where PATTERN1, PATTERN2, etc., are comma-separated patterns of fully-qualified module names, with some components optionally replaced by the ‘*’ character (e.g. foo.bar, foo.bar.*, foo.*.baz). These sections specify additional flags that only apply to modules whose name matches at least one of the patterns.

the TL;DR is for the `tests.*`,  `trino.auth`, etc. modules errors are currently being ignored.

mdesmet:
@hashhar : Let's merge this. I will add another PR with updated instructions.

ebyhr:
@john-bodley Could you rebase on upstream to resolve conflicts?

hashhar:
@john-bodley I've rebased to resolve conflicts - no other changes. Once CI finishes and is green we'll merge this. Thanks for doing this work and sorry it took so long to get it done.

hovaesco (approved):

hashhar (approved):

PR #212, opened 2022-08-02
Support setting roles in dbapi and sqlalchemy

### LINKED ISSUE #230: It's not possible to set role in trino.dbapi.connect method
### Describe the feature

With the recent changes with https://github.com/trinodb/trino-python-client/pull/200 `X-Trino-Role` and `X-Trino-Set-Role` are not allowed to be set. The following code:

```python
connection = trino.dbapi.connect(
    host="trino.example.net",
    port=443,
    http_scheme="https",
    catalog="hive",
    user='abuzer',
    auth=trino.auth.JWTAuthentication('Xvfdas.....'),
    http_headers={'X-Trino-Role': 'hive=ROLE{admin}'}
)
```
gives an error `ValueError: cannot override reserved HTTP header X-Trino-Role`.

We should make the `role` parameter part of `trino.dbapi.connect` method.

I can to set `role` in `connection._client_session.role` or `connection._http_session.headers`. Since `_client_session` property should be an internal abstraction, `connection._client_session.role='hive=ROLE{admin}'` is ugly and prone to change. We need a proper interface to set such values in `connect` method.

We should keep release notes and publish critical breaking changes in future.

mdesmet:
fixes #230

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,20 @@ cur.execute('SELECT * FROM system.runtime.nodes')
 rows = cur.fetchall()
 ```
 
+## Roles
+
+Authorization roles to use for catalogs, specified as a dict with key-value pairs for the catalog and role. For example, `{"catalog1": "roleA", "catalog2": "roleB"}` sets `roleA` for `catalog1` and `roleB` for `catalog2`. See Trino docs.
+
+```python
+import trino
+conn = trino.dbapi.connect(
+    host='localhost',
+    port=443,
+    user='the-user',
+    roles={"catalog1": "roleA", "catalog2": "roleB"},
+)
+```
+
 ## SSL
 
 ### SSL verification
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ engine = create_engine(
       "session_properties": {'query_max_run_time': '1d'},
       "client_tags": ["tag1", "tag2"],
       "experimental_python_types": True,
+      "roles": {"catalog1": "role1"},
     }
 )
 
@@ -115,7 +116,8 @@ engine = create_engine(
     'trino://user@localhost:8080/system?'
     'session_properties={"query_max_run_time": "1d"}'
     '&client_tags=["tag1", "tag2"]'
-    '&experimental_python_types=true',
+    '&experimental_python_types=true'
+    '&roles={"catalog1": "role1"}'
 )
 ```
 
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -1070,6 +1070,19 @@ def test_set_role_trino_351(run_trino):
     assert_role_headers(cur, "tpch=ALL")
 
 
+@pytest.mark.skipif(trino_version() == '351', reason="Newer Trino versions return the system role")
+def test_set_role_in_connection_trino_higher_351(run_trino):
+    _, host, port = run_trino
+
+    trino_connection = trino.dbapi.Connection(
+        host=host, port=port, user="test", catalog="tpch", roles={"system": "ALL"}
+    )
+    cur = trino_connection.cursor()
+    cur.execute('SHOW TABLES FROM information_schema')
+    cur.fetchall()
+    assert_role_headers(cur, "system=ALL")
+
+
 def assert_role_headers(cursor, expected_header):
     assert cursor._request.http_headers[constants.HEADER_ROLE] == expected_header
 
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -19,6 +19,7 @@ import requests
 
 import trino
 from tests.integration.conftest import trino_version
+from trino import constants
 from trino.exceptions import TrinoQueryError, TrinoUserError, NotSupportedError
 from trino.transaction import IsolationLevel
 
@@ -1045,11 +1046,11 @@ def test_set_role_trino_higher_351(run_trino):
     cur = trino_connection.cursor()
     cur.execute('SHOW TABLES FROM information_schema')
     cur.fetchall()
-    assert cur._request._client_session.role is None
+    assert cur._request._client_session.roles == {}
 
     cur.execute("SET ROLE ALL")
     cur.fetchall()
-    assert cur._request._client_session.role == "system=ALL"
+    assert_role_headers(cur, "system=ALL")
 
 
 @pytest.mark.skipif(trino_version() != '351', reason="Trino 351 returns the role for the current catalog")
@@ -1062,11 +1063,15 @@ def test_set_role_trino_351(run_trino):
     cur = trino_connection.cursor()
     cur.execute('SHOW TABLES FROM information_schema')
     cur.fetchall()
-    assert cur._request._client_session.role is None
+    assert cur._request._client_session.roles == {}
 
     cur.execute("SET ROLE ALL")
     cur.fetchall()
-    assert cur._request._client_session.role == "tpch=ALL"
+    assert_role_headers(cur, "tpch=ALL")
+
+
+def assert_role_headers(cursor, expected_header):
+    assert cursor._request.http_headers[constants.HEADER_ROLE] == expected_header
 
 
 def test_prepared_statements(run_trino):
--- a/tests/unit/test_dbapi.py
+++ b/tests/unit/test_dbapi.py
@@ -254,3 +254,13 @@ def test_tags_are_set_when_specified(mock_client):
 
     _, passed_client_tags = mock_client.ClientSession.call_args
     assert passed_client_tags["client_tags"] == client_tags
+
+
+@patch("trino.dbapi.trino.client")
+def test_role_is_set_when_specified(mock_client):
+    roles = {"system": "finance"}
+    with connect("sample_trino_cluster:443", roles=roles) as conn:
+        conn.cursor().execute("SOME FAKE QUERY")
+
+    _, passed_role = mock_client.ClientSession.call_args
+    assert passed_role["roles"] == roles
--- a/tests/unit/test_dbapi.py
+++ b/tests/unit/test_dbapi.py
@@ -248,11 +248,9 @@ def test_token_retrieved_once_when_multithreaded(sample_post_response_data, samp
 
 @patch("trino.dbapi.trino.client")
 def test_tags_are_set_when_specified(mock_client):
-    # WHEN
     client_tags = ["TAG1", "TAG2"]
     with connect("sample_trino_cluster:443", client_tags=client_tags) as conn:
         conn.cursor().execute("SOME FAKE QUERY")
 
-    # THEN
     _, passed_client_tags = mock_client.ClientSession.call_args
     assert passed_client_tags["client_tags"] == client_tags
--- a/trino/client.py
+++ b/trino/client.py
@@ -224,7 +224,7 @@ def get_header_values(headers, header):
 def get_session_property_values(headers, header):
     kvs = get_header_values(headers, header)
     return [
-        (k.strip(), urllib.parse.unquote(v.strip()))
+        (k.strip(), urllib.parse.unquote_plus(v.strip()))
         for k, v in (kv.split("=", 1) for kv in kvs)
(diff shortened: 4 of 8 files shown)


tests/integration/test_dbapi_integration.py
```diff
@@ -1063,8 +1063,21 @@ def test_set_role_trino_351(run_trino):
...
+def test_set_role_in_connection_trino_higher_351(run_trino):
+    _, host, port = run_trino
+
+    trino_connection = trino.dbapi.Connection(
+        host=host, port=port, user="test", catalog="tpch", roles={"system": "ALL"}
+    )
+    cur = trino_connection.cursor()
+    cur.execute('SHOW TABLES FROM information_schema')
+    cur.fetchall()
+    assert cur._request._client_session.roles == "system=ALL"
```
hovaesco, line 1083: Please add assertion which validates headers being sent.

trino/client.py
```diff
@@ -197,14 +197,19 @@ def __setstate__(self, state):
...
-    def role(self, role):
+    @roles.setter
+    def roles(self, roles):
         with self._object_lock:
-            self._role = role
+            self._roles = roles
+
+    def _create_roles(self, roles: Optional[Dict[str, str]]) -> Optional[str]:
+        if roles is not None:
+            return ";".join(map(lambda role: f"{role[0]}={role[1]}", roles.items()))
```
hovaesco, line 211: I'm not sure whether other clients do that maybe it's worth to catch incorrect roles value and put an error message about a proper roles syntax.

mdesmet: Our param is a `Dict[str, str]`, we don't really check if the passed string is properly escaped if you supply a role containing chars like `=` or `;`. I would think it needs to follow quoted identifiers rules but not sure. I will check the JDBC client and see if I can find something.

mdesmet: It seems to not work as expected: See https://github.com/trinodb/trino/issues/14252

mdesmet: I don't think it should be handled by us, but be based on user input. I think the user should just pass in quoted input and it's up to the engine to correctly handle quoted identifiers:

Something like `{"hive":'"my=weird;role_name="'}`

hovaesco:
Please rebase on master to resolve CI failures.

hovaesco (approved):

mdesmet:
@hashar: This would enable us to support grants in dbt-trino.

trino/client.py
hashhar: I agree, quoting should be left to user to perform.
We don't do it for them in queries or catalog/schema names - we shouldn't do it other places either.

trino/client.py
```diff
@@ -197,14 +197,19 @@ def __setstate__(self, state):
...
+        return ""
```
hashhar, line 212: why not return `None`? Is the header required even if no role is set?

README.md
```diff
@@ -321,6 +321,20 @@ cur.execute('SELECT * FROM system.runtime.nodes')
 rows = cur.fetchall()
 ```
 
+## Roles
+
+Authorization roles to use for catalogs, specified as a dict with key-value pairs for the catalog and role. For example, `{"catalog1": "roleA", "catalog2": "roleB"}` sets roleA for catalog1 and roleB for catalog2.
```
hashhar, line 328:
```suggestion
Authorization roles to use for catalogs, specified as a dict with key-value pairs for the catalog and role. For example, `{"catalog1": "roleA", "catalog2": "roleB"}` sets `roleA` for `catalog1` and `roleB` for `catalog2`.
```

Also might be useful to link to Trino docs about roles since AFAIK there is also something called a system role.

tests/unit/test_dbapi.py
```diff
@@ -256,3 +256,15 @@ def test_tags_are_set_when_specified(mock_client):
     # THEN
     _, passed_client_tags = mock_client.ClientSession.call_args
     assert passed_client_tags["client_tags"] == client_tags
+
+
+@patch("trino.dbapi.trino.client")
+def test_role_is_set_when_specified(mock_client):
+    # WHEN
```
hashhar, line 263: seems redundant comment

hashhar: Can we inspect current role for user from the server using a query?

If so I'd prefer if the test used it instead of just asserting arguments are passed correctly.

README.md
```diff
@@ -108,7 +109,8 @@ engine = create_engine(
     'trino://user@localhost:8080/system?'
     'session_properties={"query_max_run_time": "1d"}'
     '&client_tags=["tag1", "tag2"]'
-    '&experimental_python_types=true',
+    '&experimental_python_types=true'
+    '&roles={"catalog1": "role1"}'
```
hashhar, line 113: no test for sqlalchemy integration?

README.md
mdesmet: There seems no specific page about roles, only about the SQL syntax. I think a proper page about how roles work in Trino will be definitely beneficial.

tests/unit/test_dbapi.py
mdesmet: We could test it with Hive but that means we would have to change our docker setup to use docker compose.

```
❯ trino --user admin
trino> show current roles in hive;
  Role  
--------
 public 
(1 row)

Query 20220923_160052_00000_kq3fp, FINISHED, 1 node
Splits: 1 total, 1 done (100,00%)
1,51 [0 rows, 0B] [0 rows/s, 0B/s]

trino> show current roles;
 Role 
------
(0 rows)

Query 20220923_160107_00001_kq3fp, FINISHED, 1 node
Splits: 1 total, 1 done (100,00%)
0,05 [0 rows, 0B] [0 rows/s, 0B/s]

trino> create role public;
Query 20220923_160134_00002_kq3fp failed: System roles are not enabled

```

README.md
mdesmet: See https://github.com/trinodb/trino-python-client/pull/212#discussion_r978841837

mdesmet:
@hashhar: Please advise on how to get this merged. Note that we had roles already supported (within the session) and they were tested in the same way (now expanded up to http_headers) but not in integration tests as we don't have a connector that supports roles in this project.

As mentioned we need this feature in as we want to support roles security within dbt_trino.

README.md
hashhar: cc: @mosabua @dain regarding some discussion of what roles mean - we can just explain the concept because AFAIK there is no plugin providing roles in Trino - just the SPI.

tests/unit/test_dbapi.py
hashhar: Can you create an issue about integration tests for roles and we can discuss how to add them as follow up (since once we have them we'd probably want to use it for other tests too).

hashhar:
I'm fine with current tests for now but as we start to work more on this area we'd eventually need better integration tests - let's create an issue with current ideas and we can follow up their.

@ebyhr WDYT?

hashhar:
@mdesmet Can you also update PR description with the template + propose a release note?

tests/integration/test_dbapi_integration.py
```diff
@@ -19,6 +19,7 @@
 
 import trino
 from tests.integration.conftest import trino_version
+from trino import constants
```
hashhar, line 22: nit: `Refactor role into roles` -> `Add support for system and connector roles`?

Does it make sense (and is it accurate?)?

hashhar (approved):
LGTM % comments.

tests/integration/test_dbapi_integration.py
mdesmet: It is really just a rename of role into roles. There is no new feature being introduced in that commit actually. The `role` field actually already contained multiple roles. I thought this is more consistent with the roles being passed on from `dbapi.connect`, introduced in the next commit.

mdesmet:
> I'm fine with current tests for now but as we start to work more on this area we'd eventually need better integration tests - let's create an issue with current ideas and we can follow up their.

Created #241 241

tests/unit/test_dbapi.py
mdesmet: Done

hashhar (dismissed):

tests/integration/test_dbapi_integration.py
hashhar: That's the reason why the first commit by itself fails tests. See 

```
======================================================================================================================================= FAILURES ========================================================================================================================================
_____________________________________________________________________________________________________________________ test_set_role_in_connection_trino_higher_351 ______________________________________________________________________________________________________________________

run_trino = (<Popen: returncode: None args: ['docker', 'run', '--rm', '-p', '56257:8080',...>, 'localhost', 56257)

    @pytest.mark.skipif(trino_version() == '351', reason="Newer Trino versions return the system role")
    def test_set_role_in_connection_trino_higher_351(run_trino):
        _, host, port = run_trino

>       trino_connection = trino.dbapi.Connection(
            host=host, port=port, user="test", catalog="tpch", roles={"system": "ALL"}
        )
E       TypeError: Connection.__init__() got an unexpected keyword argument 'roles'

tests/integration/test_dbapi_integration.py:1077: TypeError
------------------------------------------------------------------------------------------------------------------------------- Captured stdout teardown --------------------------------------------------------------------------------------------------------------------------------
trino-python-client-tests-637e3b1
____________________________________________________________________________________________________________________________ test_role_is_set_when_specified ____________________________________________________________________________________________________________________________

mock_client = <MagicMock name='client' id='4532039360'>

    @patch("trino.dbapi.trino.client")
    def test_role_is_set_when_specified(mock_client):
        roles = {"system": "finance"}
>       with connect("sample_trino_cluster:443", roles=roles) as conn:

tests/unit/test_dbapi.py:262:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = ('sample_trino_cluster:443',), kwargs = {'roles': {'system': 'finance'}}

    def connect(*args, **kwargs):
        """Constructor for creating a connection to the database.

        See class :py:class:`Connection` for arguments.

        :returns: a :py:class:`Connection` object.
        """
>       return Connection(*args, **kwargs)
E       TypeError: Connection.__init__() got an unexpected keyword argument 'roles'

trino/dbapi.py:82: TypeError
================================================================================================================================ short test summary info ================================================================================================================================
FAILED tests/integration/test_dbapi_integration.py::test_set_role_in_connection_trino_higher_351 - TypeError: Connection.__init__() got an unexpected keyword argument 'roles'
FAILED tests/unit/test_dbapi.py::test_role_is_set_when_specified - TypeError: Connection.__init__() got an unexpected keyword argument 'roles'
```

hashhar:
@mdesmet Can you rebase, I see some failures when running tests locally. Possibly something changed on master in the meantime.

tests/integration/test_dbapi_integration.py
mdesmet: I probably messed up something in my fixup comments locally. All good now.

mdesmet:
@hashhar : Unfortunately something fishy is going on: 351 and 395 pass. However 398 does fail consistently (also locally), while 397 works fine.

I checked the release notes but didn't immediately find something. I'll keep digging.

hashhar:
See https://github.com/trinodb/trino/pull/13907/files#r984659618, a "bug" has been fixed - we now don't get result sets for non-SELECT queries like it should've been all along.

mdesmet:
@hashhar: I've realigned the implementation on the session_properties and your changes in #242 

It would be nice to get this one in as well.

hashhar (approved):

Issue #194, opened 2022-07-06
Add JSON functionality to SQLAlchemy Dialect


laserkaplan:
In messing around with getting the Trino SQLAlchemy Dialect to work with JSON columns/data, I came across at least a couple of necessary additions:

- a `visit_JSON` method is needed in the `TrinoTypeCompiler` class since SQLAlchemy, by default, doesn't support a `JSON` type (as it is dialect-specific).
- the `TrinoDialect` class needs both a `_json_serializer` and `_json_deserializer` member, which I believe can simply be `json.dumps` and `json.loads`, respectively.

Adding those two items allowed me to create tables with a JSON type and write to them all using SQLAlchemy. However, when attempting to read back from the table, I got the following error:

```
Traceback (most recent call last):
  File "/Users/laserkaplan/PycharmProjects/proxima/proxima_warehouse/pipelines/process_shopify_orders/process.py", line 95, in <module>
    print(session.execute(select(sis_table).limit(10)).all())
  File "/Users/laserkaplan/.pyenv/versions/proxima-3.9.7/lib/python3.9/site-packages/sqlalchemy/engine/result.py", line 1068, in all
    return self._allrows()
  File "/Users/laserkaplan/.pyenv/versions/proxima-3.9.7/lib/python3.9/site-packages/sqlalchemy/engine/result.py", line 403, in _allrows
    made_rows = [make_row(row) for row in rows]
  File "/Users/laserkaplan/.pyenv/versions/proxima-3.9.7/lib/python3.9/site-packages/sqlalchemy/engine/result.py", line 403, in <listcomp>
    made_rows = [make_row(row) for row in rows]
  File "/Users/laserkaplan/.pyenv/versions/proxima-3.9.7/lib/python3.9/site-packages/sqlalchemy/sql/sqltypes.py", line 2693, in process
    return json_deserializer(value)
TypeError: loads() takes 1 positional argument but 2 were given
```

So something is off with the deserializer, I just haven't had the chance to dig into it further yet.

Does anyone have any suggestions of how to proceed here? Or, if you have more SQLAlchemy JSON experience than I do (which isn't hard to do :) ), do you know if anything else is missing?

mdesmet:
Hi @laserkaplan, can you make a PR? This way It will be much easier to have a look at your code and get it working.

PR #214, opened 2022-08-09
Add JSON functionality to SQLAlchemy Dialect

laserkaplan:
## Description

Implement the ability to reference JSON columns using SQLAlchemy. This allows JSON columns to be created in new tables, as well as writes to/reads from them using normal Python `json` module serialization/deserialization.

Originally referenced [in this issue](https://github.com/trinodb/trino-python-client/issues/194).

## Release notes

```markdown
* Add support for creating tables containing `JSON` columns and reading and writing to them with SQLAlchemy. ([#194](https://trinodb/trino-python-client/issues/194))
```

### DIFF
--- a/tests/integration/test_sqlalchemy_integration.py
+++ b/tests/integration/test_sqlalchemy_integration.py
@@ -13,6 +13,8 @@ import pytest
 import sqlalchemy as sqla
 from sqlalchemy.sql import and_, or_, not_
 
+from trino.sqlalchemy.datatype import JSON
+
 
 @pytest.fixture
 def trino_connection(run_trino, request):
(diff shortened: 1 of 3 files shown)


trino/sqlalchemy/dialect.py
```diff
@@ -64,6 +64,13 @@ class TrinoDialect(DefaultDialect):
     # Support proper ordering of CTEs in regard to an INSERT statement
     cte_follows_insert = True
 
+    # JSON serializer and deserializer
+    @staticmethod
+    def trino_json_serializer(text):
+        return TrinoJSONType(json.dumps(text))
+    _json_serializer = trino_json_serializer
+    _json_deserializer = None
```
hovaesco, line 72: Based on the PR description:
> using normal Python json module serialization/deserialization

deserialization should be also done using `json` module, why then here it's `None`?

laserkaplan: Yes, good eye :)
This is actually because of a minor bug in the SQLAlchemy code. See here: https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/sqltypes.py#L2627
If you don't specify a `_json_serializer` or `_json_deserializer` at all in the dialect, then this will throw an error trying to access an attribute that doesn't exist.
However, if you try to set this attribute to just `json.loads`, Python thinks it's an instance method, and expects 2 parameters to be passed (one of which is `self`), and thus causes the downstream call to fail.
I can get around this by setting `_json_deserializer = staticmethod(json.loads)` if you think this is more clear. I can also just add a comment to this effect.

trino/sqlalchemy/dialect.py
```diff
@@ -64,6 +64,13 @@ class TrinoDialect(DefaultDialect):
     # Support proper ordering of CTEs in regard to an INSERT statement
     cte_follows_insert = True
 
+    # JSON serializer and deserializer
```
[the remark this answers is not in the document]
laserkaplan: Thanks! Great suggestion, I will absolutely look into this and update the PR.
SQLAlchemy is such a massive package that it's easy to miss good functionality like this :)

trino/sqlalchemy/types.py
```diff
@@ -0,0 +1,27 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import json
+
+import sqlalchemy.types as types
+
+
+class JSON(types.TypeDecorator):
```
hovaesco, line 17: This should go to `trino/sqlalchemy/datatype.py`

mdesmet (approved):
@laserkaplan: please have a look at the pending remark of @hovaesco on moving the TypeDecorator to datatype.py.

LGTM % the above remark and squash the commits.

laserkaplan:
Whoops, thanks for the reminder! Moved to `datatype.py`.

laserkaplan:
Seems like these tests failing might be a larger issue? Multiple open PRs are all failing at the same spot.

mdesmet:
> Seems like these tests failing might be a larger issue? Multiple open PRs are all failing at the same spot.

Yes, it should be gone when #220 is merged.

hovaesco (approved):
Please squash commits.

hashhar (approved):
I'll merge this after I merge https://github.com/trinodb/trino-python-client/pull/220 tomorrow morning my time.

Thanks for contributing @laserkaplan and sorry about the very long delay.

PR #218, opened 2022-08-11
specify `catalog_name` in `get_table_comment`

### LINKED ISSUE #217: `get_table_comment` issues
There are 2 issue with [current implementation of `get_table_comment`](https://github.com/trinodb/trino-python-client/blob/e26027375326a1e0fc8a472d242af75b4e6049bc/trino/sqlalchemy/dialect.py#L268-L275) in SQLAlchemy dialect

1. The query select from `"system"."metadata"."table_comments"` but not specify `catalog_name` that can lead to wrong result if Trino has 2 tables with the same name & schema but in different catalogs.
image
2. The query will be hang if one of the connector is not connected. Demo: I shutdown PostgreSQL and get comment of `actor` table in MySQL catalog.
image

dungdm93:
Fix #217

### DIFF
--- a/trino/sqlalchemy/dialect.py
+++ b/trino/sqlalchemy/dialect.py
@@ -262,19 +262,26 @@ class TrinoDialect(DefaultDialect):
         return []
 
     def get_table_comment(self, connection: Connection, table_name: str, schema: str = None, **kw) -> Dict[str, Any]:
-        schema = schema or self._get_default_schema_name(connection)
-        if schema is None:
+        catalog_name = self._get_default_catalog_name(connection)
+        if catalog_name is None:
+            raise exc.NoSuchTableError("catalog is required in connection")
+        schema_name = schema or self._get_default_schema_name(connection)
+        if schema_name is None:
             raise exc.NoSuchTableError("schema is required")
         query = dedent(
             """
             SELECT "comment"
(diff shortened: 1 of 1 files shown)


hovaesco (approved):

trino/sqlalchemy/dialect.py
```diff
@@ -262,19 +262,26 @@ def get_check_constraints(
         return []
 
     def get_table_comment(self, connection: Connection, table_name: str, schema: str = None, **kw) -> Dict[str, Any]:
-        schema = schema or self._get_default_schema_name(connection)
-        if schema is None:
+        catalog_name = self._get_default_catalog_name(connection)
+        if catalog_name is None:
+            raise exc.NoSuchTableError("catalog is required in connection")
```
hashhar, line 267: Probably also update `raise exc.NoSuchTableError("schema is required")` -> `raise exc.NoSuchTableError("schema is required in connection")`?

hashhar (approved):
Thank you.

trino/sqlalchemy/dialect.py
dungdm93: @hashhar `schema` can be specified in either param or connection URL, while only specified in connection URL

PR #219, opened 2022-08-11
return empty list in un-partitioned table

dungdm93:
Currently, in un-partitioned table, trino driver still return a single-value list of a partition index with **empty columns**

image

So, it's better to **return empty list** instead.

### DIFF
--- a/trino/sqlalchemy/dialect.py
+++ b/trino/sqlalchemy/dialect.py
@@ -238,12 +238,14 @@ class TrinoDialect(DefaultDialect):
             raise exc.NoSuchTableError(f"schema={schema}, table={table_name}")
 
         partitioned_columns = self._get_columns(connection, f"{table_name}$partitions", schema, **kw)
+        if not partitioned_columns:
+            return []
         partition_index = dict(
-            name="partition", column_names=[col["name"] for col in partitioned_columns], unique=False
+            name="partition",
+            column_names=[col["name"] for col in partitioned_columns],
+            unique=False
         )
-        return [
-            partition_index,
-        ]
+        return [partition_index]
 
     def get_sequence_names(self, connection: Connection, schema: str = None, **kw) -> List[str]:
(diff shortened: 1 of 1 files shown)


trino/sqlalchemy/dialect.py
```diff
@@ -238,12 +238,14 @@ def get_indexes(self, connection: Connection, table_name: str, schema: str = Non
             raise exc.NoSuchTableError(f"schema={schema}, table={table_name}")
 
         partitioned_columns = self._get_columns(connection, f"{table_name}$partitions", schema, **kw)
```
ebyhr, line 240: `$partitions` is specific to some connectors and it's not ensured in the engine level. 

Can you explain how `get_indexes()` is used from caller side? What happens if we don't support this method?

dungdm93: Sure. `get_indexes()` is used to get indexes info. There are different kinds of indexes:
* Normal index
image
* FOREIGN KEY
image
* PARTITION in the case of Trino / Presto or Hive

This method could be used by clients like Superset to:
* display info to user (like pictures above)
* perform template query e.g: select data from latest partition.

From Trino driver side, `get_indexes` will call `_get_columns` of `$partitions`, which in turn perform query from `"information_schema"."columns"` table. So, if table is un-partitioned or catalog does not support partition, `_get_columns` just return empty list - no error/exception here.

ebyhr: Thanks for sharing the information!

>  So, if table is un-partitioned or catalog does not support partition, _get_columns just return empty list - no error/exception here.

What happens if the 3rd party connectors use `$partitions` table for different purpose and it returns different table definition? Is the current implementation really safe in such situation? 

I know this comment is unrelated to this change, so I'm going to merge this PR. However, please keep in mind that we shouldn't add such connector specific feature to this client.

ebyhr (approved):

PR #220, opened 2022-08-12
Acknowledge reception of data in `TrinoResult`

### LINKED ISSUE #232: `execute` does not kick off query on Trino Server
### Expected behavior

The call to `client.execute` does not start the query on the Trino cluster.

### Actual behavior

The query only starts on the Trino cluster once `client.fetch` is called.

### Steps To Reproduce

Not sure if this is an issue or just expected behavior but I noticed that the query is not actually run by the Trino cluster until data is requested by the client.

So this code will not actually run the query
```python

conn = connect(
    host="<host>",
    port=<port>,
    user="<username>",
    catalog="<catalog>",
    schema="<schema>",
)
cur = conn.cursor()
cur.execute("CREATE TABLE hive.<schema>.<name> AS SELECT 'a' AS a")
```
Nothing happens on the Trino server at this point.

The query only runs when we call fetch. This line is what kicks off the query.
```python
cur.fetch()
```

Basically it seems like the initial post request does not actual kick off the query. This seems unintuitive to me. Since I would have thought that execute would actually run the query. As an example if I am trying to insert data into a table, I don't feel the need to call `fetch` because there is no data to fetch. I just want the query to run and I want to be able to wait until it is successful. I guess I'm wondering what the reasoning is behind the current behavior. I believe the behavior of PyHive was to actually start the query when `execute` was called. Thanks!

### Log output

_No response_

### Operating System

macOS Monterey

0.309.0

### Trino Server version

388

### Python version

 3.8.6

mdesmet:
## Description

Fixes #232, #95

Ensures the received data is properly acknowledged by calling the next_uri. This will avoid seeing failed queries in the query log when executing scalar queries as in the following example.

```
cur.execute("SELECT VERSION()")
cur.fetchone()
cur.cancel()
```

## Release notes

```markdown
## Breaking Changes

* Make the `execute` method of the cursor block until at-least one row is received.
  This means users no longer need to call `fetchone` or `fetchall` to make sure query
  actually starts executing on the Trino server. Note that results still need to be consumed
  by calling `fetchone` or `fetchall` to make sure query isn't considered idle and terminated
  on the server. ([#232](https://trinodb/trino-python-client/issues/232))
```

```markdown
* Properly propagate query failures to the client when using `fetchone`.
  ([#95](https://trinodb/trino-python-client/issues/95))
* Fix queries returning a single row from sometimes appearing as failed on the server.
  ([#220](https://trinodb/trino-python-client/issues/220))
```

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -592,29 +592,35 @@ class TrinoResult(object):
     https://docs.python.org/3/library/stdtypes.html#generator-types
     """
 
-    def __init__(self, query, rows=None):
+    def __init__(self, query, rows: List[Any]):
         self._query = query
-        self._rows = rows or []
+        # Initial rows from the first POST request
+        self._rows = rows
         self._rownumber = 0
+
+    @property
+    def rows(self):
+        return self._rows
+
+    @rows.setter
+    def rows(self, rows):
+        self._rows = rows
 
     @property
     def rownumber(self) -> int:
         return self._rownumber
 
     def __iter__(self):
-        # Initial fetch from the first POST request
-        for row in self._rows:
-            self._rownumber += 1
-            yield row
-        self._rows = None
-
-        # Subsequent fetches from GET requests until next_uri is empty.
-        while not self._query.finished:
-            rows = self._query.fetch()
-            for row in rows:
+        # A query only transitions to a FINISHED state when the results are fully consumed:
+        # The reception of the data is acknowledged by calling the next_uri before exposing the data through dbapi.
+        while not self._query.finished or self._rows is not None:
+            next_rows = self._query.fetch() if not self._query.finished else None
+            for row in self._rows:
                 self._rownumber += 1
                 logger.debug("row %s", row)
                 yield row
+
+            self._rows = next_rows
 
     @property
     def response_headers(self):
@@ -641,7 +647,7 @@ class TrinoQuery(object):
         self._request = request
         self._update_type = None
         self._sql = sql
-        self._result = TrinoResult(self)
+        self._result: Optional[TrinoResult] = None
         self._response_headers = None
         self._experimental_python_types = experimental_python_types
         self._row_mapper: Optional[RowMapper] = None
@@ -652,7 +658,7 @@ class TrinoQuery(object):
             while not self._columns and not self.finished and not self.cancelled:
                 # Columns are not returned immediately after query is submitted.
                 # Continue fetching data until columns information is available and push fetched rows into buffer.
-                self._result._rows += self.fetch()
+                self._result.rows += self.fetch()
         return self._columns
 
     @property
@@ -697,8 +703,11 @@ class TrinoQuery(object):
             self._finished = True
 
         rows = self._row_mapper.map(status.rows) if self._row_mapper else status.rows
-
         self._result = TrinoResult(self, rows)
+
+        # Execute should block until at least one row is received
+        while not self.finished and not self.cancelled and len(self._result.rows) == 0:
+            self._result.rows += self.fetch()
         return self._result
 
     def _update_state(self, status):
--- a/trino/client.py
+++ b/trino/client.py
@@ -930,8 +930,7 @@ class RowMapper:
     """
     Maps a row of data given a list of mapping functions
     """
-
-    def __init__(self, columns=[]):
+    def __init__(self, columns):
         self.columns = columns
 
     def map(self, rows):
--- a/trino/sqlalchemy/dialect.py
+++ b/trino/sqlalchemy/dialect.py
@@ -231,7 +231,7 @@ class TrinoDialect(DefaultDialect):
         """
         ).strip()
         res = connection.execute(sql.text(query), schema=schema, view=view_name)
-        return res.scalar()
+        return res.scalar_one_or_none()
 
     def get_indexes(self, connection: Connection, table_name: str, schema: str = None, **kw) -> List[Dict[str, Any]]:
         if not self.has_table(connection, table_name, schema):
@@ -284,7 +284,7 @@ class TrinoDialect(DefaultDialect):
                 sql.text(query),
                 catalog_name=catalog_name, schema_name=schema_name, table_name=table_name
             )
-            return dict(text=res.scalar())
+            return dict(text=res.scalar_one_or_none())
         except error.TrinoQueryError as e:
             if e.error_name in (
                 error.PERMISSION_DENIED,
@@ -326,7 +326,7 @@ class TrinoDialect(DefaultDialect):
         query = "SELECT version()"
         try:
             res = connection.execute(sql.text(query))
-            version = res.scalar()
+            version = res.scalar_one()
             return tuple([version])
         except exc.ProgrammingError as e:
             logger.debug(f"Failed to get server version: {e.orig.message}")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,8 @@ jobs:
         include:
           # Test with older Trino versions for backward compatibility
           - { python: "3.10", trino: "351" }  # first Trino version
+          # Test with Trino version that requires result set to be fully exhausted
+          - { python: "3.10", trino: "395" }
     env:
       TRINO_VERSION: "${{ matrix.trino }}"
     steps:
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -153,8 +153,8 @@ def test_execute_many_without_params(trino_connection):
     cur = trino_connection.cursor()
     cur.execute("CREATE TABLE memory.default.test_execute_many_without_param (value varchar)")
     cur.fetchall()
-    cur.executemany("INSERT INTO memory.default.test_execute_many_without_param (value) VALUES (?)", [])
     with pytest.raises(TrinoUserError) as e:
+        cur.executemany("INSERT INTO memory.default.test_execute_many_without_param (value) VALUES (?)", [])
         cur.fetchall()
     assert "Incorrect number of parameters: expected 1 but found 0" in str(e.value)
 
@@ -883,13 +883,12 @@ def test_transaction_autocommit(trino_connection_in_autocommit):
     with trino_connection_in_autocommit as connection:
         connection.start_transaction()
         cur = connection.cursor()
-        cur.execute(
-            """
-            CREATE TABLE memory.default.nation
-            AS SELECT * from tpch.tiny.nation
-            """)
-
         with pytest.raises(TrinoUserError) as transaction_error:
+            cur.execute(
+                """
+                CREATE TABLE memory.default.nation
+                AS SELECT * from tpch.tiny.nation
+                """)
             cur.fetchall()
         assert "Catalog only supports writes using autocommit: memory" \
                in str(transaction_error.value)
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -37,10 +37,10 @@ def sample_post_response_data():
     """
 
     yield {
-        "nextUri": "coordinator:8080/v1/statement/20210817_140827_00000_arvdv/1",
+        "nextUri": "https://coordinator:8080/v1/statement/20210817_140827_00000_arvdv/1",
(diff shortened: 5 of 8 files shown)


trino/client.py
```diff
@@ -603,20 +604,19 @@ def rownumber(self) -> int:
...
-            yield self._map_row(self._experimental_python_types, row, self._query.columns)
-        self._rows = None
-
-        # Subsequent fetches from GET requests until next_uri is empty.
-        while not self._query.finished:
-            rows = self._query.fetch()
+        # A query only transitions to a FINISHED state when the results are fully consumed:
+        # The reception of the data is acknowledged by calling the next_uri before exposing the data through dbapi.
+        while not self._query.finished or self._rows is not None:
+            rows = self._rows
```
nineinchnick, line 610: `rows` is not needed, it would be enough to iterate over `_row`, especially if it's only assigned at the end of the loop.

nineinchnick (approved):

mdesmet:
The CI issues can be simulated by following code. Seems that although the result has returned the Trino API still provides one or more next_uri's to fetch.

```
def test_query_cancellation_not_triggered(trino_connection):
    count_not_finished = 0
    for _ in range(0, 1000):
        cur.execute("SELECT VERSION()")
        cur.fetchone()
        if not cur._query.finished:
            count_not_finished += 1

    print(str(count_not_finished) + " unfinished queries")
```

Some questions:

- can we handle this on the serverside to not let this type of queries fail if all results have been consumed?
- Why do we mark a cancelled query as failed if it's a valid use case to only retrieve a number of results and bail out?

hashhar:
> Why do we mark a cancelled query as failed if it's a valid use case to only retrieve a number of results and bail out?

It's not valid. It's a hack that BI tools and clients use instead of limiting their query to pull only what is needed.
e.g. As a cluster admin it's very useful to see clients who run a query that returns billions of rows but just take 100 rows and leave the query hanging (instead of either cancelling or consuming results) which means that the output buffer on the server (and other processes) keeps occupying memory until the query times out.

> can we handle this on the serverside to not let this type of queries fail if all results have been consumed?

You know that `version()` would return one row but the server does not since results are streamed back to coordinator from workers and coordinator can't know that there isn't more data coming until it asks workers about it.

Change your experiment to a query which returns arbitrary number of rows and then you can't know anymore whether query is finished or not. Special casing the client protocol for queries which return single row doesn't seem useful.

hashhar:
The alternative is to have a client protocol which is based on persistent TCP connections instead of HTTP long polling - which brings it's own set of problems.

hashhar:
Also, specifically on why you cannot assume an empty rows being returned from API as proof that query has finished is the pipelined execution model. In queries it's possible for Trino to perform both table scans and output results at the same time. e.g. If you have a long chain of UNION ALL statements then Trino can start returning results as soon as the first UNION ALL query is done while other parts of the query are still executing .

This means that the client might observe periods of time where there is no data returned but a nextUri is still included. If the client were to assume no data == query finished then it'll drop any upcoming rows that will be produced.

nineinchnick:
@hashhar would you agree that to address the original issue we should add a fetch call after `scalar()` to drain the cursor? We might also want to document this somewhere. I don't think we should try to make the driver too smart to work around these protocol limitations.

trino/client.py
```diff
@@ -594,7 +594,8 @@ class TrinoResult(object):
 
     def __init__(self, query, rows=None, experimental_python_types: bool = False):
         self._query = query
-        self._rows = rows or []
+        # Initial rows from the first POST request
```
hashhar, line 597: There's no guarantee initial rows will be present

trino/client.py
```diff
@@ -603,20 +604,17 @@ def rownumber(self) -> int:
...
-            yield self._map_row(self._experimental_python_types, row, self._query.columns)
-        self._rows = None
-
-        # Subsequent fetches from GET requests until next_uri is empty.
-        while not self._query.finished:
-            rows = self._query.fetch()
-            for row in rows:
+        # A query only transitions to a FINISHED state when the results are fully consumed:
+        # The reception of the data is acknowledged by calling the next_uri before exposing the data through dbapi.
+        while not self._query.finished or self._rows is not None:
```
hashhar, line 609: the or is wrong. rows can be missing while the query is not yet finished.

hashhar:
@nineinchnick I don't agree 100%. It's still useful to make the client as smart as the JDBC driver where the implementation detail of the Trino REST protocol isn't visible to users.

But yes, since this might take more time than the quickfix I think we should go with the quickfix for now and then think about how to stop the protocol from leaking into user code.

trino/client.py
mdesmet: 2 points here:

- It seems that there is a `TrinoResult` instantiation in  https://github.com/trinodb/trino-python-client/blob/aee6064c53536da6a73d0f2fa387eb2dfb222600/trino/client.py#L726. I'm not sure why that would be useful, as before fetching a query should be executed which will override the `_result` field in `TrinoQuery`.
- Trino's `QueryResults` can indeed expose a `null` data field, potentially on each fetch response, but that's handled in https://github.com/trinodb/trino-python-client/blob/aee6064c53536da6a73d0f2fa387eb2dfb222600/trino/client.py#L568

mdesmet:
> @hashhar would you agree that to address the original issue we should add a fetch call after `scalar()` to drain the cursor? We might also want to document this somewhere. I don't think we should try to make the driver too smart to work around these protocol limitations.

This is exactly what sqlalchemy does when you `scalar_one` or `scalar_one_or_none`. Again we already fetch the next row in client module's `fetch` now, but even with fetching another record we have no guarantee that would finish the query (nextUri set to null) as @hashhar said.

https://github.com/sqlalchemy/sqlalchemy/blob/f8c4dba4e9f130c18ce00597c036bc26ae7abf90/lib/sqlalchemy/engine/result.py#L745-L748

trino/client.py
```diff
@@ -697,6 +697,9 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
         rows = self._row_mapper.map(status.rows) if self._row_mapper else status.rows
 
         self._result = TrinoResult(self, rows)
+        # Execute should block until at least one row is received
+        while not self.finished and not self.cancelled and len(self._result._rows) == 0:
+            self._result._rows += self.fetch()
```
hovaesco, line 702: You could add `rows` getter property and setter to `TrinoQuery` to not directly access `_rows` and then change it also in `columns` property.

hovaesco (approved):
LGTM % https://github.com/trinodb/trino-python-client/pull/220#discussion_r970515709

hashhar (commented):
This is a clever solution to the original problem. Thanks for coming up with the idea to make `execute` block for first response.

However this is also a breaking change (even though it improves experience) so I'm approving it but not merging until we release a version from current `master` so that people can upgrade to that version and then choose to decide whether they want to stay there for sometime before migrating to new blocking API.

And now that we are taking this direction it might be wise to decouple the Trino API handling from the db-api client and instead make available the fetched rows to the db-api cursor via a queue (list) instead of it directly fetching things from API. That gives future flexibility to introduce performance enhancements as well like the double-bufferring that the JDBC driver does for example and also make it possible to provide different cursor implementations.

hashhar:
@findepi could you help with a release from current master (and also see my comment above)?

(I don't consider this issue a release blocker since it's a change in server behaviour only and as long as you call `fetchall` you're guaranteed to not run into this issue).

findepi:
> could you help with a release from current master

sure, will coordinate over DM

mdesmet:
> However this is also a breaking change (even though it improves experience) so I'm approving it but not merging until we release a version from current `master` so that people can upgrade to that version and then choose to decide whether they want to stay there for sometime before migrating to new blocking API.

Can you explain why you think this is a breaking change? IMHO the only way to break existing usage is if users didn't catch exceptions on` execute()`, which is a bug already as query submission can throw the same exceptions as `fetch*()`. The `fetch*()` operations continue to work as before, as proven by the integration tests.

Note that sqlalchemy doesn't call `fetch*()` on DML operations. Bringing this in would fix that.

> And now that we are taking this direction it might be wise to decouple the Trino API handling from the db-api client and instead make available the fetched rows to the db-api cursor via a queue (list) instead of it directly fetching things from API. That gives future flexibility to introduce performance enhancements as well like the double-bufferring that the JDBC driver does for example and also make it possible to provide different cursor implementations.

IMHO this is already decoupled. The `TrinoQuery` exists in client module and the exposes a lazy collection (an Iterator powered by a generator). I think this is the correct abstraction to use. I don't see leakage of the API details being introduced in this PR, actually the opposite is true: I would argue that in current code `execute()` is only query submission while `fetch*()` is query execution and result set scrolling, while this PR makes `execute()` query submission and execution and `fetch*()` result set scrolling which seems semantically more correct and in line with other dbapi implementations.

The [double buffering as in the java Trino client](https://github.com/mdesmet/trino/blob/8f27eca7ca34f172a95680d706c8c9427a5176b2/client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoResultSet.java#L235-L265), is implemented in this PR. Note that also the Java client doesn't use threading at this moment. I think it is a good idea to investigate but can be done independent from this PR.

I don't see why cursor implementations are impacted, cursors would take the `Iterable` and convert it for example in a dict instead of a tuple (many dbapi implementations have a `DictCursor`).

hashhar:
> Can you explain why you think this is a breaking change? 

Because the API is now blocking.

>  I don't see leakage of the API details being introduced in this PR, actually the opposite is true:

I don't mean that this PR leaks the API details. I meant the opposite. Now we are one step closer to hide the API details within TrinoQuery and TrinoResult. An example of what this PR allows to do (but probably doesn't make sense) is to have a different impl of TrinoQuery which can probably use a different fictional transport mechanism to talk to Trino (instead of the REST API).

> The [double buffering as in the java Trino client](https://github.com/mdesmet/trino/blob/8f27eca7ca34f172a95680d706c8c9427a5176b2/client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoResultSet.java#L235-L265), is implemented in this PR.

True

> Note that also the Java client doesn't use threading at this moment. I think it is a good idea to investigate but can be done independent from this PR.

Exactly what I said above.

> I don't see why cursor implementations are impacted, cursors would take the Iterable and convert it for example in a dict instead of a tuple (many dbapi implementations have a DictCursor).

Again exactly what I said your PR allows us to do in future.

trino/sqlalchemy/dialect.py
```diff
@@ -231,7 +231,7 @@ def get_view_definition(self, connection: Connection, view_name: str, schema: st
         """
         ).strip()
         res = connection.execute(sql.text(query), schema=schema, view=view_name)
-        return res.scalar()
+        return res.scalar_one_or_none()
```
hashhar, line 234: can the sqlalchemy changes be a preparatory commit? Some context in the commit message will be useful (either as separate commit if possible or in current commit if not).

hashhar (approved):

hashhar:
Newer Trino versions will include https://github.com/trinodb/trino/pull/14122 which can mean CI can be green without this change.

I think we should add one more entry to matrix with 395 as the version being tested for the meantime.

trino/sqlalchemy/dialect.py
mdesmet: Done

mdesmet:
> Newer Trino versions will include [trinodb/trino#14122](https://github.com/trinodb/trino/pull/14122) which can mean CI can be green without this change.
> 
> I think we should add one more entry to matrix with 395 as the version being tested for the meantime.

I added the entry.

hashhar (approved):
[Acknowledge reception of data in TrinoResult](https://github.com/trinodb/trino-python-client/pull/220/commits/3f07f68402fab647d7cf3764407ae731e48589c8) fails without [Execute should block until at least one row is received](https://github.com/trinodb/trino-python-client/pull/220/commits/2539a2c1bcc3467907d1c55ef183c5395b1bb85f)

So it seems we need to squash those.

LGTM otherwise.

mdesmet:
@hashhar: Squashed those commits. Are we good to go?

hashhar:
Thanks. Good to go.

PR #225, opened 2022-08-18
Support custom cache for OAuth2 tokens

### LINKED ISSUE #223: Custom cache for oauth2 tokens
Hello, 
I have a flask app, which uses trino python client to query the data from my trino servers and I have enabled oatuh2 authentication against my azure active directory, currently, the token is caching per host, I need to have it cached per user who authenticates on the front end of my web application.  I have implemented my custom cache below and it is working as per my use case, but I prefer not to change any private properties as it's not the best practice.
```
from  flask import session


class _CustomCache(_OAuth2TokenCache):
    """
    In-memory token cache implementation. The token is stored per user.
    """

    def __init__(self):
        self._cache = {}
    def get_token_from_cache(self, host: str) -> Optional[str]:
        userName=session['user']   
        return self._cache.get(userName)

    def store_token_to_cache(self, host: str, token: str) -> None:
        userName=session['user']  
        self._cache[userName] = token
    


temp=OAuth2Authentication()
temp._bearer._token_cache=_CustomCache()


conn=connect(

    host='******',
    port=443,
    auth=temp,
    http_scheme="https"  
)

cursor=conn.cursor()


```

mdesmet:
Fixes #223


trino/auth.py
```diff
@@ -291,6 +290,14 @@ def __call__(self, r):
 
         return r
 
+    def _setup_cache(self, custom_cache):
+        if custom_cache is not None:
+            return custom_cache
```
hovaesco, line 295: Maybe is worth to add some validation and error handling for `custom_cache`?

README.md
```diff
@@ -198,6 +198,8 @@ A callback to handle the redirect url can be provided via param `redirect_auth_u
 
 The OAuth2 token will be cached either per `trino.auth.OAuth2Authentication` instance or, when keyring is installed, it will be cached within a secure backend (MacOS keychain, Windows credential locker, etc) under a key including host of the Trino connection. Keyring can be installed using `pip install 'trino[external-authentication-token-cache]'`.
 
+A custom caching implementation can be provided by creating a class implementing the `trino.auth.OAuth2TokenCache` abstract class and adding it as in `OAuth2Authentication(cache=my_custom_cache_impl)`. The custom caching implementation enables usage in multi-user environments (notebooks, web applications) in combination with a custom `redirect_auth_url_handler` as explained above.
```
hovaesco, line 201: Let's explain further what can be passed as `my_custom_cache_impl` or give some code snippets here.

trino/auth.py
mdesmet: I've added an `isinstance` check. I don't think there is anything more we can do. We might rethrow the exception as one of the dbapi exception types. WDYT?

README.md
mdesmet: Done

hashhar:
I'd prefer if we made the defaults more useful instead of adding hooks for custom behaviour and leaving users to implement them themselves.

It seems default should be to cache based on identity (which is what Trino itself seems to do).

mdesmet:
> I'd prefer if we made the defaults more useful instead of adding hooks for custom behaviour and leaving users to implement them themselves.

That won't work in a distributed web application. You would need to store and retrieve the tokens in a distributed cache (eg redis) as every request may be handled by different nodes. 

> It seems default should be to cache based on identity (which is what Trino itself seems to do).

This is not entirely true. The user is not required in the `connect(args)` for OAuth connections. Without the user we would only be able to retrieve the identity by running a `SELECT CURRENT_USER()` query.

PR #228, opened 2022-08-19
Don't automatically label issues

hashhar:
Issues must be triaged before being labelled as a bug - which confirms
it's a bug to be fixed instead of just using something wrong or being
labelled as an enhancement - which confirms it's something we think
should be added.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,7 +1,6 @@
 ---
 name: Bug report
 description: Report a bug or an issue you've found with Trino Python client
-labels: bug
 body:
   - type: textarea
     attributes:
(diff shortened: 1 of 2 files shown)


hashhar:
We can auto-add a "triage" label to them however if it helps to identify which issues need to be looked at and which ones are already looked at.

mdesmet:
> We can auto-add a "triage" label to them however if it helps to identify which issues need to be looked at and which ones are already looked at.

I'm more in favour of adding a triage label, it would most clear towards every stakeholder in what stage the issue resides.

martint:
Wouldn’t that be the default state of an issue? Why do we need a separate label?

mdesmet:
My suggestion would be to add both triage and the issue type label (bug or enhancement) and then remove the triage label after triage counting on the fact that most issues will be correctly labeled by the issue reporter.

hashhar:
@mdesmet That's rarely the case. Lots of issues are for troubleshooting or asking for a feature we don't plan to implement.

martint:
Also, a triage label can be ambiguous. Triage for what purpose? When is triaging “done”? Is it just a way of saying “the bug/feature label is tentative” or does it also involve investigating, and reproducing a bug, or making a decision to move forward with a feature?

ebyhr (approved):

PR #229, opened 2022-08-23
Add variable precision time and timestamp support

lpoulain:
Add variable precision time and timestamp support for the experimental_python_types flag by passing the `X-Trino-Client-Capabilities: PARAMETRIC_DATETIME` header in the query request, just like the Trino JDBC and ODBC drivers do.

Updates the tests/integration/test_types_integration.py accordingly.

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -283,7 +283,7 @@ def test_datetime_query_param(trino_connection):
     rows = cur.fetchall()
 
     assert rows[0][0] == params
-    assert cur.description[0][1] == "timestamp"
+    assert cur.description[0][1] == "timestamp(6)"
 
 
 def test_datetime_with_utc_time_zone_query_param(trino_connection):
@@ -295,7 +295,7 @@ def test_datetime_with_utc_time_zone_query_param(trino_connection):
     rows = cur.fetchall()
 
     assert rows[0][0] == params
-    assert cur.description[0][1] == "timestamp with time zone"
+    assert cur.description[0][1] == "timestamp(6) with time zone"
 
 
 def test_datetime_with_numeric_offset_time_zone_query_param(trino_connection):
@@ -309,19 +309,19 @@ def test_datetime_with_numeric_offset_time_zone_query_param(trino_connection):
     rows = cur.fetchall()
 
     assert rows[0][0] == params
-    assert cur.description[0][1] == "timestamp with time zone"
+    assert cur.description[0][1] == "timestamp(6) with time zone"
 
 
 def test_datetime_with_named_time_zone_query_param(trino_connection):
     cur = trino_connection.cursor(experimental_python_types=True)
 
-    params = datetime(2020, 1, 1, 16, 43, 22, 320000, tzinfo=pytz.timezone('America/Los_Angeles'))
+    params = pytz.timezone('America/Los_Angeles').localize(datetime(2020, 1, 1, 16, 43, 22, 320000))
 
     cur.execute("SELECT ?", params=(params,))
     rows = cur.fetchall()
 
     assert rows[0][0] == params
-    assert cur.description[0][1] == "timestamp with time zone"
+    assert cur.description[0][1] == "timestamp(6) with time zone"
 
 
 def test_datetime_with_trailing_zeros(trino_connection):
@@ -371,7 +371,7 @@ def test_doubled_datetimes(trino_connection):
     cur.execute("SELECT ?", params=(params,))
     rows = cur.fetchall()
 
-    assert rows[0][0] == datetime(2002, 10, 27, 1, 30, 0, tzinfo=pytz.timezone('US/Eastern'))
+    assert rows[0][0] == pytz.timezone('US/Eastern').localize(datetime(2002, 10, 27, 1, 30, 0))
 
     cur = trino_connection.cursor(experimental_python_types=True)
 
@@ -380,7 +380,7 @@ def test_doubled_datetimes(trino_connection):
     cur.execute("SELECT ?", params=(params,))
     rows = cur.fetchall()
 
-    assert rows[0][0] == datetime(2002, 10, 27, 1, 30, 0, tzinfo=pytz.timezone('US/Eastern'))
+    assert rows[0][0] == pytz.timezone('US/Eastern').localize(datetime(2002, 10, 27, 1, 30, 0))
 
 
 def test_date_query_param(trino_connection):
--- a/tests/integration/test_types_integration.py
+++ b/tests/integration/test_types_integration.py
@@ -1,6 +1,5 @@
 import math
-import datetime
-from datetime import timedelta
+from datetime import timedelta, datetime, date, time
 import pytest
 import pytz
 from decimal import Decimal
@@ -206,19 +205,19 @@ def test_digest(trino_connection):
 def test_date(trino_connection):
     SqlTest(trino_connection) \
         .add_field(sql="CAST(null AS DATE)", python=None) \
-        .add_field(sql="DATE '2001-08-22'", python=datetime.date(2001, 8, 22)) \
-        .add_field(sql="DATE '0001-01-01'", python=datetime.date(1, 1, 1)) \
-        .add_field(sql="DATE '1582-10-04'", python=datetime.date(1582, 10, 4)) \
-        .add_field(sql="DATE '1582-10-05'", python=datetime.date(1582, 10, 5)) \
-        .add_field(sql="DATE '1582-10-14'", python=datetime.date(1582, 10, 14)) \
+        .add_field(sql="DATE '2001-08-22'", python=date(2001, 8, 22)) \
+        .add_field(sql="DATE '0001-01-01'", python=date(1, 1, 1)) \
+        .add_field(sql="DATE '1582-10-04'", python=date(1582, 10, 4)) \
+        .add_field(sql="DATE '1582-10-05'", python=date(1582, 10, 5)) \
+        .add_field(sql="DATE '1582-10-14'", python=date(1582, 10, 14)) \
         .execute()
 
 
 def test_time(trino_connection):
-    time_0 = datetime.time(1, 23, 45)
-    time_3 = datetime.time(1, 23, 45, 123000)
-    time_6 = datetime.time(1, 23, 45, 123456)
-    time_round = datetime.time(1, 23, 45, 123457)
+    time_0 = time(1, 23, 45)
+    time_3 = time(1, 23, 45, 123000)
+    time_6 = time(1, 23, 45, 123456)
+    time_round = time(1, 23, 45, 123457)
 
     SqlTest(trino_connection) \
         .add_field(sql="CAST(null AS TIME)", python=None) \
@@ -243,16 +242,16 @@ def test_time_with_timezone(trino_connection):
 
 
 def query_time_with_timezone(trino_connection, tz_str):
-    tz = datetime.datetime.strptime('+00:00', "%z").tzinfo
+    tz = datetime.strptime('+00:00', "%z").tzinfo
 
     hours_shift = int(tz_str[:3])
     minutes_shift = int(tz_str[4:])
     delta = timedelta(hours=hours_shift, minutes=minutes_shift)
 
-    time_0 = (datetime.datetime(2, 1, 1, 11, 23, 45, 0) - delta).time().replace(tzinfo=tz)
-    time_3 = (datetime.datetime(2, 1, 1, 11, 23, 45, 123000) - delta).time().replace(tzinfo=tz)
-    time_6 = (datetime.datetime(2, 1, 1, 11, 23, 45, 123456) - delta).time().replace(tzinfo=tz)
-    time_round = (datetime.datetime(2, 1, 1, 11, 23, 45, 123457) - delta).time().replace(tzinfo=tz)
+    time_0 = (datetime(2, 1, 1, 11, 23, 45, 0) - delta).time().replace(tzinfo=tz)
+    time_3 = (datetime(2, 1, 1, 11, 23, 45, 123000) - delta).time().replace(tzinfo=tz)
+    time_6 = (datetime(2, 1, 1, 11, 23, 45, 123456) - delta).time().replace(tzinfo=tz)
+    time_round = (datetime(2, 1, 1, 11, 23, 45, 123457) - delta).time().replace(tzinfo=tz)
 
     SqlTest(trino_connection) \
         .add_field(sql="CAST(null AS TIME WITH TIME ZONE)", python=None) \
@@ -271,14 +270,14 @@ def query_time_with_timezone(trino_connection, tz_str):
 
 
 def test_timestamp(trino_connection):
-    timestamp_0 = datetime.datetime(2001, 8, 22, 1, 23, 45, 0)
-    timestamp_3 = datetime.datetime(2001, 8, 22, 1, 23, 45, 123000)
-    timestamp_6 = datetime.datetime(2001, 8, 22, 1, 23, 45, 123456)
-    timestamp_round = datetime.datetime(2001, 8, 22, 1, 23, 45, 123457)
-    timestamp_ce = datetime.datetime(1, 1, 1, 1, 23, 45, 123000)
-    timestamp_julian = datetime.datetime(1582, 10, 4, 1, 23, 45, 123000)
-    timestamp_during_switch = datetime.datetime(1582, 10, 5, 1, 23, 45, 123000)
-    timestamp_gregorian = datetime.datetime(1582, 10, 14, 1, 23, 45, 123000)
+    timestamp_0 = datetime(2001, 8, 22, 1, 23, 45, 0)
+    timestamp_3 = datetime(2001, 8, 22, 1, 23, 45, 123000)
+    timestamp_6 = datetime(2001, 8, 22, 1, 23, 45, 123456)
+    timestamp_round = datetime(2001, 8, 22, 1, 23, 45, 123457)
+    timestamp_ce = datetime(1, 1, 1, 1, 23, 45, 123000)
+    timestamp_julian = datetime(1582, 10, 4, 1, 23, 45, 123000)
+    timestamp_during_switch = datetime(1582, 10, 5, 1, 23, 45, 123000)
+    timestamp_gregorian = datetime(1582, 10, 14, 1, 23, 45, 123000)
 
     SqlTest(trino_connection) \
         .add_field(sql="CAST(null AS TIMESTAMP)", python=None) \
@@ -317,7 +316,7 @@ def query_timestamp_with_timezone(trino_connection, tz_str):
         minutes_shift = int(tz_str[4:])
     else:
         tz = pytz.timezone(tz_str)
-        offset = tz.utcoffset(datetime.datetime.now())
+        offset = tz.utcoffset(datetime.now())
         offset_seconds = offset.total_seconds()
         hours_shift = int(offset_seconds / 3600)
         minutes_shift = offset_seconds % 3600 / 60
@@ -325,10 +324,10 @@ def query_timestamp_with_timezone(trino_connection, tz_str):
     tz = pytz.timezone('Etc/GMT')
     delta = timedelta(hours=hours_shift, minutes=minutes_shift)
 
-    timestamp_0 = tz.localize(datetime.datetime(2001, 8, 22, 11, 23, 45, 0)) - delta
-    timestamp_3 = tz.localize(datetime.datetime(2001, 8, 22, 11, 23, 45, 123000)) - delta
-    timestamp_6 = tz.localize(datetime.datetime(2001, 8, 22, 11, 23, 45, 123456)) - delta
-    timestamp_round = tz.localize(datetime.datetime(2001, 8, 22, 11, 23, 45, 123457)) - delta
+    timestamp_0 = tz.localize(datetime(2001, 8, 22, 11, 23, 45, 0)) - delta
+    timestamp_3 = tz.localize(datetime(2001, 8, 22, 11, 23, 45, 123000)) - delta
+    timestamp_6 = tz.localize(datetime(2001, 8, 22, 11, 23, 45, 123456)) - delta
+    timestamp_round = tz.localize(datetime(2001, 8, 22, 11, 23, 45, 123457)) - delta
 
     SqlTest(trino_connection) \
         .add_field(sql="CAST(null AS TIMESTAMP WITH TIME ZONE)", python=None) \
--- a/tests/integration/test_types_integration.py
+++ b/tests/integration/test_types_integration.py
@@ -1,5 +1,8 @@
 import math
+import datetime
+from datetime import timedelta
 import pytest
+import pytz
 from decimal import Decimal
 import trino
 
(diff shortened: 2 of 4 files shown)


tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+        .add_field(sql="CAST(null AS TIME(0) WITH TIME ZONE)", python=None) \
+        .add_field(sql="CAST(null AS TIME(3) WITH TIME ZONE)", python=None) \
+        .add_field(sql="CAST(null AS TIME(6) WITH TIME ZONE)", python=None) \
+        .add_field(sql="CAST(null AS TIME(9) WITH TIME ZONE)", python=None) \
+        .add_field(sql="CAST(null AS TIME(12) WITH TIME ZONE)", python=None) \
+        .add_field(sql="CAST('01:23:45 %s' AS TIME(0) WITH TIME ZONE)" % (timezone_str), python=time_0) \
+        .add_field(sql="TIME '01:23:45.123 %s'" % (timezone_str), python=time_3) \
+        .add_field(sql="CAST('01:23:45.123 %s' AS TIME(3) WITH TIME ZONE)" % (timezone_str), python=time_3) \
+        .add_field(sql="CAST('01:23:45.123456 %s' AS TIME(6) WITH TIME ZONE)" % (timezone_str), python=time_6) \
+        .add_field(sql="CAST('01:23:45.123456789 %s' AS TIME(9) WITH TIME ZONE)" % (timezone_str), python=time_6) \
```
nineinchnick, line 256: That may be a naive question, but is there a way of reading results with data types with high precision without them getting truncated? Would it require modifying the query to cast them to some other data type or can the driver be configured to return them in some raw form, like strings?

trino/client.py
```diff
@@ -686,7 +686,14 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
         if self.cancelled:
             raise exceptions.TrinoUserError("Query has been cancelled", self.query_id)
 
-        response = self._request.post(self._sql, additional_http_headers)
+        if self._experimental_python_types:
+            http_headers = {constants.HEADER_CLIENT_CAPABILITIES: 'PARAMETRIC_DATETIME'}
```
nineinchnick, line 690: Another naive question - how come this doesn't require any other changes in how the client parses results from the server?

nineinchnick: What's the current behavior, without setting this header?

hashhar: The server just starts sending back the precision in the type signatures and the values have additional precision to match.

hashhar (commented):
cc: @findepi since he's more familiar with the parametric timestamp changes.

tests/integration/test_types_integration.py
findepi: i don't think we should automatically switch between `datetime.time` and `str` depending on the precision.
in JDBC, user needs to explicitly ask for a String to get one (this works for all precisions).
Is it possible with Python cursor API?

findepi (commented):
lgtm

tests/integration/test_types_integration.py
nineinchnick: I didn't suggest switching based on precision. There must be an option for users that work with high-precision timestamps.

trino/client.py
nineinchnick: The client capabilities doesn't seem to be documented anywhere, but I found this comment in the server: https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/server/protocol/QueryResultRows.java#L355
So looks like it was rounded before in the server, and now it'll get truncated in the client, but still with higher precision (6 vs 3). So it's an improvement.

nineinchnick (approved):

tests/integration/test_types_integration.py
hashhar: @findepi Note that until recently everything returned by the Python client was untyped and `str`.

So it's still possible to return `str` when `experimental_python_types` is not enabled on the cursor.

But yes, for strongly typed time objects Python's stdlib is very limiting and there's nothing like `getObject(idx, clazz)` in DB-API.

hashhar:
one question - why can this new behaviour not be exposed for cursors without `experimental_python_types=true`?

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
         .execute()
 
 
+def test_date(trino_connection):
+    SqlTest(trino_connection) \
+        .add_field(sql="CAST(null AS DATE)", python=None) \
+        .add_field(sql="DATE '2001-08-22'", python=datetime.date(2001, 8, 22)) \
+        .add_field(sql="DATE '02001-08-22'", python=datetime.date(2001, 8, 22)) \
```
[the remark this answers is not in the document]
hashhar: 🤯 It seems Trino allows this for some reason. @findepi Do you know why?

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+        .execute()
+
+
+def test_time(trino_connection):
+    time_0 = datetime.time(1, 23, 45)
+    time_3 = datetime.time(1, 23, 45, 123000)
+    time_6 = datetime.time(1, 23, 45, 123456)
```
hashhar, line 215: Can you check whether the JDBC driver truncates or rounds values with higher precision?

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+        .add_field(sql="CAST('01:23:45.123456' AS TIME(6))", python=time_6) \
+        .add_field(sql="CAST('01:23:45.123456789' AS TIME(9))", python=time_6) \
+        .add_field(sql="CAST('01:23:45.123456789123' AS TIME(12))", python=time_6) \
+        .execute()
+
+
+def test_time_with_timezone(trino_connection):
+    query_time_with_timezone(trino_connection, '-08:00')
+    query_time_with_timezone(trino_connection, '+08:00')
+    query_time_with_timezone(trino_connection, '+05:30')
```
hashhar, line 236: Can we also test named timezones?

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+        .execute()
+
+
+def test_timestamp_with_timezone(trino_connection):
+    execute_timestamp_with_timezone(trino_connection, '-08:00')
+    execute_timestamp_with_timezone(trino_connection, '+08:00')
+    execute_timestamp_with_timezone(trino_connection, '+05:30')
+
+
+def execute_timestamp_with_timezone(trino_connection, timezone_str):
```
hashhar, line 288: nit: the other method is called `query_*`. Rename both to be consistent.

trino/client.py
lpoulain: All the other Trino clients I've seen (JDBC, ODBC, Go) are all passing `X-Trino-Client-Capabilityes: PARAMETRIC_DATETIME` to Trino when executing a query. The Python library is the only one that doesn't. Unfortunately, Python datetimes only support a 6-digit precision max.

lpoulain: @nineinchnick as far as how come no other change is required on how the client parses the results from the server, it's because in my last PR I made sure that the code worked with either type of result.

tests/integration/test_types_integration.py
lpoulain: Do you have examples of SQL queries that generate such named timezones?

hashhar: `select timestamp '2022-01-01 01:01:01.123456789 Asia/Kolkata'`. 

We should probably add examples of this as well to https://trino.io/docs/current/language/types.html#date-and-time. cc: @mosabua @jhlodin can you run with this?

tests/integration/test_types_integration.py
lpoulain: The Java `Timestamp` type seems to support types up to `TIMESTAMP(9)`. Everything beyond this is truncated, but the nanoseconds are returned by the JDBC driver.

tests/integration/test_types_integration.py
lpoulain: The code right now only supports timezone in the `+08:00` format. Trying to find a way to parse any type of timezone (named or not), although I'm not entirely sure if it's possible.

jhlodin: `TIMESTAMP WITH TIME ZONE` docs currently say that "Time zones are expressed as the numeric UTC offset value", so if that's the only support format then that's correct for now. If we add support for named timezones let us know and we can run with the docs.

tests/integration/test_types_integration.py
lpoulain: Good catch. Fixed.

tests/integration/test_types_integration.py
hashhar: @jhlodin Named offsets are supported. Try out the example from my comment above.

tests/integration/test_types_integration.py
hashhar: Thanks, just wanted to confirm whether behaviour is truncation or rounding. It seems however that the behaviour is rounding by seeing https://github.com/trinodb/trino/blob/a977d88b051e6616d1fd9c355362627383abec29/client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java#L2012.

We should be consistent across drivers where possible.

tests/integration/test_types_integration.py
findepi: What other database clients do? (for DBs with more than ms precision)

tests/integration/test_types_integration.py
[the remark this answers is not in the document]
findepi: DATE literals (and varchar cast to date) allow zero padding, surrounding whitespace and year sign (+ or -).
It's however unrelated to the client. 
Since `DATE '2001-08-22'` and `DATE '02001-08-22'` is just two ways of representing the same value, py client tests don't (and shouldn't) test both.

tests/integration/test_types_integration.py
findepi: > `TIMESTAMP(P) WITH TIME ZONE` docs currently say that "Time zones are expressed as the numeric UTC offset value",

This is not Trino docs, right?

> In case current code only supports numeric offsets 

i guess we need `pytz` for named zones. (here my py-fu ends)

tests/integration/test_types_integration.py
findepi: It seems it rounds, but IMO it happens here
https://github.com/trinodb/trino/blob/63f2fdab69094776bc464f2db729f8070ff98505/client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java#L2041-L2044

tests/integration/test_types_integration.py
jhlodin: @findepi It is Trino docs, see https://trino.io/docs/current/language/types.html#timestamp-p-with-time-zone

jhlodin: Thanks @hashhar , will follow up

mosabua: We used to support named offsets and that got removed...

hashhar: @mosabua We do support IANA names so things like Asia/Kolkata but we don't support the 3 letter non-standard country codes like IST, PST etc.

lpoulain: That's for timestamps. What about time? `SELECT CAST('01:01:01.123456789 Asia/Kolkata' AS TIME WITH TIME ZONE)` is not accepted by Trino.

hashhar: TIME doesn't accept named zones - you're correct. Sorry for the confusion.

hashhar:
@lpoulain Can you please update behavior to match the JDBC driver (rounding instead of truncation)? Other than that (and possibly allowing using this with experimental_python_types disabled) this seems good to me.

tests/integration/test_types_integration.py
hashhar: I know of only Oracle as another database that supports higher than millis precision and they don't expose that to the Python client since stdlib is limited to milliseconds.

tests/integration/test_types_integration.py
lpoulain: Updated the code to round the milliseconds and updated associated tests

tests/integration/test_types_integration.py
findepi: PostgreSQL, MySQL support microsecond timestamps
SQL Server, SAP HANA support timestamp with p=7

hashhar: Sorry I meant micros above. Micros is supported in Python stdlib so no issues with that. I'll see SQL Server and HANA. Thanks for the tip.

lpoulain:
@hashhar done.

tests/integration/test_types_integration.py
mosabua: > @mosabua We do support IANA names so things like Asia/Kolkata but we don't support the 3 letter non-standard country codes like IST, PST etc.

I am pretty sure there was some discussion around this when we updated the docs and removed the 3 letter support that named pairs are also not supported, but cant find the details. From what I recall it was something around the named pairs also not being really standardized fully/and or available in the relevant tzdata libraries and integrations. Can @martint please confirm so we can decide to either update the docs to clarify, and/or improve test coverage, and/or update code to remove support ...

martint: `time(p) with time zone` only supports timezone offsets (hour + minute) and not political time zones, for the reason @hashhar described above.

See https://github.com/trinodb/trino/issues/191

lpoulain: @mosabua @martint @hashhar what's the status about this issue?

tests/integration/test_types_integration.py
```diff
@@ -213,6 +214,7 @@ def test_time(trino_connection):
     time_0 = datetime.time(1, 23, 45)
     time_3 = datetime.time(1, 23, 45, 123000)
     time_6 = datetime.time(1, 23, 45, 123456)
+    time_trunc = datetime.time(1, 23, 45, 123457)
```
aalbu, line 217: Call it `time_round`, maybe?

tests/integration/test_types_integration.py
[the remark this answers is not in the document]
hashhar: ping @lpoulain - let's not test engine level functionality in this test. i.e. the conversion from `02001` to `2001` happens inside Trino so the client shouldn't care at all how or whether it works or not.

It might be more useful to test dates before `1582` (when Julian to Gregorian calendar switch happenned).

- `0001-01-01`
- `1582-10-04` (before the switch)
- `1582-10-05` (beginning of switch)
- `1582-10-14` (end of switch)

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+        .add_field(sql="CAST(null AS TIME(3))", python=None) \
+        .add_field(sql="CAST(null AS TIME(6))", python=None) \
+        .add_field(sql="CAST(null AS TIME(9))", python=None) \
+        .add_field(sql="CAST(null AS TIME(12))", python=None) \
+        .add_field(sql="CAST('01:23:45' AS TIME(0))", python=time_0) \
+        .add_field(sql="TIME '01:23:45.123'", python=time_3) \
+        .add_field(sql="CAST('01:23:45.123' AS TIME(3))", python=time_3) \
+        .add_field(sql="CAST('01:23:45.123456' AS TIME(6))", python=time_6) \
+        .add_field(sql="CAST('01:23:45.123456789' AS TIME(9))", python=time_6) \
+        .add_field(sql="CAST('01:23:45.123456789123' AS TIME(12))", python=time_6) \
```
hashhar: If this test is passing then it seems we aren't performing the rounding.

tests/integration/test_types_integration.py
hashhar: Resolving this comment since `TIME` only supports numeric offsets.

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+    query_time_with_timezone(trino_connection, '-08:00')
+    query_time_with_timezone(trino_connection, '+08:00')
+    query_time_with_timezone(trino_connection, '+05:30')
+
+
+def query_time_with_timezone(trino_connection, timezone_str):
+    tz = datetime.datetime.strptime(timezone_str, "%z").tzinfo
+    time_0 = datetime.time(1, 23, 45, 0).replace(tzinfo=tz)
+    time_3 = datetime.time(1, 23, 45, 123000).replace(tzinfo=tz)
+    time_6 = datetime.time(1, 23, 45, 123456).replace(tzinfo=tz)
```
hashhar, line 246: This isn't testing much. Replace just replaces the timezone and keeps the same values for hour, minute and seconds. We'd want to test whether the same "instant" in time when queried using different zones returns same instant when read back or not.

i.e. `10:10:10 +00:00` == `15:40:10 +05:30`

tests/integration/test_types_integration.py
hashhar: @lpoulain If we're actually performing rounding then `01:23:45.123456789` should've been read back as `01:23:45.123457` instead.

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+        .add_field(sql="CAST('01:23:45.123456 %s' AS TIME(6) WITH TIME ZONE)" % (timezone_str), python=time_6) \
+        .add_field(sql="CAST('01:23:45.123456789 %s' AS TIME(9) WITH TIME ZONE)" % (timezone_str), python=time_6) \
+        .add_field(sql="CAST('01:23:45.123456789123 %s' AS TIME(12) WITH TIME ZONE)" % (timezone_str), python=time_6) \
+        .execute()
+
+
+def test_timestamp(trino_connection):
+    timestamp_0 = datetime.datetime(2001, 8, 22, 1, 23, 45, 0)
+    timestamp_3 = datetime.datetime(2001, 8, 22, 1, 23, 45, 123000)
+    timestamp_6 = datetime.datetime(2001, 8, 22, 1, 23, 45, 123456)
```
hashhar, line 268: test julian to gregorian switch + 0001-01-01

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+        .add_field(sql="CAST(null AS TIMESTAMP(3))", python=None) \
+        .add_field(sql="CAST(null AS TIMESTAMP(6))", python=None) \
+        .add_field(sql="CAST(null AS TIMESTAMP(9))", python=None) \
+        .add_field(sql="CAST(null AS TIMESTAMP(12))", python=None) \
+        .add_field(sql="CAST('2001-08-22 01:23:45' AS TIMESTAMP(0))", python=timestamp_0) \
+        .add_field(sql="TIMESTAMP '2001-08-22 01:23:45.123'", python=timestamp_3) \
+        .add_field(sql="CAST('2001-08-22 01:23:45.123' AS TIMESTAMP(3))", python=timestamp_3) \
+        .add_field(sql="CAST('2001-08-22 01:23:45.123456' AS TIMESTAMP(6))", python=timestamp_6) \
+        .add_field(sql="CAST('2001-08-22 01:23:45.123456789' AS TIMESTAMP(9))", python=timestamp_6) \
+        .add_field(sql="CAST('2001-08-22 01:23:45.123456789123' AS TIMESTAMP(12))", python=timestamp_6) \
```
hashhar: If we're actually rounding this should have failed since value read back would be `.123457` instead.

tests/integration/test_types_integration.py
```diff
@@ -200,6 +201,119 @@ def test_digest(trino_connection):
...
+    execute_timestamp_with_timezone(trino_connection, '+08:00')
+    execute_timestamp_with_timezone(trino_connection, '+05:30')
+
+
+def execute_timestamp_with_timezone(trino_connection, timezone_str):
+    tz = datetime.datetime.strptime(timezone_str, "%z").tzinfo
+
+    timestamp_0 = datetime.datetime(2001, 8, 22, 1, 23, 45, 0, tzinfo=tz)
+    timestamp_3 = datetime.datetime(2001, 8, 22, 1, 23, 45, 123000, tzinfo=tz)
+    timestamp_6 = datetime.datetime(2001, 8, 22, 1, 23, 45, 123456, tzinfo=tz)
```
hashhar, line 306: This is not very meaningful since it replaces the timezone instead of adjusting the datetime to represent same instant in time.

trino/client.py
```diff
@@ -686,6 +686,10 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
         if self.cancelled:
             raise exceptions.TrinoUserError("Query has been cancelled", self.query_id)
 
+        if self._experimental_python_types:
+            new_header = {constants.HEADER_CLIENT_CAPABILITIES: 'PARAMETRIC_DATETIME'}
```
hashhar: nit: extract constant for header value. Do we do this for other well-named header values as well?

trino/client.py
```diff
@@ -686,6 +686,10 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
...
+            additional_http_headers = new_header | additional_http_headers if additional_http_headers else new_header
```
hashhar: avoid using dict unions since they overwrite any existing values which can lead to subtle bugs being introduced.

Ideally we'd want to complain if a header whose value was already set in some-place is then tried to be overwritten (either because of bug in our code or because user is trying to access internals of client).

tests/integration/test_dbapi_integration.py
```diff
@@ -282,7 +282,7 @@ def test_datetime_query_param(trino_connection):
     rows = cur.fetchall()
 
     assert rows[0][0] == params
-    assert cur.description[0][1] == "timestamp"
+    assert cur.description[0][1] == "timestamp(6)"
```
hashhar, line 285: Squash "Fixed failing tests " with previous commit.

tests/integration/test_types_integration.py
```diff
@@ -280,12 +280,12 @@ def test_timestamp(trino_connection):
 
 
 def test_timestamp_with_timezone(trino_connection):
```
hashhar, line 289: Squash "Refactor some test function names for consistency" since it changes something introduced in the first commit itself.

tests/integration/test_types_integration.py
```diff
@@ -249,19 +253,20 @@ def query_time_with_timezone(trino_connection, timezone_str):
...
+        .add_field(sql="CAST('01:23:45.123456789 %s' AS TIME(9) WITH TIME ZONE)" % (tz_str), python=time_trunc) \
+        .add_field(sql="CAST('01:23:45.123456789123 %s' AS TIME(12) WITH TIME ZONE)" % (tz_str), python=time_trunc) \
         .execute()
 
 
 def test_timestamp(trino_connection):
     timestamp_0 = datetime.datetime(2001, 8, 22, 1, 23, 45, 0)
     timestamp_3 = datetime.datetime(2001, 8, 22, 1, 23, 45, 123000)
     timestamp_6 = datetime.datetime(2001, 8, 22, 1, 23, 45, 123456)
+    timestamp_trunc = datetime.datetime(2001, 8, 22, 1, 23, 45, 123457)
```
hashhar, line 269: this is rounded value not truncated

trino/client.py
```diff
@@ -864,19 +864,22 @@ def _double_map_func(self):
     def _timestamp_map_func(self, column, col_type):
         datetime_default_size = 20  # size of 'YYYY-MM-DD HH:MM:SS.' (the datetime string up to the milliseconds)
```
hashhar, line 865: nit: change this to something self-explanatory like `len('YYYY-MM-DD HH:MM:SS.')`.

BTW years are not guaranteed to be 4 digits only.

trino/client.py
```diff
@@ -913,15 +927,27 @@ def _get_time_with_timezome(self, value, time_size, pattern):
...
+        assert len(matches.groups()) == 4
+        if matches.group(2) == '-':
+            tz = -timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
+        else:
+            tz = timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
+        time_str = matches.group(1)[:time_size]
+        ms_str = str(round(int(matches.group(1)[time_size:]) / ms_div))
+        return datetime.strptime(time_str + ms_str, pattern).time().replace(tzinfo=timezone(tz))
+
     def _get_number_of_digits(self, column):
```
hashhar, line 942: this method is doing too many things and it's unclear what the return value means and method name doesn't help either

trino/client.py
```diff
@@ -913,15 +927,27 @@ def _get_time_with_timezome(self, value, time_size, pattern):
             tz = timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
         return datetime.strptime(matches.group(1)[:time_size], pattern).time().replace(tzinfo=timezone(tz))
 
+    def _get_time_with_timezone_trunc_ms(self, value, time_size, ms_div, pattern):
```
hashhar, line 930: `truncate_time_with_timezone_to_millis`?

trino/client.py
```diff
@@ -913,15 +927,27 @@ def _get_time_with_timezome(self, value, time_size, pattern):
...
+        matches = re.match(r'^(.*)([\+\-])(\d{2}):(\d{2})$', value)
+        assert matches is not None
+        assert len(matches.groups()) == 4
+        if matches.group(2) == '-':
+            tz = -timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
+        else:
+            tz = timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
+        time_str = matches.group(1)[:time_size]
+        ms_str = str(round(int(matches.group(1)[time_size:]) / ms_div))
```
hashhar, line 939: `millis_str`

trino/client.py
hashhar: `ms_div` -> more meaningful name

trino/client.py
```diff
@@ -913,15 +927,27 @@ def _get_time_with_timezome(self, value, time_size, pattern):
             tz = timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
         return datetime.strptime(matches.group(1)[:time_size], pattern).time().replace(tzinfo=timezone(tz))
 
+    def _get_time_with_timezone_trunc_ms(self, value, time_size, ms_div, pattern):
+        matches = re.match(r'^(.*)([\+\-])(\d{2}):(\d{2})$', value)
```
hashhar, line 931: can we use named groups instead - more readable code

trino/client.py
```diff
@@ -864,19 +864,22 @@ def _double_map_func(self):
     def _timestamp_map_func(self, column, col_type):
         datetime_default_size = 20  # size of 'YYYY-MM-DD HH:MM:SS.' (the datetime string up to the milliseconds)
         pattern = "%Y-%m-%d %H:%M:%S"
-        ms_size, ms_to_trim = self._get_number_of_digits(column)
+        ms_size, ms_to_trim, ms_div = self._get_number_of_digits(column)
         if ms_size > 0:
             pattern += ".%f"
 
-        dt_size = datetime_default_size + ms_size - ms_to_trim
+        dt_size = datetime_default_size
```
hashhar, line 871: replace size with length?

`dt_size`-> `timestamp_length`

trino/client.py
```diff
@@ -864,19 +864,22 @@ def _double_map_func(self):
...
         dt_tz_offset = datetime_default_size + ms_size
         if 'with time zone' in col_type:
-
             if ms_to_trim > 0:
                 return lambda val: \
-                    [datetime.strptime(val[:dt_size] + val[dt_tz_offset:], pattern + ' %z')
+                    [datetime.strptime(val[:dt_size]
+                     + str(round(int(val[dt_size:dt_tz_offset]) / ms_div))
```
hashhar, line 877: note: `round` returns floats so it'll lose precision and cause incorrect rounding in some cases.

tests/integration/test_types_integration.py
hashhar, line 217: Also squash "Round milliseconds instead of truncating when needed" into others.

trino/client.py
hashhar: I see you changed logic in later commits.

tests/integration/test_dbapi_integration.py
lpoulain: Done

trino/client.py
lpoulain: The round returns a float but the code only cares about the integer part. It takes the total number of digits after the decimal (e.g. `123456789`), divides it by `ms_div` to only keep 6 digits (e.g. `123456.789`), rounds this (`123457`) and converts it to a string. So to lose precision you would need to have a timestamp way more precise than 12 as `round(0.499999999999999999)` returns 0 and `round(0.5000000000000001)` return 1.

hashhar: That's not the issue I'm thinking of. Try `round(2.675, 2)`. It can happen for any other value since floats cannot represent all values exactly.

hashhar:
@lpoulain Please try using fixup commits when addressing review comments otherwise it's impossible to see what changed between the force push and before that.

tests/integration/test_types_integration.py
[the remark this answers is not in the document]
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Done

tests/integration/test_types_integration.py
lpoulain: Fixed

tests/integration/test_types_integration.py
lpoulain: Done

trino/client.py
lpoulain: Done

trino/client.py

trino/client.py

trino/client.py

trino/client.py
lpoulain: Fixed

tests/integration/test_types_integration.py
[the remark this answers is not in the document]
hashhar: reminder to remove `DATE '02001-08-22'` from test.

trino/client.py
hashhar: I still see the magic value of 20.

lpoulain: Ah you meant computing the value dynamically instead of a hard-coded value and an explanation in the comment. Done

tests/integration/test_types_integration.py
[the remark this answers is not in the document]
lpoulain: Done

mdesmet:
@lpoulain : can you rebase your changes on latest master?

trino/client.py
```diff
@@ -907,66 +914,95 @@ def _double_map_func(self):
...
         else:
-            tz = timedelta(hours=int(matches.group(3)), minutes=int(matches.group(4)))
-        return datetime.strptime(matches.group(1)[:time_size], pattern).time().replace(tzinfo=timezone(tz))
+            tz = timedelta(hours=int(matches.group('hours')), minutes=int(matches.group('minutes')))
+        time_str = matches.group('time')[:time_size]
+        millis_str = str(round(int(matches.group('time')[time_size:]) / ms_div))
+        return datetime.strptime(time_str + millis_str, pattern).time().replace(tzinfo=timezone(tz))
 
-    def _get_number_of_digits(self, column):
+    def _get_number_of_millis_digits(self, column):
```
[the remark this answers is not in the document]
lpoulain: Fixed

trino/client.py
```diff
@@ -907,66 +914,95 @@ def _double_map_func(self):
...
+        if matches.group('sign') == '-':
+            tz = -timedelta(hours=int(matches.group('hours')), minutes=int(matches.group('minutes')))
         else:
-            return lambda val: datetime.strptime(val[:time_size], pattern).time()
+            tz = timedelta(hours=int(matches.group('hours')), minutes=int(matches.group('minutes')))
+        return datetime.strptime(matches.group('time')[:time_size], pattern).time().replace(tzinfo=timezone(tz))
 
-    def _get_time_with_timezome(self, value, time_size, pattern):
-        matches = re.match(r'^(.*)([\+\-])(\d{2}):(\d{2})$', value)
+    def _get_time_with_timezone_round_ms(self, value, time_size, ms_div, pattern):
```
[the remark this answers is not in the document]
lpoulain: Fixed

trino/client.py
```diff
@@ -907,66 +914,95 @@ def _double_map_func(self):
...
-            return lambda val: datetime.strptime(val[:dt_size] + val[dt_tz_offset:], pattern)
+        if millis_div > 1:
+            return lambda val: datetime.strptime(val[:timestamp_length]
+                                                 + str(round(int(val[timestamp_length:dt_tz_offset]) / millis_div))
+                                                 + val[dt_tz_offset:], pattern)
         else:
             return lambda val: datetime.strptime(val, pattern)
 
     def _time_map_func(self, column, col_type):
+        datetime_default_size = 9  # size of 'HH:MM:SS.'
```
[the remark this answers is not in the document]
lpoulain: Fixed

tests/integration/test_types_integration.py
```diff
@@ -1,5 +1,8 @@
 import math
+import datetime
+from datetime import timedelta
```
hovaesco, line 3: Full datetime module already imported above

hovaesco (approved):

trino/client.py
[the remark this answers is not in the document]
mdesmet: Could you please have another look, it seems they are still all in `RowMapperFactory`?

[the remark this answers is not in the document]
hovaesco: Yeah, it makes sense to split them to separate classes.

tests/integration/test_types_integration.py
lpoulain: Fixed

trino/client.py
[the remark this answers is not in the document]
lpoulain: Sorry, I misread "classes" with "functions". Fixed for real this time.

trino/client.py
```diff
@@ -906,67 +913,138 @@ def _double_map_func(self):
...
+            tz = -timedelta(hours=int(matches.group('hours')), minutes=int(matches.group('minutes')))
+        else:
+            tz = timedelta(hours=int(matches.group('hours')), minutes=int(matches.group('minutes')))
+        time_str = matches.group('time')[:time_size]
+        millis_str = str(round(int(matches.group('time')[time_size:]) / millis_digits_to_trim))
+        return datetime.strptime(time_str + millis_str, time_format).time().replace(tzinfo=timezone(tz))
+
+
+class TimestampRowMapperFactory(AbstractTemporalRowMapperFactory):
+    def create(self, column, col_type):
```
mdesmet, line 999: Add types and return types in all methods

trino/client.py
```diff
@@ -906,67 +913,138 @@ def _double_map_func(self):
...
-            return lambda val: datetime.strptime(val[:dt_size] + val[dt_tz_offset:], pattern)
-        else:
-            return lambda val: datetime.strptime(val, pattern)
 
-    def _time_map_func(self, column, col_type):
-        pattern = "%H:%M:%S"
-        ms_size, ms_to_trim = self._get_number_of_digits(column)
-        if ms_size > 0:
-            pattern += ".%f"
+class TimeRowMapperFactory(AbstractTemporalRowMapperFactory):
```
mdesmet, line 934: This is actually a value mapper and not a row mapper. Please adjust the name

trino/client.py
lpoulain: Fixed

trino/client.py
lpoulain: Fixed

hashhar (commented):
skimmed once again.

To be honest the rounding code can use a lot of simplification. I'll propose a PR. In current shape I'm not confident at all in it without a lot of test coverage specific to rounding.

tests/integration/test_types_integration.py
```diff
@@ -200,6 +202,159 @@ def test_digest(trino_connection):
...
+        .add_field(sql="DATE '1582-10-05'", python=date(1582, 10, 5)) \
+        .add_field(sql="DATE '1582-10-14'", python=date(1582, 10, 14)) \
+        .execute()
+
+
+def test_time(trino_connection):
+    time_0 = time(1, 23, 45)
+    time_3 = time(1, 23, 45, 123000)
+    time_6 = time(1, 23, 45, 123456)
+    time_round = time(1, 23, 45, 123457)
```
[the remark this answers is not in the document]
lpoulain: Actually the rounding code does not work, and the only examples I could fine involve third party packages such as numpy. So open to suggestions for a better approach

trino/client.py
```diff
@@ -906,67 +914,143 @@ def _double_map_func(self):
...
-        pattern = "%H:%M:%S"
-        ms_size, ms_to_trim = self._get_number_of_digits(column)
-        if ms_size > 0:
-            pattern += ".%f"
+class TimeValueMapperFactory(AbstractTemporalValueMapperFactory):
+    def create(self, column: Dict[str, Any], col_type: str) -> Callable[[Any], tim]:
+        datetime_default_size = 9  # size of 'HH:MM:SS.'
+        time_format = "%H:%M:%S"
+        millis_length = self._get_number_of_millis_digits(column)
+        millis_to_trim_div = self._get_number_of_millis_digits_to_trim(column, millis_length)
```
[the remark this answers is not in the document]
lpoulain: It represents the number to divide by to remove the desired number of digits

tests/integration/test_types_integration.py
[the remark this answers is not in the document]
lpoulain: Is `23:59:59.999999949999` really supposed to be rounded down when converted to microseconds? Shouldn't 999999.949999 microseconds be rounded up?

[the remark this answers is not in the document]
hashhar: Sorry I added one more 9 in every case.

lpoulain:
Closing PR as too many things have changed since. Will file a new PR

PR #234, opened 2022-09-12
Add pull request template

hovaesco:
Almost 1:1 copy from Trino repository. It will help maintain release notes in the same way as in Trino project.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,21 @@
+<!-- Thank you for submitting a pull request! Find more information in our development guide at https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md and contact us on #python-client in Slack. -->
+<!-- Provide an overview of the PR for maintainers and reviewers. -->
+## Description
+
+
+
+<!-- Provide a user-friendly explanation, keep it brief if it isn't user-visible. -->
+## Non-technical explanation
+
+
+
+<!-- Mark the appropriate option with an (x). Propose a release note if you can. -->
+## Release notes
+
+( ) This is not user-visible and no release notes are required.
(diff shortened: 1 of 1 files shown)


hashhar:
@colebow PTAL as well.

.github/pull_request_template.md
```diff
@@ -0,0 +1,22 @@
...
+
+<!-- Mark the appropriate option with an (x). Propose a release note if you can. -->
+## Release notes
+
+( ) This is not user-visible and no release notes are required.
+( ) Release notes are required, please propose a release note for me.
+( ) Release notes are required, with the following suggested text:
+
+```markdown
+# Section
```
mosabua, line 20: This might not be needed in the python client since the release notes are probably just one list of items without sections/headers.

mosabua:
Should we also institute the same process of getting release notes assembled .. maybe even separate file .. or easier.. just an update to the changelog file that was proposed in #221

.github/pull_request_template.md
hashhar: SQLAlchemy, DB-API, General improvements? But yes - it feels overkill for now. We can add it back if needed.

hashhar:
> Should we also institute the same process of getting release notes assembled .. maybe even separate file .. or easier.. just an update to the changelog file that was proposed in #221

We can keep updating the proposed changelog in #221 for every release.

.github/pull_request_template.md
hovaesco: Should I remove a whole markdown section then?

hashhar: I think Manfred just means the `# Section` header. The markdown template is useful to have.

hovaesco: Done

PR #235, opened 2022-09-19
Escape url parameters in sqlalchemy connection strings

mdesmet:
Currently it is not possible to create a Trino url  in sqlalchemy urls containing characters like &, ' ' or @. The values should be properly encoded and decoded.

This PR provides a helper method that properly encodes the passed values and ensures that values are properly decoded before passing down as `dbapi.connect` parameters.

```
>>> URL(
...                         user="user@test.org/my_role",
...                         password="pass /*&",
...                         host="localhost",
...                         session_properties={"query_max_run_time": "1d"},
...                         http_headers={"trino": 1},
...                         extra_credential=[
...                             ("user1@test.org/my_role", "user2@test.org/my_role"),
...                             ("user3@test.org/my_role", "user36@test.org/my_role")],
...                         experimental_python_types=True,
...                         client_tags=["1 @& /\"", "sql"],
...                     )
'trino://user%40test.org%2Fmy_role:pass %2F*&@localhost:8080?source=trino-sqlalchemy&session_properties=%7B%22query_max_run_time%22%3A+%221d%22%7D&http_headers=%7B%22trino%22%3A+1%7D&extra_credential=%5B%28%27user1%40test.org%2Fmy_role%27%2C+%27user2%40test.org%2Fmy_role%27%29%2C+%28%27user3%40test.org%2Fmy_role%27%2C+%27user36%40test.org%2Fmy_role%27%29%5D&client_tags=%5B%221+%40%26+%2F%5C%22%22%2C+%22sql%22%5D&experimental_python_types=true'
```

## Release notes

```markdown
* Support sqlalchemy url generation through `trino.sqlalchemy.URL` helper method
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,9 +100,14 @@ Attributes can also be passed in the connection string.
 
 ```python
 from sqlalchemy import create_engine
+from trino.sqlalchemy import URL
 
 engine = create_engine(
-    'trino://user@localhost:8080/system',
+    URL(
+        host="localhost",
+        port=8080,
+        catalog="system"
+    ),
     connect_args={
       "session_properties": {'query_max_run_time': '1d'},
       "client_tags": ["tag1", "tag2"],
@@ -119,6 +124,14 @@ engine = create_engine(
     '&experimental_python_types=true'
     '&roles={"catalog1": "role1"}'
 )
+
+# or using the URL factory method
+engine = create_engine(URL(
+  host="localhost",
+  port=8080,
+  client_tags=["tag1", "tag2"],
+  experimental_python_types=True
+))
 ```
 
 ## Authentication mechanisms
--- a/tests/unit/sqlalchemy/test_dialect.py
+++ b/tests/unit/sqlalchemy/test_dialect.py
@@ -9,6 +9,7 @@ from trino.auth import BasicAuthentication
 from trino.dbapi import Connection
 from trino.sqlalchemy.dialect import CertificateAuthentication, JWTAuthentication, TrinoDialect
 from trino.transaction import IsolationLevel
+from trino.sqlalchemy import URL as trino_url
 
 
 class TestTrinoDialect:
@@ -16,20 +17,35 @@ class TestTrinoDialect:
         self.dialect = TrinoDialect()
 
     @pytest.mark.parametrize(
-        "url, expected_args, expected_kwargs",
+        "url, generated_url, expected_args, expected_kwargs",
         [
             (
-                make_url("trino://user@localhost"),
+                make_url(trino_url(
+                    user="user",
+                    host="localhost",
+                )),
+                'trino://user@localhost:8080?source=trino-sqlalchemy',
                 list(),
-                dict(host="localhost", catalog="system", user="user", source="trino-sqlalchemy"),
+                dict(host="localhost", catalog="system", user="user", port=8080, source="trino-sqlalchemy"),
             ),
             (
-                make_url("trino://user@localhost:8080"),
+                make_url(trino_url(
+                    user="user",
+                    host="localhost",
+                    port=443,
+                )),
+                'trino://user@localhost:443?source=trino-sqlalchemy',
                 list(),
-                dict(host="localhost", port=8080, catalog="system", user="user", source="trino-sqlalchemy"),
+                dict(host="localhost", port=443, catalog="system", user="user", source="trino-sqlalchemy"),
             ),
             (
-                make_url("trino://user:pass@localhost:8080?source=trino-rulez"),
+                make_url(trino_url(
+                    user="user",
+                    password="pass",
+                    host="localhost",
+                    source="trino-rulez",
+                )),
+                'trino://user:***@localhost:8080?source=trino-rulez',
                 list(),
                 dict(
                     host="localhost",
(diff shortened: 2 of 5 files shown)


trino/sqlalchemy/util.py
```diff
@@ -0,0 +1,90 @@
+import json
+from urllib.parse import quote_plus
+
+from typing import Optional, Dict, List, Union, Tuple
+from sqlalchemy import exc
+from sqlalchemy.engine.url import _rfc_1738_quote  # noqa
+
+
+def _url(
```
hovaesco, line 9: Some parameters are missing like `verify` or `redirect_handler`

tests/unit/sqlalchemy/test_dialect.py
```diff
@@ -19,17 +20,29 @@ def setup(self):
         "url, expected_args, expected_kwargs",
         [
             (
-                make_url("trino://user@localhost"),
+                make_url(trino_url(
```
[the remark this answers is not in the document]
mdesmet: Maybe we can simply assert that the string being generated is equal to the above string?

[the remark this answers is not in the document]
hovaesco: Sounds good.

trino/sqlalchemy/util.py
mdesmet: OAuth is not yet supported in the url. I think we can leave that for another PR. I have added verify

tests/unit/sqlalchemy/test_dialect.py
```diff
@@ -59,13 +84,66 @@ def setup(self):
...
+                '&session_properties=%7B%22query_max_run_time%22%3A+%221d%22%7D'
+                '&source=trino-sqlalchemy'
+                '&verify=false',
+                list(),
+                dict(
+                    host="localhost",
+                    port=8080,
+                    catalog="system",
+                    user="user@test.org/my_role",
+                    auth=BasicAuthentication("user@test.org/my_role", "pass /*&"),
```
hovaesco, line 123: Add test for other supported auth methods: JWT and Certificate.

trino/sqlalchemy/util.py
hovaesco: Sure, just please create a follow-up issue for it after merging, Kerberos doesn't seem supported as well, right?

mdesmet: Correct I supported only the fields that are currently parsed in the sqlalchemy connection string (and added `verify`).

tests/unit/sqlalchemy/test_dialect.py
mdesmet: Added both tests.

mdesmet:
@hovaesco: PTAL

hovaesco (approved):

mdesmet:
@hashhar, @ebyhr : Please have a look. This is important for Galaxy users to be able to use sqlalchemy.

README.md
```diff
@@ -72,8 +72,13 @@ NOTE: `password` and `schema` are optional
...
 from sqlalchemy.schema import Table, MetaData
 from sqlalchemy.sql.expression import select, text
+from trino.sqlalchemy import URL
 
-engine = create_engine('trino://user@localhost:8080/system')
+engine = create_engine(URL(
+  host="localhost",
+  port=8080,
+  catalog="system"
+))
```
hashhar, line 81: I'd suggest to keep a "simple" example using the normal URL form.

The utility of the new way is better shown by the `connect_args` example below.

trino/sqlalchemy/dialect.py
```diff
@@ -80,49 +81,52 @@ def create_connect_args(self, url: URL) -> Tuple[Sequence[Any], Mapping[str, Any
 
         db_parts = (url.database or "system").split("/")
         if len(db_parts) == 1:
-            kwargs["catalog"] = db_parts[0]
+            kwargs["catalog"] = unquote_plus(db_parts[0])
```
hashhar, line 84: some things are `unquote` others `unquote_plus`? Why?

mdesmet: Should be unquote_plus everywhere.

mdesmet:
@hashhar : PTAL

hashhar (approved):

trino/sqlalchemy/dialect.py
```diff
@@ -80,49 +81,52 @@ def create_connect_args(self, url: URL) -> Tuple[Sequence[Any], Mapping[str, Any
...
-            kwargs["session_properties"] = json.loads(url.query["session_properties"])
+            kwargs["session_properties"] = json.loads(unquote_plus(url.query["session_properties"]))
 
         if "http_headers" in url.query:
-            kwargs["http_headers"] = json.loads(url.query["http_headers"])
+            kwargs["http_headers"] = json.loads(unquote_plus(url.query["http_headers"]))
 
         if "extra_credential" in url.query:
-            kwargs["extra_credential"] = literal_eval(url.query["extra_credential"])
+            kwargs["extra_credential"] = literal_eval(unquote_plus(url.query["extra_credential"]))
```
hashhar, line 120: unsafe to do literal_eval on user input. Might not be a concern for the client itself but any tool using the client then becomes impacted.

trino/sqlalchemy/util.py
```diff
@@ -0,0 +1,94 @@
...
+    trino_url += f"?source={quote_plus(source)}"
+
+    if session_properties is not None:
+        trino_url += f"&session_properties={quote_plus(json.dumps(session_properties))}"
+
+    if http_headers is not None:
+        trino_url += f"&http_headers={quote_plus(json.dumps(http_headers))}"
+
+    if extra_credential is not None:
+        trino_url += f"&extra_credential={quote_plus(repr(extra_credential))}"
```
hashhar, line 74: using `repr` for serialization seems like a shortcut.

trino/sqlalchemy/dialect.py
mdesmet: Note that literal_eval only takes into account literals. So it is not actually evaluating arbitrary code only structural and literal types.

```
def literal_eval(node_or_string):
    """
    Safely evaluate an expression node or a string containing a Python
    expression.  The string or node provided may only consist of the following
    Python literal structures: strings, bytes, numbers, tuples, lists, dicts,
    sets, booleans, and None.
    """
```

trino/sqlalchemy/util.py
mdesmet: `json.dumps` converts tuples into array. that's why we can't use it there.

trino/sqlalchemy/dialect.py
hashhar: It's safe from executing arbitrary code but I can still crash the Python VM with my input. Try `"1 + 100"*1000000` as input.

My point is there are alternative ways to serialize this and we should use them instead of doing the easy thing.

trino/sqlalchemy/util.py
hashhar: I mean we can write our own value class over this list of tuple, serialize the value class, deserialize value class and provide a method on value class to convert back into list of tuple.

mdesmet:
@hashhar : PTAL. I have not added another class as it is a simple one liner using the typical json.loads recipe but looping through it and instantiating a tuple based on the array.

```
kwargs["extra_credential"] = [
    tuple(extra_credential) for extra_credential in json.loads(unquote_plus(url.query["extra_credential"]))
]
```

hashhar (approved):

PR #242, opened 2022-09-30
Fix prepared statement handling

hashhar:
## Description

The prepared statement handling code assumed that for each query we'll
always receive some non-empty response even after the initial response
which is not a valid assumption.

This assumption worked because earlier Trino used to send empty fake
results even for queries which don't return results (like PREPARE and
DEALLOCATE) but is now invalid with
https://github.com/trinodb/trino/commit/bc794cd49a616fa95eecfcc384b761f67176240b.

The other problem with the code was that it leaked HTTP protocol details
into dbapi.py and worked around it by keeping a deep copy of the request
object from the PREPARE execution and re-using it for the actual query
execution.

The new code fixes both issues by processing the prepared statement
headers as they are received and storing the resulting set of active
prepared statements on the ClientSession object. The ClientSession's set
of prepared statements is then rendered into the prepared statement
request header in TrinoRequest. Since the ClientSession is created and
reused for the entire Connection this also means that we can now
actually implement re-use of prepared statements within a single
Connection.

## Non-technical explanation

Fix errors when using prepared statements with Trino >= 398.

## Release notes

```markdown
* Fix errors when using prepared statements with Trino servers newer than 398.
```

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -881,21 +881,6 @@ class FakeGatewayResponse(object):
         return http_response
 
 
-def test_trino_result_response_headers():
-    """
-    Validates that the `TrinoResult.response_headers` property returns the
-    headers associated to the TrinoQuery instance provided to the `TrinoResult`
-    class.
-    """
-    mock_trino_query = mock.Mock(respone_headers={
-        'X-Trino-Fake-1': 'one',
-        'X-Trino-Fake-2': 'two',
-    })
-
-    result = TrinoResult(query=mock_trino_query, rows=[])
-    assert result.response_headers == mock_trino_query.response_headers
-
-
 def test_trino_query_response_headers(sample_get_response_data):
     """
     Validates that the `TrinoQuery.execute` function can take addtional headers
--- a/trino/client.py
+++ b/trino/client.py
@@ -125,6 +125,7 @@ class ClientSession(object):
         self._extra_credential = extra_credential
         self._client_tags = client_tags
         self._role = role
+        self._prepared_statements: Dict[str, str] = {}
         self._object_lock = threading.Lock()
 
     @property
@@ -206,6 +207,15 @@ class ClientSession(object):
         with self._object_lock:
             self._role = role
 
+    @property
+    def prepared_statements(self):
+        return self._prepared_statements
+
+    @prepared_statements.setter
+    def prepared_statements(self, prepared_statements):
+        with self._object_lock:
+            self._prepared_statements = prepared_statements
+
 
 def get_header_values(headers, header):
     return [val.strip() for val in headers[header].split(",")]
@@ -215,6 +225,14 @@ def get_session_property_values(headers, header):
     kvs = get_header_values(headers, header)
     return [
         (k.strip(), urllib.parse.unquote(v.strip()))
+        for k, v in (kv.split("=", 1) for kv in kvs)
+    ]
+
+
+def get_prepared_statement_values(headers, header):
+    kvs = get_header_values(headers, header)
+    return [
+        (k.strip(), urllib.parse.unquote_plus(v.strip()))
         for k, v in (kv.split("=", 1) for kv in kvs)
     ]
 
@@ -392,6 +410,13 @@ class TrinoRequest(object):
             for name, value in self._client_session.properties.items()
         )
 
+        if len(self._client_session.prepared_statements) != 0:
+            # ``name`` must not contain ``=``
+            headers[constants.HEADER_PREPARED_STATEMENT] = ",".join(
+                "{}={}".format(name, urllib.parse.quote_plus(statement))
+                for name, statement in self._client_session.prepared_statements.items()
+            )
+
         # merge custom http headers
         for key in self._client_session.headers:
             if key in headers.keys():
@@ -556,6 +581,18 @@ class TrinoRequest(object):
         if constants.HEADER_SET_ROLE in http_response.headers:
             self._client_session.role = http_response.headers[constants.HEADER_SET_ROLE]
 
+        if constants.HEADER_ADDED_PREPARE in http_response.headers:
+            for name, statement in get_prepared_statement_values(
+                http_response.headers, constants.HEADER_ADDED_PREPARE
+            ):
+                self._client_session.prepared_statements[name] = statement
+
+        if constants.HEADER_DEALLOCATED_PREPARE in http_response.headers:
+            for name in get_header_values(
+                http_response.headers, constants.HEADER_DEALLOCATED_PREPARE
+            ):
+                self._client_session.prepared_statements.pop(name)
+
         self._next_uri = response.get("nextUri")
 
         return TrinoStatus(
@@ -622,10 +659,6 @@ class TrinoResult(object):
 
             self._rows = next_rows
 
-    @property
-    def response_headers(self):
-        return self._query.response_headers
-
 
 class TrinoQuery(object):
     """Represent the execution of a SQL statement by Trino."""
@@ -648,7 +681,6 @@ class TrinoQuery(object):
         self._update_type = None
         self._sql = sql
         self._result: Optional[TrinoResult] = None
-        self._response_headers = None
         self._experimental_python_types = experimental_python_types
         self._row_mapper: Optional[RowMapper] = None
 
@@ -705,7 +737,7 @@ class TrinoQuery(object):
         rows = self._row_mapper.map(status.rows) if self._row_mapper else status.rows
         self._result = TrinoResult(self, rows)
 
-        # Execute should block until at least one row is received
+        # Execute should block until at least one row is received or query is finished or cancelled
         while not self.finished and not self.cancelled and len(self._result.rows) == 0:
             self._result.rows += self.fetch()
         return self._result
@@ -725,7 +757,6 @@ class TrinoQuery(object):
         status = self._request.process(response)
         self._update_state(status)
         logger.debug(status)
-        self._response_headers = response.headers
         if status.next_uri is None:
             self._finished = True
 
@@ -762,10 +793,6 @@ class TrinoQuery(object):
     @property
     def cancelled(self) -> bool:
         return self._cancelled
-
-    @property
-    def response_headers(self):
-        return self._response_headers
 
 
 def _retry_with(handle_retry, handled_exceptions, conditions, max_attempts):
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -20,7 +20,6 @@ decide to convert then to a list of tuples.
 from decimal import Decimal
 from typing import Any, List, Optional  # NOQA for mypy types
 
-import copy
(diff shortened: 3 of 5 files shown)


hashhar:
FYI @mdesmet - note that this is incomplete code - the commit for now just cleans up the HTTP portion (client.py).

dbapi.py changes I'm still working on - but created this to receive some early feedback.
In dbapi.py with the new changes I think I don't need to do anything at all other than just executing the `PREPARE`, `EXECUTE` and `DEALLOCATE` since the headers will all be persisted as long as we use same Cursor object.

In a future PR I plan to add additional parameter if user wants prepared statements to be re-used.

trino/dbapi.py
```diff
@@ -301,52 +301,35 @@ def setinputsizes(self, sizes):
...
 
-    def _prepare_statement(self, operation, statement_name):
-        """
-        Prepends the given `operation` with "PREPARE <statement_name> FROM" and
-        executes as a prepare statement.
-
-        :param operation: sql to be executed.
-        :param statement_name: name that will be assigned to the prepare
-            statement.
+    def _new_request_with_session_from(self, request):
```
hashhar, line 304: This is better than `copy.deepcopy` but I still hate it.

I think the way we try to make prepared statement work needs improvement.
Currently it's modelled as:

- A `Connection` creates a `ClientSession` - makes total sense - all persistent information for a connection is maintained using the session (roles, prepared statements, session properties, transactions)
- A `Cursor` is created from `Connection` which either gets it's own `TrinoRequest` or re-uses one from an existing `transaction`. The `TrinoRequest` inherits a reference to the `Connection`'s `ClientSession`. This doesn't make too much sense. It assumes a Cursor will only execute a query at a time (reasonable) but also leaks HTTP protocol information to the `Cursor`.
- `Cursor.execute(sql, params)` is called which creates a `TrinoQuery` using the `Cursor`'s `TrinoRequest`.
- Any operation on the cursor then affects the `TrinoRequest` - which means we should not share `TrinoRequest` objects.

These are the current rules, we break them very badly:

- In `Cursor.execute(sql, params)` if params are provided we need to execute `PREPARE statement FROM sql`
- To execute something we need a `TrinoRequest` - so we re-use the one from current Cursor - bad bad because any other future query or running query would get affected
- We workaround this by deepcopy. This copies too much and means that the `ClientSession` is no longer shared.
- We then workaround this by creating new request but re-using the `ClientSession` from old one (to persist headers etc.) What if we miss other things that need to be re-used? Things will break.
- We then execute our PREPARE with copied request, EXECUTE using original request and then DEALLOCATE using copied request.

I'll create an issue with some pseudocode class outline mentioning what refactoring I want to do to improve this - we can discuss other possible improvements there and improve what is honestly a very organically grown code which does not have enough structure to safeguard us from doing mistakes or introducing changes with confidence that we only affect what we changed.

tests/unit/test_client.py
```diff
@@ -881,21 +881,6 @@ def __call__(self, *args, **kwargs):
         return http_response
 
 
-def test_trino_result_response_headers():
```
hashhar, line 884: this is redundant now since the actual user-facing functionality is that `additional_http_headers` works by merging passed headers to existing headers which is tested in `test_trino_query_response_headers` already.

The use of this test was to ensure that headers were travelling between `TrinoQuery` and `TrinoResult` since older prepared statement code basically passed-through the prepare headers through all layers - which was very simple solution but unclean and couples things together (and also buggy as Trino 398 shows).

trino/client.py
```diff
@@ -218,6 +228,12 @@ def get_session_property_values(headers, header):
         for k, v in (kv.split("=", 1) for kv in kvs)
     ]
 
+def get_prepared_statement_values(headers, header):
```
hashhar, line 231: this is copy of the method above this with quoting method replaced. This will be unified as follow-up.

Note that there are quoting bugs pre-exisiting for session properties for example - I'll add tests and fix them in separate PR.

trino/client.py
```diff
@@ -392,6 +408,13 @@ def http_headers(self) -> Dict[str, str]:
             for name, value in self._client_session.properties.items()
         )
 
+        if len(self._client_session.prepared_statements) != 0:
+            # ``name`` must not contain ``=``
```
hashhar, line 412: again a copy - proper assertion will be added + quoting/unquoting of name as well since that is what Trino server does.

trino/dbapi.py
```diff
@@ -295,58 +295,42 @@ def warnings(self):
             return self._query.warnings
         return None
 
+    def _new_request_with_session_from(self, request):
```
hashhar, line 298: This is better than `copy.deepcopy` but I still hate it.

I think the way we try to make prepared statement work needs improvement.
Currently it's modelled as:

- A `Connection` creates a `ClientSession` - makes total sense - all persistent information for a connection is maintained using the session (roles, prepared statements, session properties, transactions)
- A `Cursor` is created from `Connection` which either gets it's own `TrinoRequest` or re-uses one from an existing `transaction`. The `TrinoRequest` inherits a reference to the `Connection`'s `ClientSession`. This doesn't make too much sense. It assumes a Cursor will only execute a query at a time (reasonable) but also leaks HTTP protocol information to the `Cursor`.
- `Cursor.execute(sql, params)` is called which creates a `TrinoQuery` using the `Cursor`'s `TrinoRequest`.
- Any operation on the cursor then affects the `TrinoRequest` - which means we should not share `TrinoRequest` objects.

These are the current rules, we break them very badly:

- In `Cursor.execute(sql, params)` if params are provided we need to execute `PREPARE statement FROM sql`
- To execute something we need a `TrinoRequest` - so we re-use the one from current Cursor - bad bad because any other future query or running query would get affected
- We workaround this by deepcopy. This copies too much and means that the `ClientSession` is no longer shared.
- We then workaround this by creating new request but re-using the `ClientSession` from old one (to persist headers etc.) What if we miss other things that need to be re-used? Things will break.
- We then execute our PREPARE with copied request, EXECUTE using original request and then DEALLOCATE using copied request.

I'll create an issue with some pseudocode class outline mentioning what refactoring I want to do to improve this - we can discuss other possible improvements there and improve what is honestly a very organically grown code which does not have enough structure to safeguard us from doing mistakes or introducing changes with confidence that we only affect what we changed.

hashhar:
BTW the existing code was buggy even before the Trino change because it didn't get to inspect the initial response from the coordinator because that response's headers were not being set to `TrinoResult` because the setting of response headers was handled in `fetch` which only sees the 2nd response and onward.

trino/client.py
```diff
@@ -556,6 +581,18 @@ def process(self, http_response) -> TrinoStatus:
...
+                http_response.headers, constants.HEADER_ADDED_PREPARE
+            ):
+                self._client_session.prepared_statements[name] = statement
+
+        if constants.HEADER_DEALLOCATED_PREPARE in http_response.headers:
+            for name in get_header_values(
+                http_response.headers, constants.HEADER_DEALLOCATED_PREPARE
+            ):
+                self._client_session.prepared_statements.pop(name)
+
```
hovaesco, line 595: Can you please add tests for setting proper headers in ClientSession?

trino/client.py
hovaesco: In the follow-up we can add more meaningful var names. I understand key-values nomenclature but would prefer something explicit.

trino/dbapi.py
hovaesco: > A Cursor is created from Connection which either gets it's own TrinoRequest or re-uses one from an existing transaction

Does it mean that for each cursor object a new `TrinoRequest` is created? Should it be somehow reused, you mentioned that is re-uses one from an existing transaction but where it's handled?

mdesmet: Why not simply call `self.connection._create_request()`? 

The ClientSession is already shared and the prepared statement headers are also persisted in the `ClientSession`.

hashhar: > Does it mean that for each cursor object a new TrinoRequest is created? Should it be somehow reused, you mentioned that is re-uses one from an existing transaction but where it's handled?

Creating new TrinoRequest per cursor makes sense since the TrinoRequest tracks query state (nextUri, stats etc.) - it must not be re-used across queries where possible. Transactions re-use it here - https://github.com/trinodb/trino-python-client/blob/efb6680873f8ed2a7493101e637e7c6d04633013/trino/dbapi.py#L212-L215 (but it's safe since even with transactions a Cursor can only be executing one query at a time).

hashhar: > Why not simply call self.connection._create_request()?
>
> The ClientSession is already shared and the prepared statement headers are also persisted in the ClientSession.

Good catch. Much simpler. Let me try if that works (it should).

trino/client.py
hashhar: ClientSession doesn't store header values - it stores information read from headers.

Do you mean a test that a request which includes ADDED header actually affects the `prepared_statements` dict and so on? Isn't that covered by the prepared statements tests themselves?

trino/client.py
hashhar: Agreed. Plan is to have a helper module/module functions to assist with reading:

- value of a header by name
- values of a "collection" header
- key-value from a map-type "collection" header

Current we have one such function for each header (roles, prepared statements, session properties).

hashhar:
applied comments @mdesmet @hovaesco . Please take a look again.

trino/client.py
hovaesco: Yes, you are right. But I don't think those tests are there. I can only see `test_trino_query_response_headers` in `test_client.py`.

hashhar: They live in test_dbapi_integration.py and test_sqlalchemy_integration.py. See the various `_params` test methods.

I can add some tests verifying that TrinoRequest#http_headers returns correct headers when prepared statements exist in ClientSession but testing the "implementation" is generally not any more useful than testing the "feature".

Let me know what you'd prefer.

hovaesco: It makes more sense to test the "feature" in this case.  

Disscused offline to add a test to assert that after each query the `prepared_statements` dict is empty (to make sure we don't leak things)

ebyhr (approved):

hashhar:
@ebyhr @hovaesco Added an explicit test for prepared statement + also verified that we don't leak prepared statements in the ClientSession.

hovaesco (approved):

mdesmet:
> Python: analysis.........................................................Failed
> - hook id: flake8
> - exit code: 1
> 
> tests/integration/test_dbapi_integration.py:1071:1: E302 expected 2 blank lines, found 1
> 
> Python: types............................................................Passed
> Error: Process completed with exit code 1.

That's why I quite like the pre-commit checks to be present as git hooks. Note that if you haven't explicitly installed the pre-commit git hook (running `pre-commit install`, it wouldn't run on commit. Maybe we should consider activating this? It would save us some CI minutes and a failing build.

cc: @hovaesco, @hashhar

PR #247, opened 2022-10-05
Make defensive copy of reference types in `ClientSession`

mdesmet:
If multiple connections are created from same passed parameters (eg session properties, client tags, roles, ...), changes in one connection's client_session should not affect other connections.

## Release notes

```markdown
```


ebyhr (approved):

ebyhr:
Merged, thanks!

PR #249, opened 2022-10-06
Ignore already removed deallocated prepared statement headers

### LINKED ISSUE #248: KeyError raised on prepared statement deallocate
### Expected behavior

When executing a prepared statement, the statement shall be executed succesfully

### Actual behavior

Multipe responses from Trino contain the same deallocate header, although the code only accepts this operation to be executed once.

See https://github.com/trinodb/trino-python-client/blob/f97aea6afb2663f0a29beb6477c18c9d9a88832e/trino/client.py#L610

### Steps To Reproduce

I wasn't able to reproduce this locally, only in CI. So I assume some type of race condition is happening.

### Log output

```
140581443462912: before: {}
140581443462912: after: {"st_ccb622831a2d43dfb03a186180695c1e": "INSERT INTO delta.test16650588382631922575_test_incremental_delete_insert.seed (\"state\", \"county\", \"city\", \"last_visit_date\")\n VALUES \n  ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n"}
140581443462912: Deallocating st_ccb622831a2d43dfb03a186180695c1e
140581443462912: before: {"st_ccb622831a2d43dfb03a186180695c1e": "INSERT INTO delta.test16650588382631922575_test_incremental_delete_insert.seed (\"state\", \"county\", \"city\", \"last_visit_date\")\n VALUES \n  ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n, ROW (?, ?, ?, ?)\n"}
140581443462912http://localhost:8080/v1/statement/executing/20[2210](https://github.com/mdesmet/dbt-trino/actions/runs/3197070204/jobs/5219814756#step:4:2211)06_122055_01035_2z8ej/ye8e084beb54f7dd32c5b3d682239264dbe6a6262/0 : <trino.client.TrinoRequest object at 0x7fdbd40a4b50> : st_ccb622831a2d43dfb03a186180695c1e
140581443462912http://localhost:8080/v1/statement/executing/20221006_122055_01035_2z8ej/y382a838b78d9dcd5df455e0fd1335d282cd7f1a2/1 : <trino.client.TrinoRequest object at 0x7fdbd40a4b50> : st_ccb622831a2d43dfb03a186180695c1e
Error while running:
insert into delta.test16650588382631922575_test_incremental_delete_insert.seed ("state", "county", "city", "last_visit_date") values
          (?,?,?,?),(?,?,?,?),(?,?,?,?),(?,?,?,?),(?,?,?,?),(?,?,?,?)
'st_ccb622831a2d43dfb03a186180695c1e'
```

### Operating System

Ubuntu

0.317.0

### Trino Server version

398

### Python version

3.10.5

mdesmet:
## Description

Resolves #248 

`X-Trino-Deallocated-Prepare` may appear on multiple responses. Query execution should not fail on prepared statement removal if the prepared statement has already been removed.

## Release notes

```markdown
* Fix failure on already removed deallocated prepared statement headers
```

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -607,7 +607,7 @@ class TrinoRequest(object):
             for name in get_header_values(
                 http_response.headers, constants.HEADER_DEALLOCATED_PREPARE
             ):
-                self._client_session.prepared_statements.pop(name)
+                self._client_session.prepared_statements.pop(name, None)
 
         self._next_uri = response.get("nextUri")
 


hashhar (approved):
LGTM. @findepi is the server supposed to be doing this?

trino/client.py
```diff
@@ -607,7 +607,12 @@ def process(self, http_response) -> TrinoStatus:
             for name in get_header_values(
                 http_response.headers, constants.HEADER_DEALLOCATED_PREPARE
             ):
-                self._client_session.prepared_statements.pop(name)
+                # Multiple responses may contain the same deallocated statement name,
+                # so we can safely ignore `KeyError` if the same statement is already removed from the client session.
+                try:
+                    self._client_session.prepared_statements.pop(name)
+                except KeyError:
```
findepi, line 614: avoid KeyError -- https://stackoverflow.com/a/15411146

findepi:
> is the server supposed to be doing this?

i don't think it is.

from the issue

> Steps To Reproduce
> I wasn't able to reproduce this locally, only in CI. So I assume some type of race condition is happening.

this is concerning a bit. the client protocol is "single threaded", i.e. there is typically exactly one ongoing request.

trino/client.py
mdesmet: Done!

hashhar: Is there some way to know whether this results in a single lookup or two - multiple lookups will cause races when same connection has two cursors both of which are deallocating statements (highly unlikely since our statement names (and hence dict keys) are effectively unique).

trino/client.py
```diff
@@ -607,7 +607,7 @@ def process(self, http_response) -> TrinoStatus:
             for name in get_header_values(
                 http_response.headers, constants.HEADER_DEALLOCATED_PREPARE
             ):
-                self._client_session.prepared_statements.pop(name)
+                self._client_session.prepared_statements.pop(name, None)
```
hashhar, line 610: Can we add a debug log or a `warnings.warn` here since ideally this should not be happenning at all. We should try to collect when this happens and why and fix it.

hashhar: Useful information to log would be the statement we are trying to deallocate, the current state of the prepared_statements dict and the response headers we received.

mdesmet:
@hashhar : An alternative solution would be to implement the deallocated prepared statements as a `set` instead and merge them when performing a request (as in the JDBC driver). The way it's done in JDBC makes me think that this behaviour might actually not be erroneous but I'm still investigating if this is the case or not.

findepi (approved):

Issue #27, opened 2020-05-04
The python client seems to ignore system timezone unlike Trino CLI


RoboticHuman:
I'm using at_timezone in my query but it seems that the session timezone is passed incorrectly to PrestoServer. Adding a local offset in the python script does solve the issue but that's not a solution

findepi:
@RoboticHuman that's correct, it's not implemented.

You can see in `constants.py` which headers are supported by the client.

https://github.com/prestosql/presto-python-client/blob/f920129faa26f5e886710ed5c2a2cbbfd182ef1c/presto/constants.py#L32-L43

-- `X-Presto-Time-Zone` is not one of them.

For the record, here is the list of headers supported by the server

https://github.com/prestosql/presto/blob/fa02de3ba715b6533610f148d9bb45ce44ae8703/presto-client/src/main/java/io/prestosql/client/PrestoHeaders.java#L16-L44

RoboticHuman:
Thanks for your fast reply @findepi , any expectation of when this can be fixed? I can try to implement it on my end and do a PR if that works too.

RoboticHuman:
For now I just passed X-Presto-Time-Zone in the http_headers in the Connection. I guess that's a good enough solution till it's properly implemented in case anyone falls for this as well

findepi:
Challenge is how do we provide the default. The header should be filled with IANA zone id.
The caveat here is that I don't know how to get one in Python standard lib (in Python 2.7 and 3.x)
https://www.python.org/dev/peps/pep-0615/ looks like the future.

@RoboticHuman do you happen to know?

RoboticHuman:
Yeah I only noticed this problem when I referred to how things are implemented in the Java side, I think your suggestion is right as it needs to fall back properly.

Isn't pytz a good interface for IANA? or you think it's better to wait for the standard lib integration?

findepi:
I have nothing against pytz, I am more concerned about pulling in a dependency. (The more dependencies we have, more problems for the users.)
We may need to adda depency though...

FWIW pyhive uses `python-dateutil` (https://pypi.org/project/python-dateutil/) which also does the job.

ebyhr:
`pytz` is required module from #168.

cc: @hashhar @mdesmet

mdesmet:
I actually looked into this a few weeks ago: what is missing from `pytz`, is the ability to retrieve the local timezone. We would need to either replace pytz or add in another dependency.

[...]

PR #252, opened 2022-10-07
Add support for TIMEZONE

ulisesojeda:
Add support for TIMEZONE

Fixes https://github.com/trinodb/trino-python-client/issues/27

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,6 +362,24 @@ conn = trino.dbapi.connect(
 )
 ```
 
+## Timezone
+
+The time zone for the session can be explicitly set using the IANA time zone
+name. When not set the time zone defaults to the client side local timezone.
+
+```python
+import trino
+conn = trino.dbapi.connect(
+    host='localhost',
+    port=443,
+    user='username',
+    timezone='Europe/Brussels',
+)
+```
+
+> **NOTE: The behaviour till version 0.320.0 was the same as setting session timezone to UTC.**
+> **To preserve that behaviour pass `timezone='UTC'` when creating the connection.**
+
 ## SSL
 
 ### SSL verification
--- a/setup.py
+++ b/setup.py
@@ -79,8 +79,8 @@ setup(
         "Programming Language :: Python :: Implementation :: PyPy",
         "Topic :: Database :: Front-Ends",
     ],
-    python_requires=">=3.7",
-    install_requires=["pytz", "requests"],
+    python_requires='>=3.7',
+    install_requires=["pytz", "requests", "tzlocal"],
     extras_require={
         "all": all_require,
         "kerberos": kerberos_require,
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -16,6 +16,7 @@ from decimal import Decimal
 import pytest
 import pytz
 import requests
+from tzlocal import get_localzone_name  # type: ignore
 
 import trino
 from tests.integration.conftest import trino_version
@@ -1107,3 +1108,31 @@ def test_prepared_statements(run_trino):
     cur.execute('DEALLOCATE PREPARE test_prepared_statements')
     cur.fetchall()
     assert cur._request._client_session.prepared_statements == {}
+
+
+def test_set_timezone_in_connection(run_trino):
+    _, host, port = run_trino
+
+    trino_connection = trino.dbapi.Connection(
+        host=host, port=port, user="test", catalog="tpch", timezone="Europe/Brussels"
+    )
+    cur = trino_connection.cursor()
+    cur.execute('SELECT current_timezone()')
+    res = cur.fetchall()
+    assert res[0][0] == "Europe/Brussels"
+
+
+def test_connection_without_timezone(run_trino):
+    _, host, port = run_trino
+
+    trino_connection = trino.dbapi.Connection(
+        host=host, port=port, user="test", catalog="tpch"
+    )
+    cur = trino_connection.cursor()
+    cur.execute('SELECT current_timezone()')
+    res = cur.fetchall()
+    session_tz = res[0][0]
+    localzone = get_localzone_name()
+    assert session_tz == localzone or \
+        (session_tz == "UTC" and localzone == "Etc/UTC") \
+        # Workaround for difference between Trino timezone and tzlocal for UTC
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -22,6 +22,7 @@ import pytest
 import requests
 from httpretty import httprettified
 from requests_kerberos.exceptions import KerberosExchangeError
+from tzlocal import get_localzone_name  # type: ignore
 
 import trino.exceptions
 from tests.unit.oauth_test_utils import (
@@ -47,6 +48,11 @@ from trino.client import (
     _retry_with,
     _RetryWithExponentialBackoff,
 )
+
+try:
+    from zoneinfo import ZoneInfoNotFoundError  # type: ignore
+except ModuleNotFoundError:
+    from backports.zoneinfo._common import ZoneInfoNotFoundError  # type: ignore
 
 
 @mock.patch("trino.client.TrinoRequest.http")
@@ -81,6 +87,7 @@ def test_request_headers(mock_get_and_post):
     schema = "test_schema"
     user = "test_user"
     source = "test_source"
+    timezone = "Europe/Brussels"
     accept_encoding_header = "accept-encoding"
     accept_encoding_value = "identity,deflate,gzip"
     client_info_header = constants.HEADER_CLIENT_INFO
@@ -94,6 +101,7 @@ def test_request_headers(mock_get_and_post):
             source=source,
             catalog=catalog,
             schema=schema,
+            timezone=timezone,
             headers={
                 accept_encoding_header: accept_encoding_value,
                 client_info_header: client_info_value,
@@ -109,9 +117,10 @@ def test_request_headers(mock_get_and_post):
         assert headers[constants.HEADER_SOURCE] == source
         assert headers[constants.HEADER_USER] == user
         assert headers[constants.HEADER_SESSION] == ""
+        assert headers[constants.HEADER_TIMEZONE] == timezone
         assert headers[accept_encoding_header] == accept_encoding_value
         assert headers[client_info_header] == client_info_value
-        assert len(headers.keys()) == 8
+        assert len(headers.keys()) == 9
 
     req.post("URL")
     _, post_kwargs = post.call_args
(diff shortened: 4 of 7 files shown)


tests/integration/test_dbapi_integration.py
```diff
@@ -1120,3 +1121,27 @@ def test_prepared_statements(run_trino):
...
+def test_set_timezone_in_connection(run_trino):
+    _, host, port = run_trino
+
+    trino_connection = trino.dbapi.Connection(
+        host=host, port=port, user="test", catalog="tpch", timezone="Europe/Brussels"
+    )
+    cur = trino_connection.cursor()
+    cur.execute('SHOW TABLES FROM information_schema')
+    cur.fetchall()
+    assert cur._request.http_headers[constants.HEADER_TIMEZONE] == "Europe/Brussels"
```
mdesmet, line 1135: instead of the request header testing, please check using `SELECT current_timezone()`

tests/integration/test_dbapi_integration.py
```diff
@@ -1120,3 +1121,27 @@ def test_prepared_statements(run_trino):
...
+def test_connection_without_timezone(run_trino):
+    _, host, port = run_trino
+
+    trino_connection = trino.dbapi.Connection(
+        host=host, port=port, user="test", catalog="tpch"
+    )
+    cur = trino_connection.cursor()
+    cur.execute('SHOW TABLES FROM information_schema')
+    cur.fetchall()
+    assert cur._request.http_headers[constants.HEADER_TIMEZONE] == get_localzone_name()
```
mdesmet, line 1147: Same as in `test_set_timezone_in_connection`.

tests/unit/test_client.py
```diff
@@ -1056,3 +1060,51 @@ def test_request_headers_role_empty(mock_get_and_post):
     req.get("URL")
     _, get_kwargs = get.call_args
     assert_headers_with_roles(post_kwargs["headers"], None)
+
+
+def assert_headers_timezone(headers: Dict[str, str], timezone: Optional[str]):
```
mdesmet, line 1065: Make second parameter required instead of optional and move `get_localzone_name()` call to the test itself. Makes the test easier to follow.

trino/client.py
```diff
@@ -127,6 +131,7 @@ def __init__(
         self._roles = roles.copy() if roles is not None else {}
         self._prepared_statements: Dict[str, str] = {}
         self._object_lock = threading.Lock()
+        self._timezone = timezone or get_localzone_name()
```
mdesmet, line 134: This might be a breaking change, In the sense that suddenly the local timezone will be used compared to UTC.

I'll leave this to the maintainers to decide if we need to put it behind a feature flag or announce it as a breaking change.

setup.py
```diff
@@ -39,6 +39,7 @@
     "pytest",
     "pytest-runner",
     "click",
+    "tzlocal",
```
mdesmet, line 42: This can be removed as it is already included in the `install_requires`.

mdesmet (commented):
Thanks for this contribution, overall looks very good.

Note that this also should be documented.

tests/unit/test_client.py
ulisesojeda: Sure.

trino/client.py
ulisesojeda: Ok, how should we proceed?

ulisesojeda:
> Thanks for this contribution, overall looks very good.
> 
> Note that this also should be documented.

Done the doc. Thanks for your review and comments @mdesmet !

trino/client.py
mdesmet, line 134: is there a way to validate the time zone using tzlocal?

README.md
```diff
@@ -357,6 +357,20 @@ conn = trino.dbapi.connect(
 )
 ```
 
+## Timezone
+
+Sets the time zone for the session using the time zone name. Defaults to the timezone set on your workstation.
```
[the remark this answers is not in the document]
ulisesojeda: :+1: Thanks

tests/integration/test_dbapi_integration.py
```diff
@@ -1120,3 +1121,31 @@ def test_prepared_statements(run_trino):
...
+    trino_connection = trino.dbapi.Connection(
+        host=host, port=port, user="test", catalog="tpch"
+    )
+    cur = trino_connection.cursor()
+    cur.execute('SELECT current_timezone()')
+    res = cur.fetchall()
+    session_tz = res[0][0]
+    localzone = get_localzone_name()
+    assert session_tz == localzone or \
+        (session_tz == "UTC" and localzone == "Etc/UTC") \
```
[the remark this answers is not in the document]
ulisesojeda: Exactly

trino/client.py
ulisesojeda: Good point. I added the check with ZoneInfo

mdesmet:
@ulisesojeda : Please don't merge master into your branch but rebase on master.

ulisesojeda:
@ebyhr @hashhar could you please check it?

trino/client.py
hashhar: Q: Before this change the server returned values in UTC? If yes, then won't setting `timezone=UTC` here explicitly preserve old behaviour? If yes then it means a workaround exists to preserve old behaviour and we can mention it in release notes.

Note that in larger picture this makes Python client match JDBC client etc. So I'd consider it a bugfix which happens to be breaking change.

hashhar (commented):
Thanks for implementing this @ulisesojeda. I've started reviewing this and sorry about the delays here.

tests/unit/test_client.py
```diff
@@ -47,6 +48,12 @@
     _RetryWithExponentialBackoff,
 )
 
+try:
+    from zoneinfo._common import ZoneInfoNotFoundError  # type: ignore
```
hashhar, line 52: Can directly import `zoneinfo.ZoneInfoNotFoundError` - that's the public location.

trino/client.py
```diff
@@ -127,6 +138,9 @@ def __init__(
         self._roles = roles.copy() if roles is not None else {}
         self._prepared_statements: Dict[str, str] = {}
         self._object_lock = threading.Lock()
+        if timezone:  # Check timezone validity
+            ZoneInfo(timezone)
```
hashhar, line 142: Just for other reviewers this throws on invalid timezones (invalidity is determined by the OS's tzdb files or the tzdata package shipped by Python itself):

```
ZoneInfoNotFoundError: 'No time zone found with key Foobar'
```

trino/client.py
```diff
@@ -127,6 +138,9 @@ def __init__(
...
+        self._timezone = timezone or get_localzone_name()
```
hashhar, line 143: `get_localzone_name` can also return invalid zone names.

IMO the `ZoneInfo` check should be moved after this line as `ZoneInfo(self._timezone)`

hashhar (commented):
Nice, thanks for a lot for adding this support.

Left some comments.

trino/client.py
ulisesojeda: The timezone returned by Trino CLI is the local timezone.

README.md
```diff
@@ -362,6 +362,20 @@ conn = trino.dbapi.connect(
 )
 ```
 
+## Timezone
+
+The time zone for the session can be explicitly set using the official IANA time zone name. When not set the time zone defaults to the client side local timezone.
```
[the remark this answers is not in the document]
ulisesojeda: Good point. Yes, it keep the old behaviour.

trino/client.py
```diff
@@ -100,6 +108,8 @@ class ClientSession(object):
     :param client_tags: Client tags as list of strings.
     :param roles: roles for the current session. Some connectors do not
                  support role management. See connector documentation for more details.
+    :param timezone: The timezone for query processing. Defaults to the timezone
+                      of the Trino cluster, and not the timezone of the client.
```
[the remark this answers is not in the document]
ulisesojeda: it's a mistake. Should be 'Defaults to the timezone set on your workstation.'

tests/integration/test_dbapi_integration.py
```diff
@@ -16,6 +16,7 @@
 import pytest
 import pytz
 import requests
+from tzlocal import get_localzone_name  # type: ignore
```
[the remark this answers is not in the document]
ulisesojeda: Could you please provide a snippet using `astimezone` to get name/offset timezone?

tests/unit/test_client.py
ulisesojeda: as per https://pypi.org/project/backports.zoneinfo/, with need to import the backport for Python <3.9. Did I miss something?

trino/client.py
ulisesojeda: Good catch. Thanks

tests/unit/test_client.py
hashhar: Oh, good point. I was checking with 3.10 only.

Can you add a TODO comment that it can be simplified once we move past 3.9?

tests/integration/test_dbapi_integration.py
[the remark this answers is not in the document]
hashhar: Numeric offsets also work with Trino. You are correct that we cannot use `astimezone` to get named offset.

However after thinking more about it it's possible for the numeric offset to cause incorrect results in a long running application because the offset will be determined once and if after few mins/hours DST change happens then the numeric offset will be incorrect while named offset should work correctly.

So leave it as it is. Thanks.

tests/unit/test_client.py
hashhar: Sorry I was saying something like this:

```python
try:
    from zoneinfo import ZoneInfoNotFoundError
except ModuleNotFoundError:
    from backports.zoneinfo._common import ZoneInfoNotFoundError
```

.pre-commit-config.yaml
```diff
@@ -13,9 +13,4 @@ repos:
         additional_dependencies:
           - "types-pytz"
           - "types-requests"
-
```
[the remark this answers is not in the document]
hashhar: good catch. I didn't notice this. 🙈

ulisesojeda:
> LGTM % some comments.
> 
> Thanks for this.
> 
> I plan to not merge this for 0.320.0 so that other features can be part of 0.320.0 which people might be waiting for. Then 0.321.0 can introduce this change. I hope that's ok?
> 
> While there's a way to preserve old beahviour it requires someone to read release notes so I'd like to announce on Trino community before this change.

Sure. Thanks for the review @hashhar

hashhar:
Rebased. Will merge once CI is done.

PR #254, opened 2022-10-10
Add sqlalchemy test matrix to ensure compatibility with 1.3 version

### LINKED ISSUE #250: Trino 0.317 broke compatibility with SQLAlchemy 1.3.24
### Expected behavior

Calling `_get_server_version_info` works with SQLAlchemy 1.3.24.

### Actual behavior

Calling `_get_server_version_info` throws an AttributeError

```
    def _get_server_version_info(self, connection: Connection) -> Any:
        query = "SELECT version()"
        try:
>           version = res.scalar_one()
E           AttributeError: 'ResultProxy' object has no attribute 'scalar_one'

venv/lib/python3.10/site-packages/trino/sqlalchemy/dialect.py:337: AttributeError
```

### Steps To Reproduce

I believe 7c97873af5a3bf390d8e4d75721b6c60e1a085e9 is the culprit here.

### Log output

see above

### Operating System

ubuntu in github actions

0.317

### Trino Server version

n/a

### Python version

repro'd on both 3.7 and 3.10

mdesmet:
Fixes #250 

## Description

`scalar_one()` and `scalar_one_or_none()` are not available in sqlalchemy 1.3. They have been replaced by `scalar()` calls.

sqlalchemy has also been added into the ci matrix to ensure compatibility with 1.3 and 1.4 versions.

## Release notes

```markdown
* Use sqlalchemy 1.3 compatible methods in sqlalchemy dialect.
```


.github/workflows/ci.yml
```diff
@@ -47,11 +47,16 @@ jobs:
         trino: [
           "latest",
         ]
+        sqlalchemy: [
+          "1.4"
```
hovaesco, line 51: How about testing it always with the latest `1.3.x` and `1.4.x` version? It could help identify some issues earlier. Upstream client tools like superset use the latest patch.

.github/workflows/ci.yml
```diff
@@ -47,11 +47,16 @@ jobs:
...
+        ]
         include:
           # Test with older Trino versions for backward compatibility
-          - { python: "3.10", trino: "351" }  # first Trino version
+          - { python: "3.10", trino: "351", sqlalchemy: "1.4" }  # first Trino version
           # Test with Trino version that requires result set to be fully exhausted
-          - { python: "3.10", trino: "395" }
+          - { python: "3.10", trino: "395", sqlalchemy: "1.4" }
+          # Test with sqlalchemy 1.3
+          - { python: "3.10", trino: "395", sqlalchemy: "1.3" }
```
hashhar, line 59: We can probably remove this now since the actual issue was fixed in the client itself (you can even say the client did not follow the protocol correctly).

(Do it as separate commit).

.github/workflows/ci.yml
hashhar: I don't see why matrix has `1.4`? Shouldn't it have `1.3` instead since setup.py currently installs latest `1.x`.

mdesmet: > I don't see why matrix has 1.4? Shouldn't it have 1.3 instead since setup.py currently installs latest 1.x.

I added one extra include with latest 1.3. similar like we always test with latest trino except for some additional includes.

mdesmet: > How about testing it always with the latest `1.3.x` and `1.4.x` version?

Fixed in latest commit

.github/workflows/ci.yml
mdesmet: Done

.github/workflows/ci.yml
```diff
@@ -63,7 +68,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install libkrb5-dev
-          pip install .[tests]
+          pip install .[tests] sqlalchemy~=${{ matrix.sqlalchemy }}
```
hashhar, line 69: I'd instead say to inline the constraint in the matrix entry itself. 

i.e. `~=1.4.0` (latest 1.4)
`==1.4.2` (some specific version)

tests/unit/sqlalchemy/test_compiler.py
```diff
@@ -45,24 +46,40 @@ def dialect():
...
 def test_offset(dialect):
     statement = select(table_without_catalog).offset(0)
     query = statement.compile(dialect=dialect)
     assert str(query) == 'SELECT "table".id, "table".name \nFROM "table"\nOFFSET :param_1'
 
 
+@pytest.mark.skipif(
+    sqlalchemy_version() < "1.4",
+    reason="columns argument to select() must be a Python list or other iterable"
+)
```
hashhar, line 82: we're basically dropping lot of coverage. Seems like to make sure we can reasonably rely on the matrix we'd need to create two test classes or modify tests to work with both.

tests/unit/sqlalchemy/test_dialect.py
```diff
@@ -196,14 +202,14 @@ def setup(self):
                     source="trino-sqlalchemy",
                 ),
             ),
-        ],
+        ] if sqlalchemy_version() >= "1.4" else [],  # make_url is not supported in sqlalchemy 1.3
```
hashhar, line 205: do docs need updating as well that the `URL` static helper can only be used with 1.4.x?

hashhar (commented):
seems to be an unfortunate situation since we're adding a matrix but also removing most of the tests.

How do other clients provide support for multiple SQLAlchemy versions?

tests/unit/sqlalchemy/test_compiler.py
mdesmet: Agreed but I suggest we do that as a follow up issue. Most test cases are basically sqlalchemy feature tests. In this fix we mainly lay the foundation (the test matrix) and fix the issue at hand: #250

.github/workflows/ci.yml
mdesmet: Makes sense!

tests/unit/sqlalchemy/test_dialect.py
mdesmet: Actually I fixed it now by adding a slash after the port. So now the helper is working for 1.3 and 1.4.

hovaesco:
Let's create a compatibility section under https://github.com/trinodb/trino-python-client#sqlalchemy and document which SQLAlchemy versions are supported.

mdesmet:
> Let's create a compatibility section under https://github.com/trinodb/trino-python-client#sqlalchemy and document which SQLAlchemy versions are supported.

Let's handle this as follow up.

mdesmet:
@hashhar, @hovaesco: My point is that both 1.3 and 1.4 are supported but there will be some rough edges which will only surfaced by adding more integration tests.

I have added the compatibility section though.

hashhar (approved):

john-bodley:
@mdesmet what's wrong with simply adding a dependency constraint that only SQLAlchemy >= 1.4 is supported? This adheres to the KISS principle and doesn't pollute the CI build matrix etc.

hashhar:
Because we can't control what other third party tools use as the version of SQLAlchemy. And SQLAlchemy 1.4.x is still relatively new with not as much adoption AFAIK.

I'd love to be able to force a version of SQLAlchemy on the consumer to simply our lives but it's not possible today.

Issue #262, opened 2022-10-12
get_prepared_statement_values fails if HEADER_ADDED_PREPARE is present in header but empty


derek-pyne:
### Expected behavior

Moving from 0.316 to 0.317 causes Trino queries to (run with pandas read_sql) to fail inside get_prepared_statement_values (https://github.com/trinodb/trino-python-client/blob/68f3d9d843e307c57b6ce728493a9837939ba60a/trino/client.py#L232) with a Value Error (`ValueError: not enough values to unpack (expected 2, got 1)`)

This seems to be caused by this [commit](https://github.com/trinodb/trino-python-client/commit/d5f779bc1c2ddab270f367bab30ecdcaffc1baf6#diff-cf9b945b8cb541d38680c39caa2de6568793a6d53644d6f2659de6a77ff96127) which checks for the HEADER_ADDED_PREPARE and parses the data in it.

However, in our case, our Trino setup is always returning this header ('X-Trino-Added-Prepare'), just with an empty string if there is no data in it. `get_prepared_statement_values` then fails trying to parse the data in an empty string.

One fix, would just to not only check that this header is present (which is what is currently done), but also check that there is a non-empty value in it.

Note: I have very little understanding of any of the guts here and only a surface level understanding of a Trino setup. If the feedback is 'Your Trino setup should not be returning headers with empty values' then I can get some more information about why thats the case on my side :) 

### Actual behavior

Value Error if `X-Trino-Added-Prepare` is present but empty

### Steps To Reproduce

Add `X-Trino-Added-Prepare` to response with empty string. Unsure if this is expected behaviour (can get more information from my side if this is weird)

### Operating System

macOS 12.6

### Trino Python client version

0.317

### Trino Server version

388.007

### Python version

3.9.9

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

hashhar:
> Trino Server version
> 
> 388.007

Doesn't look like a real Trino version. Are you running a fork with some modifications?

hashhar:
BTW even if it's not expected we should still make the code more defensive either way (for example if there's some proxy sitting between Trino and the client which mangles the headers) - so please send a PR.

hashhar:
Fixed in https://github.com/trinodb/trino-python-client/pull/263

PR #263, opened 2022-10-13
Checking for config headers to be both present and non-empty

derek-pyne:
Addresses #262 .

Handles cases where these config headers are present in the request, but with an empty string as their value. Currently, an error is raised when trying to parse the empty strings. This PR addresses this by checking that these headers are not only present, but not empty.

These headers being present with empty values is not normally expected, although we'd like to protect against it for cases like a proxy sitting in between Trino and the client that is altering the headers (which is the case for me).

Notes: 
- Added checks for all headers in `process` that might hit this issue because the values are parsed. HEADER_SET_SESSION, HEADER_SET_ROLE and HEADER_ADDED_PREPARE are currently failing to parse if the header value is empty
- The test reproduces the error, but I don't fully understand the mocking that is happening. Looking for guidance if there is a different preferred way of testing this

## Non-technical explanation
Protecting against headers being present but without data

## Release notes

```markdown
* Fix value error with present but empty headers. ([#262](https://github.com/trinodb/trino-python-client/issues/262))
```

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -225,7 +225,7 @@ def get_session_property_values(headers, header):
     kvs = get_header_values(headers, header)
     return [
         (k.strip(), urllib.parse.unquote_plus(v.strip()))
-        for k, v in (kv.split("=", 1) for kv in kvs)
+        for k, v in (kv.split("=", 1) for kv in kvs if kv)
     ]
 
 
@@ -233,7 +233,7 @@ def get_prepared_statement_values(headers, header):
     kvs = get_header_values(headers, header)
     return [
         (k.strip(), urllib.parse.unquote_plus(v.strip()))
-        for k, v in (kv.split("=", 1) for kv in kvs)
+        for k, v in (kv.split("=", 1) for kv in kvs if kv)
     ]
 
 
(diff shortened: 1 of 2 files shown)


trino/client.py
```diff
@@ -579,25 +579,25 @@ def process(self, http_response) -> TrinoStatus:
             ):
                 self._client_session.properties.pop(prop, None)
 
-        if constants.HEADER_SET_SESSION in http_response.headers:
+        if http_response.headers.get(constants.HEADER_SET_SESSION):
```
hashhar, line 582: I'd instead fix the various `get_session_property_values` and `get_roles_values` etc.

That way you can even unit-test those methods directly instead of using mocks.

It also prevents against regressions when those functions get re-used in other places.

derek-pyne: Yah, sounds much better to me :)

Change to this approach is in

derek-pyne:
Morning! Just checking in if this is a change we'd like to include, or something we want to discuss more :)

hashhar:
@derek-pyne I've been travelling so did not get a chance to re-review/merge. I'll get to this starting the weekend/Monday.

hashhar:
@cla-bot check

derek-pyne:
> @derek-pyne I've been travelling so did not get a chance to re-review/merge. I'll get to this starting the weekend/Monday.
> 
> Sorry about the long wait.

All good! No rush, just checking in

hashhar:
Squashed the commits. Will merge once CI is done. Thanks for the fix.

Issue #264, opened 2022-10-13
Send extra client_tags with cursor.execute


benrifkind:
### Describe the feature

Currently you can only specify client_tags when you create a Trino Connection. I want to be able to update the tags whenever I run cursor.execute. My workflow is generally to create one cursor from one connection. And then use cursor.execute multiple times. Each query that I send with execute is different and so I want to tag it differently for tracking purposes. 

### Describe alternatives you've considered

One alternative is just to add `additional_http_headers` to the `Cursor.execute` method and pass it through to the `trino.client.TrinoQuery.execute` method. Then the user can update the tags by setting the header key "X-Trino-Client-Tags"

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

benrifkind:
It looks like client_tags was was initially implemented on a per query basis but then it was changed. So probably I'm missing something.
https://github.com/trinodb/trino-python-client/pull/154/files/9dc1e697f2141c1fc89083654b599015679bc456#r795281894

benrifkind:
Seems like the suggested thing to do is just to create a connection per query.

https://github.com/trinodb/trino-python-client/pull/265#pullrequestreview-1141127795

PR #265, opened 2022-10-13
Pass additional_headers to cursor.execute

benrifkind:
## Description

This adds the ability to specify `additional_http_headers` in `Cursor.execute`. 

## Non-technical explanation

Adds the ability to specify additional HTTP headers when calling Cursor.execute. Since this is already available in `trino.client.TrinoQuery.execute` it is just a matter of exposing it as an argument in Cursor.execute.

This also unblocks me for #264 without making a change that it appears was already decided against. In short, this enables me to add tags on a per query basis even though Trino does not technically support it. See https://github.com/trinodb/trino-python-client/pull/154/files/9dc1e697f2141c1fc89083654b599015679bc456#r795281894

## Release notes

```markdown
* Expose additional_headers in Cursor.execute. #264 
```

### DIFF
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -405,7 +405,7 @@ class Cursor(object):
     def _generate_unique_statement_name(self):
         return 'st_' + uuid.uuid4().hex.replace('-', '')
 
-    def execute(self, operation, params=None):
+    def execute(self, operation, params=None, additional_http_headers=None):
         if params:
             assert isinstance(params, (list, tuple)), (
                 'params must be a list or tuple containing the query '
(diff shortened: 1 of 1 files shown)


hashhar (changes requested):
What is the actual issue you're trying to solve? There might be other ways to structure it and we can work together to improve the client for more general use-cases as well.

Anything which changes session per-`execute` call would be decided against IMO.
These are already configurable on the `Connection`/`Cursor` and in Trino connection/cursors don't cost anything so you can create as many of them as needed.

This makes the code harder to reason about and leaks details across layers which makes future changes harder and easier for bugs to happen (see https://github.com/trinodb/trino-python-client/pull/242 for example).

benrifkind:
For sure. The actual issue I am trying to solve is #264. I want to be able to specify tags on a per query basis. Another way I could do this is just create a new connection for each query but that seems like overkill to me.

hashhar:
I'm creating a issue to see if there are things other than `client_tags` which a user might want to set on per-query basis.
If we can come up with a good rationale for that then we can think about how to make this change in a way that's not limited to client_tags and not as open-ended as `additional_headers`.

Yes - for your use-case I'd suggest using new Connection/Cursor per query should be fine.

Thanks for linking the use-case you had in mind.

PR #266, opened 2022-10-13
[sqlalchemy] Ensure TrinoDialect.get_table_names only returns real tables

john-bodley:
## Description

Per the [SQLAlchemy documentation](https://docs.sqlalchemy.org/en/14/core/reflection.html#sqlalchemy.engine.reflection.Inspector.get_table_names) the `get_table_names` method is intended to return the real table names, i.e., without views. The PR updates the `get_table_names` to only return real table names whereas previously it returned real tables and views.

Note there were no prior integration tests for the `TrinoDialect.get_view_names` method so I added some to cover this logic and threw in a few extra for free which should cover all the various scenarios.

## Release notes

> The `TrinoDialect.get_table_names` function logic has changed to adhere to the SQLAlchemy specification. Now only real table names are returned as opposed to both real table and view names.

```markdown
* Return only tables in SQLAlchemy  `get_table_names`. Previously, it contained views. ({issue}`266`)
```

### DIFF
--- a/tests/integration/test_sqlalchemy_integration.py
+++ b/tests/integration/test_sqlalchemy_integration.py
@@ -12,6 +12,7 @@
 import pytest
 import sqlalchemy as sqla
 from sqlalchemy.sql import and_, or_, not_
+from sqlalchemy_utils import create_view
 
 from tests.unit.conftest import sqlalchemy_version
 from trino.sqlalchemy.datatype import JSON
@@ -368,3 +369,43 @@ def test_get_table_comment(trino_connection):
         assert actual['text'] is None
     finally:
         metadata.drop_all(engine)
+
+
+@pytest.mark.parametrize('trino_connection', ['memory/test'], indirect=True)
+@pytest.mark.parametrize('schema', [None, 'test'])
+def test_get_table_names(trino_connection, schema):
+    engine, conn = trino_connection
+    name = schema or engine.dialect._get_default_schema_name(conn)
+    metadata = sqla.MetaData(schema=name)
+
+    if not engine.dialect.has_schema(conn, name):
+        engine.execute(sqla.schema.CreateSchema(name))
+
+    try:
+        create_view(
+            'my_view',
+            sqla.select(
+                [
+                    sqla.Table(
+                        'my_table',
+                        metadata,
+                        sqla.Column('id', sqla.Integer),
+                    ),
+                ],
+            ),
+            metadata,
+            cascade_on_drop=False,
+        )
+
+        metadata.create_all(engine)
+        assert sqla.inspect(engine).get_table_names(schema) == ['my_table']
+    finally:
+        metadata.drop_all(engine)
+
+
+@pytest.mark.parametrize('trino_connection', ['memory'], indirect=True)
+def test_get_table_names_raises(trino_connection):
+    engine, _ = trino_connection
+
+    with pytest.raises(sqla.exc.NoSuchTableError):
+        sqla.inspect(engine).get_table_names(None)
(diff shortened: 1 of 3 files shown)


trino/sqlalchemy/dialect.py
```diff
@@ -216,8 +217,9 @@ def get_view_names(self, connection: Connection, schema: str = None, **kw) -> Li
         query = dedent(
             """
             SELECT "table_name"
-            FROM "information_schema"."views"
+            FROM "information_schema"."tables"
             WHERE "table_schema" = :schema
+              AND "table_type" = 'VIEW'
```
hashhar, line 222: Please add a test covering this (probably in `test_sqlalchemy_integration`).

You can create a VIEW in the memory connector and verify that it's not present in output of `get_table_names` and is present in `get_view_names`.

hashhar (approved):
Thank you.

hashhar:
Also commit message to say:

```
Ensure SQLAlchemy get_table_names only returns real tables

Per the [SQLAlchemy
documentation](https://docs.sqlalchemy.org/en/14/core/reflection.html#sqlalchemy.engine.reflection.Inspector.get_table_names)
the get_table_names method is intended to return the real table names,
i.e., exclude views.

Furthermore querying the INFORMATION_SCHEMAS.VIEWS tables seems to be
sub-performant for determining which entities are views—I speculate this
is because all the view definitions need to be extracted. A more
performant approach is to use INFORMATION_SCHEMAS.TABLES with the
appropriate type filter.
```

trino/sqlalchemy/dialect.py
john-bodley: @hashhar are you aware that there's no current unit/integration tests for either the `get_table_names` and `get_view_names` methods?

hashhar: yes I do (unfortunately). SQLAlchemy integration made it in because it was useful but there hadn't been much time invested into it (until very recently).

tests/integration/test_sqlalchemy_integration.py
```diff
@@ -20,8 +23,16 @@
 @pytest.fixture
 def trino_connection(run_trino, request):
     _, host, port = run_trino
-    engine = sqla.create_engine(f"trino://test@{host}:{port}/{request.param}",
-                                connect_args={"source": "test", "max_attempts": 1})
+
+    engine = sqla.create_engine(
```
john-bodley, line 27: I needed to explicitly set the schema as part of the engine otherwise Trino barfed when trying to create the views with the `Schema must be specified when session schema is not set` error.

john-bodley:
@hashhar I've addressed your comments. I addition to adding an integration test for this specific change I added additional tests which should cover the gamut of conditions for the `TrinoDialect.get_table_names` method.

tests/integration/test_sqlalchemy_integration.py
```diff
@@ -368,3 +379,45 @@ def test_get_table_comment(trino_connection):
...
+            cascade_on_drop=False,
+        )
+
+        metadata.create_all(engine)
+        assert sqla.inspect(engine).get_table_names(schema) == ['my_table']
+    finally:
+        metadata.drop_all(engine)
+
+
+@patch('trino.sqlalchemy.dialect.TrinoDialect._get_default_schema_name')
```
mdesmet, line 416: Is this patch really necessary? We only set the catalog in the connection string, so I would expect the schema to be already `None`.

john-bodley: @mdesmet per [here](https://github.com/trinodb/trino-python-client/blob/master/trino/sqlalchemy/dialect.py#L195) if the schema is `None` it tries to fallback to the default schema. This patch ensures that on line #195 `schema` is `None` which means the exception will be invoked.

john-bodley:
@mdesmet thanks for the review. I've addressed your comments.

hovaesco (approved):

ebyhr (approved):

hashhar (commented):
Sorry about post-merge review.

It's a future-looking comment - useful to address but not required today.

PR #267, opened 2022-10-13
[sqlalchemy] Improve the performance of TrinoDialect.get_view_names

john-bodley:
## Description

Querying the `information_schema.views` table seems to be sub-performant for determining which entities are views—I speculate this is because all the view definitions need to be extracted/compiled. A more performant approach is to use `information_schema.tables` with the appropriate type filter. 

For example for a very large schema at Airbnb (comprising of over 100k entities) the following query,

```sql
SELECT
    table_name
FROM 
    information_schema.tables
WHERE 
    table_schema = '<schema>' AND 
    table_type = 'VIEW'
```

took ~ 5 seconds, whereas:

```sql
SELECT
    table_name
FROM
    information_schema.views
WHERE    
    table_schema = '<schema>'
```

was still running after 10 minutes.

Note there were no prior integration tests for the `TrinoDialect.get_view_names` method so I added some to cover this logic and threw in a few extra for free which should cover all the various scenarios.

## Release notes

```markdown
* Improve the performance of get_view_names in SQLAlchemy. ({issue}`267`)
```

### DIFF
--- a/tests/integration/test_sqlalchemy_integration.py
+++ b/tests/integration/test_sqlalchemy_integration.py
@@ -409,3 +409,43 @@ def test_get_table_names_raises(trino_connection):
 
     with pytest.raises(sqla.exc.NoSuchTableError):
         sqla.inspect(engine).get_table_names(None)
+
+
+@pytest.mark.parametrize('trino_connection', ['memory/test'], indirect=True)
+@pytest.mark.parametrize('schema', [None, 'test'])
+def test_get_view_names(trino_connection, schema):
+    engine, conn = trino_connection
+    name = schema or engine.dialect._get_default_schema_name(conn)
+    metadata = sqla.MetaData(schema=name)
+
+    if not engine.dialect.has_schema(conn, name):
+        engine.execute(sqla.schema.CreateSchema(name))
+
+    try:
+        create_view(
+            'test_get_view_names',
+            sqla.select(
+                [
+                    sqla.Table(
+                        'my_table',
+                        metadata,
+                        sqla.Column('id', sqla.Integer),
+                    ),
+                ],
+            ),
+            metadata,
+            cascade_on_drop=False,
+        )
+
+        metadata.create_all(engine)
+        assert sqla.inspect(engine).get_view_names(schema) == ['test_get_view_names']
+    finally:
+        metadata.drop_all(engine)
+
+
+@pytest.mark.parametrize('trino_connection', ['memory'], indirect=True)
+def test_get_view_names_raises(trino_connection):
+    engine, _ = trino_connection
+
+    with pytest.raises(sqla.exc.NoSuchTableError):
+        sqla.inspect(engine).get_view_names(None)
--- a/trino/sqlalchemy/dialect.py
+++ b/trino/sqlalchemy/dialect.py
@@ -215,11 +215,14 @@ class TrinoDialect(DefaultDialect):
         schema = schema or self._get_default_schema_name(connection)
         if schema is None:
             raise exc.NoSuchTableError("schema is required")
+
+        # Querying the information_schema.views table is subpar as it compiles the view definitions.
         query = dedent(
             """
             SELECT "table_name"
-            FROM "information_schema"."views"
+            FROM "information_schema"."tables"
             WHERE "table_schema" = :schema
+              AND "table_type" = 'VIEW'
         """
         ).strip()
         res = connection.execute(sql.text(query), schema=schema)


tests/integration/test_sqlalchemy_integration.py
```diff
@@ -20,8 +23,16 @@
...
 def trino_connection(run_trino, request):
     _, host, port = run_trino
-    engine = sqla.create_engine(f"trino://test@{host}:{port}/{request.param}",
-                                connect_args={"source": "test", "max_attempts": 1})
+
+    engine = sqla.create_engine(
+        f"trino://test@{host}:{port}/{request.param}",
+        connect_args={
+            "max_attempts": 1,
+            "schema": "test",
```
john-bodley, line 31: I needed to explicitly set the schema as part of the engine otherwise Trino barfed when trying to create the views with the `Schema must be specified when session schema is not set` error.

mdesmet: On first sight it can be done by doing something like:
```
create_view(
            ...,
            schema="test")
```
See https://github.com/kvesteri/sqlalchemy-utils/blob/1f6e2d6c029f5c578be0e48c7baa98e6984b7bf8/sqlalchemy_utils/view.py#L63

I also think you could include the schema in the pytest param `@pytest.mark.parametrize('trino_connection', ['memory/test'], indirect=True)` but personally I would prefer to have it in the `create_view` call.

john-bodley: @mdesmet the [create_view](https://github.com/kvesteri/sqlalchemy-utils/blob/1f6e2d6c029f5c578be0e48c7baa98e6984b7bf8/sqlalchemy_utils/view.py#L122-L127) method does not support setting an explicit schema. Regrettably also defining, 

```
metadata = sqla.MetaData(schema=schema)
``` 

did not work and thus the only solution I was able to find—I spent hours trying to get this to work—was to explicitly set the schema at the time of the engine creation.

tests/integration/test_sqlalchemy_integration.py
```diff
@@ -368,3 +379,45 @@ def test_get_table_comment(trino_connection):
         assert actual['text'] is None
     finally:
         metadata.drop_all(engine)
+
+
+@pytest.mark.parametrize('trino_connection', ['memory'], indirect=True)
```
mdesmet, line 384: Instead of overriding the schema for any test you can just do 

```suggestion
@pytest.mark.parametrize('trino_connection', ['memory/test'], indirect=True)
```

That way you can get rid of the `@patch` in the next test as there won't be a default scheme set.

tests/integration/test_sqlalchemy_integration.py
```diff
@@ -368,3 +379,45 @@ def test_get_table_comment(trino_connection):
...
+            cascade_on_drop=False,
+        )
+
+        metadata.create_all(engine)
+        assert sqla.inspect(engine).get_view_names(schema) == ['my_view']
+    finally:
+        metadata.drop_all(engine)
+
+
+@patch('trino.sqlalchemy.dialect.TrinoDialect._get_default_schema_name')
```
mdesmet, line 416: If you remove the default schema from the fixture you can rid yourself of this `@patch`

john-bodley:
@mdesmet I've addressed your comments.

mdesmet (approved):

mdesmet:
@hashhar : PTAL I think we should make this change based on your feedback on Slack

> hashhar
>  For views we call metadata.getViews in InformationSchemaPageSource - that methods ends up retrieving view definitions as well.
> For tables we call listTables + listViews (in InformationSchemaPageSource) - neither of which need view definitions. So actually the tables is faster since it just lists things.

ebyhr (approved):
Please fix the commit title as "Improve the performance of get_view_names in SQLAlchemy"

trino/sqlalchemy/dialect.py
```diff
@@ -216,8 +216,9 @@ def get_view_names(self, connection: Connection, schema: str = None, **kw) -> Li
         query = dedent(
             """
             SELECT "table_name"
-            FROM "information_schema"."views"
+            FROM "information_schema"."tables"
             WHERE "table_schema" = :schema
+              AND "table_type" = 'VIEW'
```
ebyhr, line 221: Worth leaving code comment why this query uses `information_schema.tables`.

setup.py
```diff
@@ -39,6 +39,7 @@
     "pytest",
     "pytest-runner",
     "click",
+    "sqlalchemy-utils",
```
hashhar, line 42: note that we're now starting to test integration with a non-SQLAlchemy library. This is one of many other alternative libraries which enhance SQLAlchemy.

PR #284, opened 2022-11-09
Fix refresh tokens flow

sinkuladis:
## Description

Fixes refresh token issue caused by `Error: header info didn't have x_redirect_server`. Trino coordinator with enabled refresh token after access token expiration will respond with only `x_token_server` field in authentication header, if refresh token has expired then coordinator restarts the flow. 

## Non-technical explanation

Fix refresh flow in case Trino cluster has refresh tokens enabled. 

## Release notes

```markdown
```

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -32,6 +32,7 @@ from tests.unit.oauth_test_utils import (
     MultithreadedTokenServer,
     PostStatementCallback,
     RedirectHandler,
+    RedirectHandlerWithException,
     _get_token_requests,
     _post_statement_requests,
 )
@@ -384,6 +385,48 @@ def test_oauth2_authentication_flow(attempts, sample_post_response_data):
     assert len(_get_token_requests(challenge_id)) == attempts
 
 
+@httprettified
+def test_oauth2_refresh_token_flow(sample_post_response_data):
+    token = str(uuid.uuid4())
+    challenge_id = str(uuid.uuid4())
+
+    token_server = f"{TOKEN_RESOURCE}/{challenge_id}"
+
+    post_statement_callback = PostStatementCallback(None, token_server, [token], sample_post_response_data)
+
+    # bind post statement
+    httpretty.register_uri(
+        method=httpretty.POST,
+        uri=f"{SERVER_ADDRESS}{constants.URL_STATEMENT_PATH}",
+        body=post_statement_callback)
+
+    # bind get token
+    get_token_callback = GetTokenCallback(token_server, token)
+    httpretty.register_uri(
+        method=httpretty.GET,
+        uri=token_server,
+        body=get_token_callback)
+
+    redirect_handler = RedirectHandlerWithException(
+        trino.exceptions.TrinoAuthError(
+            "Do not use redirect handler when there is no redirect_uri in the response"))
+
+    request = TrinoRequest(
+        host="coordinator",
+        port=constants.DEFAULT_TLS_PORT,
+        client_session=ClientSession(
+            user="test",
+        ),
+        http_scheme=constants.HTTPS,
+        auth=trino.auth.OAuth2Authentication(redirect_auth_url_handler=redirect_handler))
+
+    response = request.post("select 1")
+
+    assert response.request.headers['Authorization'] == f"Bearer {token}"
+    assert get_token_callback.attempts == 0
+    assert len(_post_statement_requests()) == 2
+
+
 @pytest.mark.parametrize("attempts", [6, 10])
 @httprettified
 def test_oauth2_exceed_max_attempts(attempts, sample_post_response_data):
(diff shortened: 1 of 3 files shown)


tests/unit/test_client.py
```diff
@@ -384,6 +384,59 @@ def test_oauth2_authentication_flow(attempts, sample_post_response_data):
...
+def test_oauth2_refresh_token_flow(sample_post_response_data):
+    token = str(uuid.uuid4())
+    challenge_id = str(uuid.uuid4())
+
+    redirect_server = f"{REDIRECT_RESOURCE}/{challenge_id}"
+    token_server = f"{TOKEN_RESOURCE}/{challenge_id}"
+
+    post_statement_callback = PostStatementCallback(redirect_server, token_server, [token], sample_post_response_data)
+
+    # bind post statement
```
lukasz-walkiewicz, line 397: Is first post statement (binding + actual request) testing anything more than `test_oauth2_authentication_flow`? If not then I'd remove it.

trino/auth.py
```diff
@@ -273,7 +273,8 @@ class _OAuth2TokenBearer(AuthBase):
     _BEARER_PREFIX = re.compile(r"bearer", flags=re.IGNORECASE)
 
     def __init__(self, redirect_auth_url_handler: Callable[[str], None]):
-        self._redirect_auth_url = redirect_auth_url_handler
+        self._redirect_auth_url = None
```
lukasz-walkiewicz, line 276: Why would you want to store it?

trino/auth.py
```diff
@@ -322,15 +323,18 @@ def _attempt_oauth(self, response, **kwargs):
         auth_info_headers = parse_dict_header(_OAuth2TokenBearer._BEARER_PREFIX.sub("", auth_info, count=1))
 
         auth_server = auth_info_headers.get('x_redirect_server')
-        if auth_server is None:
+        if auth_server is None and self._redirect_auth_url is None:
```
lukasz-walkiewicz, line 326: I think we can remove this condition. If `x_redirect_server` is optional then it's optional. Previous value is not important.

tests/unit/test_client.py
sinkuladis: Done.

trino/auth.py
sinkuladis: You are right. We don't need to store it. I'm removing this part

trino/auth.py
sinkuladis: Yes, there is no need to check if `x_redirect_server` is present in this case. I've also updated tests cases that were failing because of the missing `x_redirect_server` in the response

tests/unit/test_client.py
```diff
@@ -384,6 +384,46 @@ def test_oauth2_authentication_flow(attempts, sample_post_response_data):
...
+        body=post_statement_callback)
+
+    # bind get token
+    get_token_callback = GetTokenCallback(token_server, token)
+    httpretty.register_uri(
+        method=httpretty.GET,
+        uri=token_server,
+        body=get_token_callback)
+
+    redirect_handler = RedirectHandler()
```
lukasz-walkiewicz, line 409: nit: you could use a redirect handler which would produce an error when called to ensure that it's not being used when `redirect_uri` is not returned.

lukasz-walkiewicz (approved):
LGTM but I think you need to sign CLA first.

cc: @hashhar

tests/unit/test_client.py
sinkuladis: Good idea. I've updated this part.

hashhar:
Thanks for reviewing the important parts @lukasz-walkiewicz.

@sinkuladis I'll take a look by tomorrow.

sinkuladis:
@lukasz-walkiewicz CLA is signed. My first push was signed-off with the wrong username and email, that's why bot was complaining. I've corrected commit author already 👍

tests/unit/test_client.py
```diff
@@ -430,10 +473,11 @@ def test_oauth2_exceed_max_attempts(attempts, sample_post_response_data):
 
 @pytest.mark.parametrize("header,error", [
     ("", "Error: header WWW-Authenticate not available in the response."),
-    ('Bearer"', 'Error: header info didn\'t have x_redirect_server'),
-    ('x_redirect_server="redirect_server", x_token_server="token_server"', 'Error: header info didn\'t match x_redirect_server="redirect_server", x_token_server="token_server"'),  # noqa: E501
+    ('Bearer"', 'Error: header info didn\'t have x_token_server'),
+    ('x_redirect_server="redirect_server", x_token_server="token_server"',
+     'Error: header info didn\'t match x_redirect_server="redirect_server", x_token_server="token_server"'),
+    # noqa: E501
```
hashhar, line 479: the `noqa` was here to prevent wrapping across lines. Revert the wrapping.

hashhar (approved):
LGTM % style nit

tests/unit/test_client.py
sinkuladis: missed that bit. Thanks!

hashhar (approved):
Thank you.

hashhar:
@sinkuladis Can you please see if the release note at https://github.com/trinodb/trino-python-client/issues/283#issuecomment-1319901998 captures the issue correctly.

i.e. without this fix if the access token had expired the refresh token flow didn't work correctly and instead failed with `Error: header info didn't have x_redirect_server`?

sinkuladis:
Yes, the release note captures it well. Hey, many thanks for reviewing my PR! :D

PR #297, opened 2022-12-11
Enhance `Cursor.description` with type information

mdesmet:
## Description

Add internal_size, precision, scale to `Cursor.description` to make it easier to get type information of a result set.

Note that if this information can be extracted from the type, this is also removed from the type_code, making it easier to match on.

## Release notes

```markdown
* Enhance `Cursor.description` with type information
```

### DIFF
--- a/trino/constants.py
+++ b/trino/constants.py
@@ -50,3 +50,7 @@ HEADER_DEALLOCATED_PREPARE = 'X-Trino-Deallocated-Prepare'
 
 HEADER_SET_SCHEMA = "X-Trino-Set-Schema"
 HEADER_SET_CATALOG = "X-Trino-Set-Catalog"
+
+LENGTH_TYPES = ["varchar"]
+PRECISION_TYPES = ["time", "time with time zone", "timestamp", "timestamp with time zone", "decimal"]
+SCALE_TYPES = ["decimal"]
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -21,12 +21,13 @@ import datetime
 import math
 import uuid
 from decimal import Decimal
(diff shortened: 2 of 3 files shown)


trino/constants.py
```diff
@@ -50,3 +50,7 @@
 
 HEADER_SET_SCHEMA = "X-Trino-Set-Schema"
 HEADER_SET_CATALOG = "X-Trino-Set-Catalog"
+
+LENGTH_TYPES = ["varchar"]
```
hovaesco, line 54: `char` is missing.

trino/dbapi.py
```diff
@@ -220,6 +221,31 @@ def cursor(self, experimental_python_types: bool = None):
...
+        type_signature = column["typeSignature"]
+        raw_type = type_signature["rawType"]
+        arguments = type_signature["arguments"]
+        return cls(
+            column["name"],  # name
+            raw_type if raw_type in PRECISION_TYPES + LENGTH_TYPES else column["type"],  # type_code
+            None,  # display_size
+            arguments[0]["value"] if len(arguments) >= 1 and raw_type in LENGTH_TYPES else None,  # internal_size
+            arguments[0]["value"] if len(arguments) >= 1 and raw_type in PRECISION_TYPES else None,  # precision
+            arguments[1]["value"] if len(arguments) >= 2 and raw_type in SCALE_TYPES else None,  # scale
```
hovaesco, line 244: why `len(arguments) >= 2` ?

mdesmet: We can drop this all together actually as we check on the types itself already.

mdesmet:
This can be looked at again after `PARAMETRIC_DATETIME` is merged as now no type information is exposed for temporal types.

PR #299, opened 2022-12-12
Support `VARBINARY` query parameter

mdesmet:
## Release notes

```markdown
```

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -532,6 +532,25 @@ def test_null_date_with_time_zone(trino_connection):
     assert rows[0][0] is None
 
 
+@pytest.mark.parametrize(
+    "binary_input",
+    [
+        bytearray("a", "utf-8"),
+        bytearray("a", "ascii"),
+        bytearray(b'\x00\x00\x00\x00'),
+        bytearray(4),
+        bytearray([1, 2, 3]),
+    ],
+)
+def test_binary_query_param(trino_connection, binary_input):
+    cur = trino_connection.cursor(experimental_python_types=True)
+
+    cur.execute("SELECT ?", params=(binary_input,))
+    rows = cur.fetchall()
+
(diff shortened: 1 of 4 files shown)


tests/integration/test_types_integration.py
```diff
@@ -121,10 +121,9 @@ def test_char(trino_connection):
 
 def test_varbinary(trino_connection):
     SqlTest(trino_connection) \
-        .add_field(sql="X'65683F'", python='ZWg/') \
-        .add_field(sql="X''", python='') \
-        .add_field(sql="CAST('' AS VARBINARY)", python='') \
-        .add_field(sql="from_utf8(CAST('😂😂😂😂😂😂' AS VARBINARY))", python='😂😂😂😂😂😂') \
```
hovaesco, line 127: Why you removed it?

hovaesco (approved):
Looks good % one comment

tests/integration/test_types_integration.py
mdesmet: Because `from_utf8` converts into `VARCHAR` so it is not relevant here.

hovaesco (approved):

mdesmet:
@hashhar: PTAL

tests/integration/test_dbapi_integration.py
```diff
@@ -532,6 +532,15 @@ def test_null_date_with_time_zone(trino_connection):
     assert rows[0][0] is None
 
 
+def test_binary_query_param(trino_connection):
+    cur = trino_connection.cursor(experimental_python_types=True)
+
+    cur.execute("SELECT ?", params=(bytearray("a", "utf-8"),))
```
hashhar, line 538: Also add a case with non-utf-8 bytes (to make sure implementation isn't doing the stupid thing of converting to `str` and then decoding/encoding to byte-array).

hashhar (approved):
LGTM % additional test

hashhar (approved):

PR #301, opened 2022-12-13
replacement pass on clearly exception NotImplementedError

morgangt:
## Description
There is a desire to make the code unified, everywhere in the code to use the NotImplementedError exception in all abstract methods.

## Non-technical explanation
Aesthetic desire to make everything logical

## Release notes

```markdown
```


morgangt:
I think simple will be make the code unified
Because somewhere use 
`raise NotImplementedError` 
https://github.com/trinodb/trino-python-client/blob/master/trino/auth.py#L170
and somewhere use simple `pass`

PR #308, opened 2022-12-23
Change experimental_python_types to legacy_primitive_types. Invert lo…

### LINKED ISSUE #305: Enable experimental_python_types by default and change the name
### Describe the feature

Enable `experimental_python_types` by default.
Change this parameter name to `legacy_python_types`.

_No response_

damian3031:
## Description
Resolves #305
Change experimental_python_types to legacy_primitive_types. Invert logic behind interpreting this parameter.

## Release notes

```markdown
* Change experimental_python_types to legacy_primitive_types. Invert logic behind interpreting this parameter. (#305 `issuenumber`)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,6 @@ engine = create_engine(
     connect_args={
       "session_properties": {'query_max_run_time': '1d'},
       "client_tags": ["tag1", "tag2"],
-      "experimental_python_types": True,
       "roles": {"catalog1": "role1"},
     }
 )
@@ -126,7 +125,6 @@ engine = create_engine(
     'trino://user@localhost:8080/system?'
     'session_properties={"query_max_run_time": "1d"}'
     '&client_tags=["tag1", "tag2"]'
-    '&experimental_python_types=true'
     '&roles={"catalog1": "role1"}'
 )
 
@@ -134,8 +132,7 @@ engine = create_engine(
 engine = create_engine(URL(
   host="localhost",
   port=8080,
-  client_tags=["tag1", "tag2"],
-  experimental_python_types=True
+  client_tags=["tag1", "tag2"]
 ))
 ```
 
@@ -440,35 +437,33 @@ The transaction is created when the first SQL statement is executed.
 exits the *with* context and the queries succeed, otherwise
 `trino.dbapi.Connection.rollback()` will be called.
 
-## Improved Python types
+## Legacy Primitive types
 
-If you enable the flag `experimental_python_types`, the client will convert the results of the query to the 
+By default, the client will convert the results of the query to the
 corresponding Python types. For example, if the query returns a `DECIMAL` column, the result will be a `Decimal` object.
+If you want to disable this behaviour, set flag `legacy_primitive_types` to `True`.
 
 Limitations of the Python types are described in the 
 [Python types documentation](https://docs.python.org/3/library/datatypes.html). These limitations will generate an 
-exception `trino.exceptions.DataError` if the query returns a value that cannot be converted to the corresponding Python 
+exception `trino.exceptions.TrinoDataError` if the query returns a value that cannot be converted to the corresponding Python 
 type.
 
 ```python
 import trino
-import pytz
-from datetime import datetime
 
 conn = trino.dbapi.connect(
-    experimental_python_types=True,
+    legacy_primitive_types=True,
     ...
 )
 
 cur = conn.cursor()
-
-params = datetime(2020, 1, 1, 16, 43, 22, 320000, tzinfo=pytz.timezone('America/Los_Angeles'))
-
-cur.execute("SELECT ?", params=(params,))
+# Negative DATE cannot be represented with Python types
+# legacy_primitive_types needs to be enabled
+cur.execute("SELECT DATE '-2001-08-22'")
 rows = cur.fetchall()
 
-assert rows[0][0] == params
-assert cur.description[0][1] == "timestamp with time zone"
+assert rows[0][0] == "-2001-08-22"
+assert cur.description[0][1] == "date"
 ```
 
 # Need help?
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -171,22 +171,11 @@ def test_execute_many_select(trino_connection):
     assert "Query must return update type" in str(e.value)
 
 
-@pytest.mark.parametrize("connection_experimental_python_types,cursor_experimental_python_types,expected",
-                         [
-                             (None, None, False),
-                             (None, False, False),
-                             (None, True, True),
-                             (False, None, False),
-                             (False, False, False),
-                             (False, True, True),
-                             (True, None, True),
-                             (True, False, False),
-                             (True, True, True),
-                         ])
-def test_experimental_python_types_with_connection_and_cursor(
-        connection_experimental_python_types,
-        cursor_experimental_python_types,
-        expected,
+@pytest.mark.parametrize("connection_legacy_primitive_types", [None, True, False])
+@pytest.mark.parametrize("cursor_legacy_primitive_types", [None, True, False])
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -456,7 +457,7 @@ import pytz
 from datetime import datetime
 
 conn = trino.dbapi.connect(
-    experimental_python_types=True,
+    legacy_primitive_types=False,
```
mdesmet, line 460: This doesn't really make sense as this is the default. I would just remove this code fragment

tests/integration/test_dbapi_integration.py
```diff
@@ -211,7 +211,7 @@ def test_experimental_python_types_with_connection_and_cursor(
     """)
     rows = cur.fetchall()
 
-    if expected:
+    if not expected:
```
mdesmet, line 214: Maybe it makes sense to rename `expected` into something more explicit, eg `legacy_primitive_types_expected`

tests/integration/test_dbapi_integration.py
```diff
@@ -231,7 +231,7 @@ def test_experimental_python_types_with_connection_and_cursor(
 
 
 def test_decimal_query_param(trino_connection):
-    cur = trino_connection.cursor(experimental_python_types=True)
+    cur = trino_connection.cursor(legacy_primitive_types=False)
```
mdesmet, line 234: Please omit the parameter as the default is tested in `test_legacy_primitive_types_with_connection_and_cursor`

README.md
damian3031: removed

tests/integration/test_dbapi_integration.py
damian3031: renamed to `expected_legacy_primitive_types`

tests/integration/test_dbapi_integration.py
damian3031: omitted

README.md
```diff
@@ -116,7 +116,7 @@ engine = create_engine(
     connect_args={
       "session_properties": {'query_max_run_time': '1d'},
       "client_tags": ["tag1", "tag2"],
-      "experimental_python_types": True,
+      "legacy_primitive_types": True,
```
mdesmet, line 119: Remove these entries, these code parts tend to be copied by users and we would rather not let them set this to True.

README.md
```diff
@@ -126,7 +126,7 @@ engine = create_engine(
     'trino://user@localhost:8080/system?'
     'session_properties={"query_max_run_time": "1d"}'
     '&client_tags=["tag1", "tag2"]'
-    '&experimental_python_types=true'
+    '&legacy_primitive_types=true'
```
mdesmet, line 129: Same here.

README.md
```diff
@@ -135,7 +135,7 @@ engine = create_engine(URL(
   host="localhost",
   port=8080,
   client_tags=["tag1", "tag2"],
-  experimental_python_types=True
+  legacy_primitive_types=True
```
mdesmet, line 138: Same here

README.md
```diff
@@ -456,7 +457,6 @@ import pytz
 from datetime import datetime
 
 conn = trino.dbapi.connect(
-    experimental_python_types=True,
```
mdesmet, line 459: this code part can be omitted.

README.md
damian3031: removed

README.md
damian3031: removed

README.md
damian3031: removed

README.md
damian3031: It is already omitted.

hovaesco (approved):

mdesmet (approved):

damian3031:
@hashhar would you have some time to look at this PR? Please approve if you are ok with these changes 🙂

README.md
hashhar: I think Michiel meant this entire snippet can now be removed since it tries to demo Python types.

We can instead change this snippet to return a value which cannot be represented with Python types like a negative DATE and show how to use `legacy_primitive_types`.

tests/integration/test_dbapi_integration.py
```diff
@@ -171,22 +171,23 @@ def test_execute_many_select(trino_connection):
...
+        (None, None, False),
+        (None, False, False),
+        (None, True, True),
+        (False, None, False),
+        (False, False, False),
+        (False, True, True),
+        (True, None, True),
+        (True, False, False),
+        (True, True, True),
+    ])
```
hashhar, line 186:
```suggestion
booleans = [None, True, False]

@pytest.mark.parametrize("connection_legacy_primitive_types", booleans)
@pytest.mark.parametrize("cursor_legacy_primitive_types", booleans)
@pytest.mark.parametrize("expected_legacy_primitive_types", booleans)
```

This results in cartesian product of all parameters being used.

hashhar (commented):
Do we have any test with `legacy_primitive_type=True` for values which cannot be represented by Python types - just to make sure that the primitive code path doesn't use Python types anywhere at all.

The important cases I can think of are with parametric timestamps since the parsing logic uses Python types IIRC.

README.md
```diff
@@ -442,8 +439,9 @@ exits the *with* context and the queries succeed, otherwise
 
 ## Improved Python types
```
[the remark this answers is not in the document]
damian3031: changed

tests/integration/test_dbapi_integration.py
damian3031: but, do we want to receive cartesian product and check every possible combination?

In my opinion we don't want to, because for example, let's have such case:
`(None, None, False)`
It means, that if we are not passing legacy_primitive_types to connection, nor to cursor, then expected value of this parameter is False, because this is the default.

Then, checking such combination:
`(None, None, True)`
is not needed, because such case doesn't exist.

Please correct me if I'm missing something.

hashhar: Can skip those cases within the test method. But I don't think those are bad cases to test (except None, None, True).

mdesmet: I don't agree actually, there is nothing being skipped. Only the first 2 columns are inputs (with each having 3 possible values). the third value is the expected value. There are `3*3` values to test. I think the matrix is actually better than having to provide some code within the test to skip the wrong occurrences of the expected value.

damian3031: I have one more idea, we could parametrize only two first columns, and calculate 3rd column based on them, something like:


```python
booleans = [None, True, False]
@pytest.mark.parametrize("connection_legacy_primitive_types", booleans)
@pytest.mark.parametrize("cursor_legacy_primitive_types", booleans)
def test_legacy_primitive_types_with_connection_and_cursor(
...


# If legacy_primitive_types is passed to cursor, take value from it.
# If not, take value from legacy_primitive_types passed to connection.
# If legacy_primitive_types is not passed to cursor nor connection, default to False
if cursor_legacy_primitive_types is not None:
    expected_legacy_primitive_types = cursor_legacy_primitive_types
elif connection_legacy_primitive_types is not None:
    expected_legacy_primitive_types = connection_legacy_primitive_types
else:
    expected_legacy_primitive_types = False



```

So we could get rid of matrix, but incorporate some logic into test. I'm not sure what do you think is better approach?

mdesmet: @damian3031 : Let's go that!

damian3031: @mdesmet @hashhar changed accordingly to snippet above, PTAL here.

mdesmet (approved):

hovaesco (approved):

hashhar (approved):

hashhar:
Just reworded the commit message.

As a follow-up do you think we should add a table of what Trino types map to what Python types and vice-versa?

Issue #306, opened 2022-12-22
Disable prepared statements by default


damian3031:
### Describe the feature

Disable prepared statements by default.
Exact approach and details about implementation could be discussed in the comments.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

damian3031:
In this draft PR #310 3 possible approaches of handling this issue were presented. All of them were thoroughly discussed. As a result, we decided that the best approach is nr 3 - transporting prepared statement parameters in the JSON body instead of the headers. It will fix biggiest issue here - hitting header limit size by query. Query will be still parsed by SQL parser on Trino side, as opposed to approach nr 2.

Due to the low priority and high complexity of this issue, it is currently on hold.

aalbu:
@hashhar, can we close this issue?  Based on comments in the referenced PR, it seems like the the main problem it's attempting to address is the header size limitations for prepared statements.  At this point, we have a way to avoid passing prepared statement parameters in the header, but specifying `legacy_prepared_statements=False`.

hashhar:
Yes, I think this is no longer as relevant because of EXECUTE IMMEDIATE.

We can revisit if there's still interest in this.

PR #310, opened 2023-01-05
Execute prepared statements on client side by default

damian3031:
Draft PR to discuss approach for #306


mdesmet:
Just to complete the information for @hashhar.

We discussed the following approaches

1. We could also avoid the prepare and deallocation query by simply doing it directly on the `ClientSession`. The advantage is that we can still make use of the query parser on the server side. The disadvantage is that we still use the headers as transport.

2. Do the parsing on the clientside. then we have to take care of values and identifiers within the query that contain `?`.

eg. `cur.execute("SELECT '?''?', ?", 'test')` or `cur.execute('select 1 as "?""test", ?', 'test')`

trino/dbapi.py
```diff
@@ -454,6 +454,20 @@ def execute(self, operation, params=None):
...
+                    'params must be a list or tuple containing the query '
+                    'parameter values'
+                )
+
+                # substitue parameters in query in reversed order
+                question_mark_positions = [index for index, character in enumerate(operation) if character == '?']
+                question_mark_positions.reverse()
+                for index, value in enumerate(reversed(params)):
+                    operation = "".join([operation[:question_mark_positions[index]],
+                                         "'", value, "'",
```
mdesmet, line 468: The value should not be enclosed by single quotes but use the appropriate type string.

trino/dbapi.py
```diff
@@ -454,6 +454,20 @@ def execute(self, operation, params=None):
...
 
         else:
+            if params:
+                assert isinstance(params, (list, tuple)), (
+                    'params must be a list or tuple containing the query '
+                    'parameter values'
+                )
+
+                # substitue parameters in query in reversed order
+                question_mark_positions = [index for index, character in enumerate(operation) if character == '?']
```
mdesmet, line 464: Alternative to the reverse trick is to use string slicing. maybe that would be slightly more readable?

hashhar:
Doing on the client side makes the most sense to me depending on what problem we're trying to solve.

1 (directly manipulate ClientSession) seems like a hack.
2 (client side) solves the problem of multiple queries but introduces security concerns.
3 (new backward compatible API) solves the limitation that prepared statements cannot be very large but it doesn't help the fact that multiple queries need to be sent anyway.

I think we can do the following:

- Make `paramstyle` settable and only allow it to be set to either `format` or `qmark` with `qmark` as default.
- When we create Connection we "freeze" the paramstyle value i.e. a Connection's paramstyle is constant.
- `paramstyle` = `format` would do client side bindings, `paramstyle` = `qmark` would do server side (what is already done today)
- For paramstyle = format it's easier to do client side binding since you no longer need to worry about "finding" all markers and can just use `str % values` from Python to do the value setting.

For paramstyle = format we need to:

- escape all params only - query text needs no escaping - if something needs escaping there (`%`) then the user should do that - this is what all Python drivers do (I checked psycopg2, pymysql, Oracle).
- Do `query % escaped_values` and execute the query.

Where the handling of qmark and format style needs to be done? - to me it seems the logical place is to have two `Cursor` classes and override their `execute` and `executemany` methods. Alternatively we can keep existing cursor and introduce two methods `process_qmark_params` and `process_format_params` and call them from `execute` and `executemany`.

Now the question we need to answer is:

What's the biggest problem we want to solve?

1. multiple queries sent when using params?
2. the size limitation of prepared statements?

If we want to solve 1 then the only solution is what we discuss above.
If we want to solve 2 then the API change that @mdesmet mentioned in his comment is a good solution.

To be honest I don't know which of these problems people actually care about.

The benefit of customizable `paramstyle` is that we can solve both problems at once but at the cost of some additional code and complexity (both for users and for us).

PR #311, opened 2023-01-05
Encode roles as in JDBC driver

mdesmet:
## Description

In the jdbc driver custom roles are enclosed in `ROLE{rolename}` automatically. This was not yet implemented in the Trino python client, causing confusion.

This PR adds this enclosing automatically, while checking if the enclosing is not already provided to ensure backwards compatiblity.

## Release notes

```markdown
* Enclose custom roles passed in Connection into Trino protocol specific formatting (ROLE{rolename})
```

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -85,6 +85,8 @@ for i in range(0, 13):
 MAX_PYTHON_TEMPORAL_PRECISION_POWER = 6
 MAX_PYTHON_TEMPORAL_PRECISION = POWERS_OF_TEN[MAX_PYTHON_TEMPORAL_PRECISION_POWER]
 
+ROLE_PATTERN = re.compile(r"^ROLE\{(.*)\}$")
+
 
 class ClientSession(object):
     """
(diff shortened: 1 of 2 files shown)


hovaesco (commented):
Thank you for doing this, maybe we should also have an example with an old approach in README?

mdesmet:
> Thank you for doing this, maybe we should also have an example with an old approach in README?

Well, the old approach is not anymore relevant after this PR. Our docs were already compliant, see https://github.com/trinodb/trino-python-client/blob/master/README.md#roles

hovaesco (approved):

trino/client.py
```diff
@@ -234,6 +236,15 @@ def timezone(self):
         with self._object_lock:
             return self._timezone
 
+    def _format_roles(self, roles):
+        formatted_roles = {}
+        for catalog, role in roles.items():
+            if role in ("NONE", "ALL") or ROLE_PATTERN.match(role) is not None:
```
hashhar, line 242: If role matches the ROLE_PATTERN can we use `warnings.warn` to tell user to fix it instead of silently accepting it?

Long term supporting both syntaxes may not actually solve the confusion.

@mdesmet Can you do this as a follow-up PR?

PR #315, opened 2023-01-07
Enhance `Cursor.description` with type information

mdesmet:
## Description

Add internal_size, precision, scale to Cursor.description to make it easier to get type information of a result set.

Note that if this information can be extracted from the type, this is also removed from the type_code, making it easier to match upon.

## Release notes

```markdown
* Add type information to DBAPI's cursor description (internal_size, scale, precision)
```

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -119,6 +119,7 @@ def test_none_query_param(trino_connection):
     rows = cur.fetchall()
 
     assert rows[0][0] is None
+    assert_cursor_description(cur, trino_type="unknown")
 
 
 def test_string_query_param(trino_connection):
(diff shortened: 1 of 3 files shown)


trino/constants.py
```diff
@@ -53,3 +53,7 @@
 HEADER_SET_CATALOG = "X-Trino-Set-Catalog"
 
 HEADER_CLIENT_CAPABILITIES = "X-Trino-Client-Capabilities"
+
+LENGTH_TYPES = ["char", "varchar"]
```
hovaesco, line 57: `char` missing in tests

hovaesco (approved):

tests/integration/test_dbapi_integration.py
```diff
@@ -1234,3 +1259,12 @@ def test_describe_table_query(run_trino):
             aliased=False,
         )
     ]
+
+
+def assert_cursor_description(cur, trino_type, size=None, precision=None, scale=None):
+    assert cur.description[0][1] == trino_type
+    assert cur.description[0][2] is None
```
hashhar, line 1266: According to db-api the first two params are mandatory. Why do we assert here that type_code is None?

hashhar (approved):

PR #316, opened 2023-01-09
Support execute chaining

### LINKED ISSUE #279: Support Execute Chaining
### Describe the feature

Would it make more sense to return `self` from `cursor.execute` instead of `TrinoResult`

This would allow you to chain fetch operations
```python
value = cursor.execute('SELECT 1').fetchone()[0]
```

In addition, `__iter__` and `rowcount` can be implemented on the cursor itself - this seems to be how it is commonly done.

So with `return self` and `__iter__` implemented the following would still be possible for convenience:

```python
for row in cursor.execute('select * from system.runtime.nodes'):
    print(row) 
```

This is the behavior of [psycopg3](https://github.com/psycopg/psycopg/blob/master/psycopg/psycopg/cursor.py#L726)
The Python [DBAPI2 spec](https://peps.python.org/pep-0249/#id20) doesn't specify the return value of `execute` so it's up to the driver to decide, but it is common in quite a few Python DBAPI2 drivers.

_No response_

damian3031:
resolves #279 

Now it is possible to chain fetch operations. Added test for it.

### DIFF
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -445,7 +445,7 @@ class Cursor(object):
                 self._query = self._execute_prepared_statement(
                     statement_name, params
                 )
-                result = self._query.execute()
+                self._iterator = iter(self._query.execute())
             finally:
                 # Send deallocate statement
                 # At this point the query can be deallocated since it has already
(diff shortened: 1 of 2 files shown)


trino/dbapi.py
```diff
@@ -456,9 +456,8 @@ def execute(self, operation, params=None):
         else:
             self._query = trino.client.TrinoQuery(self._request, sql=operation,
                                                   legacy_primitive_types=self._legacy_primitive_types)
-            result = self._query.execute()
-        self._iterator = iter(result)
-        return result
+            self._iterator = iter(self._query.execute())
+        return self
 
     def executemany(self, operation, seq_of_params):
```
hashhar, line 462: For consistency do we also want it for `executemany`?

damian3031: added for consistency

mdesmet (approved):

hashhar (approved):

PR #323, opened 2023-01-18
Provide query read-only property on dbapi cursor

mdesmet:
## Description

Provide sql read-only property on dbapi cursor

## Release notes

```markdown
* Provide executed SQL statement as a read-only property on dbapi's cursor
```

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -732,7 +732,7 @@ class TrinoQuery(object):
     def __init__(
             self,
             request: TrinoRequest,
-            sql: str,
+            query: str,
             legacy_primitive_types: bool = False,
     ) -> None:
         self._query_id: Optional[str] = None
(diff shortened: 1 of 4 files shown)


trino/client.py
```diff
@@ -747,6 +746,14 @@ def __init__(
         self._legacy_primitive_types = legacy_primitive_types
         self._row_mapper: Optional[RowMapper] = None
 
+    @property
+    def query_id(self) -> Optional[str]:
+        return self._query_id
+
+    @property
+    def sql(self) -> Optional[str]:
```
hovaesco, line 754: Perhaps a better name is `query`, it will follow the convention as used in `query_id`. `Psycopg` follow this as well.

hashhar:
rebased

hashhar:
pre-commit failures would be fixed in https://github.com/trinodb/trino-python-client/pull/336

hashhar (approved):

Issue #292, opened 2022-11-22
Enable mypy checks for the modules


hovaesco:
This PR https://github.com/trinodb/trino-python-client/pull/208 introduced mypy checks for subset of modules `trino.exceptions`, `trino.logging`, and `trino.transactions`.  Enable checks for the rest of modules, which are disabled here: https://github.com/trinodb/trino-python-client/blob/master/setup.cfg#L22

PR #324, opened 2023-01-18
Hovaesco/mypy2

damian3031:
part of https://github.com/trinodb/trino-python-client/issues/292

Added type hints and enabled mypy checks for `client.py`, `dbapi.py`


trino/dbapi.py
```diff
@@ -151,50 +153,53 @@ def __init__(
...
+    def isolation_level(self) -> IsolationLevel:
         return self._isolation_level
 
     @property
-    def transaction(self):
+    def transaction(self) -> Optional[Transaction]:
         return self._transaction
 
-    def __enter__(self):
+    def __enter__(self) -> object:
```
mdesmet, line 167:
```suggestion
    def __enter__(self) -> "Connection":
```

trino/dbapi.py
```diff
@@ -234,21 +239,21 @@ class DescribeOutput(NamedTuple):
...
+    display_size: Optional[int]
     internal_size: int
     precision: int
     scale: int
-    null_ok: bool
+    null_ok: Optional[bool]
 
     @classmethod
-    def from_column(cls, column: Dict[str, Any]):
+    def from_column(cls, column: Dict[str, Any]) -> Any:
```
mdesmet, line 256:
```suggestion
    def from_column(cls, column: Dict[str, Any]) -> "ColumnDescription":
```

trino/dbapi.py
```diff
@@ -271,7 +286,10 @@ class Cursor(object):
 
     """
 
-    def __init__(self, connection, request, legacy_primitive_types: bool = False):
+    def __init__(self,
+                 connection: Connection,
+                 request: trino.client.TrinoRequest,
+                 legacy_primitive_types: bool = False):
```
mdesmet, line 292:
```suggestion
                 legacy_primitive_types: bool = False) -> None:
```

trino/dbapi.py
```diff
@@ -280,32 +298,32 @@ def __init__(self, connection, request, legacy_primitive_types: bool = False):
...
-    def update_type(self):
+    def update_type(self) -> Optional[str]:
         if self._query is not None:
             return self._query.update_type
         return None
 
     @property
-    def description(self) -> List[ColumnDescription]:
-        if self._query.columns is None:
+    def description(self) -> Optional[List[Tuple[Any, ...]]]:
```
mdesmet, line 325: Does mypy complain about `List[ColumnDescription]`?

trino/dbapi.py
```diff
@@ -584,66 +603,22 @@ def describe(self, sql: str) -> List[DescribeOutput]:
 
         return list(map(lambda x: DescribeOutput.from_row(x), result))
 
-    def genall(self):
-        return self._query.result
+    def genall(self) -> Any:
```
mdesmet, line 606:
```suggestion
    def genall(self) -> Optional[trino.client.TrinoResult]:
```

trino/exceptions.py
```diff
@@ -72,7 +72,7 @@ class TrinoDataError(NotSupportedError):
 
 
 class TrinoQueryError(Error):
-    def __init__(self, error: Dict[str, Any], query_id: Optional[str] = None) -> None:
+    def __init__(self, error: Any, query_id: Optional[str] = None) -> None:
```
mdesmet, line 75: Doesn't the error come from JSON, so I would expect it to follow some certain format?

damian3031: I would expect the same, but [here](https://github.com/trinodb/trino-python-client/blob/57c97061ac1735675d4cf43da16b550dcb15578e/trino/client.py#L858) `str` is passed as an argument to `error` parameter.
It is causing mypy error:
```
trino/client.py:858: error: Argument 1 to "TrinoUserError" has incompatible type "str"; expected "Dict[str, Any]"  [arg-type]
```

mdesmet: I would say that we need to fix that occurrence. If you have a look at `TrinoQueryError`, you'll see this is actually an issue.

trino/client.py
```diff
@@ -855,7 +855,7 @@ def execute(
         call fetch() until finished is true.
         """
         if self.cancelled:
-            raise exceptions.TrinoUserError("Query has been cancelled", self.query_id)
+            raise exceptions.TrinoUserError({"message": "Query has been cancelled"}, self.query_id)
```
mdesmet, line 858: Can you revert this line and put logic in the `TrinoQueryError` constructor to convert a str input into a dict if a str is passed into error?

damian3031: Done

trino/dbapi.py
```diff
@@ -280,32 +298,32 @@ def __init__(self, connection, request, legacy_primitive_types: bool = False):
         self._request = request
 
         self.arraysize = 1
-        self._iterator = None
-        self._query = None
+        self._iterator: Optional[Iterator[Any]] = None
```
mdesmet, line 301: Isn't this an `Iterator[List[Any}}`?

mdesmet (commented):
LGTM % the Iterator comment

trino/client.py
```diff
@@ -75,7 +87,7 @@
 else:
     PROXIES = {}
 
-_HEADER_EXTRA_CREDENTIAL_KEY_REGEX = re.compile(r'^\S[^\s=]*$')
+_HEADER_EXTRA_CREDENTIAL_KEY_REGEX = re.compile(r"^\S[^\s=]*$")
```
hashhar, line 90: separate out a commit, having functional and non-functional changes in single commit makes it harder to read history (bisecting for example) in future.

trino/client.py
```diff
@@ -317,15 +347,19 @@ def __repr__(self):
 
 class _DelayExponential(object):
     def __init__(
-            self, base=0.1, exponent=2, jitter=True, max_delay=2 * 3600  # 100ms  # 2 hours
-    ):
+        self,
+        base: float = 0.1,
+        exponent: int = 2,
+        jitter: bool = True,
+        max_delay: int = 2 * 3600,  # 100ms  # 2 hours
```
hashhar, line 354: the 100 ms should go against `base`.

trino/client.py
```diff
@@ -334,11 +368,15 @@ def __call__(self, attempt):
 
 class _RetryWithExponentialBackoff(object):
     def __init__(
-            self, base=0.1, exponent=2, jitter=True, max_delay=2 * 3600  # 100ms  # 2 hours
-    ):
+        self,
+        base: float = 0.1,
+        exponent: int = 2,
+        jitter: bool = True,
+        max_delay: int = 2 * 3600,  # 100ms  # 2 hours
```
hashhar, line 375: the 100ms should go against `base`

trino/client.py
```diff
@@ -334,11 +368,15 @@ def __call__(self, attempt):
...
+    ) -> None:
         self._get_delay = _DelayExponential(base, exponent, jitter, max_delay)
 
-    def retry(self, func, args, kwargs, err, attempt):
+    def retry(self, func: Any, args: Any, kwargs: Any, err: Any, attempt: int) -> None:
```
hashhar, line 379: aren't kwargs always dicts?

trino/client.py
```diff
@@ -397,12 +435,14 @@ def __init__(
...
-        http_scheme: str = None,
+        http_scheme: Optional[str] = None,
         auth: Optional[Any] = constants.DEFAULT_AUTH,
         redirect_handler: Any = None,
         max_attempts: int = MAX_ATTEMPTS,
-        request_timeout: Union[float, Tuple[float, float]] = constants.DEFAULT_REQUEST_TIMEOUT,
-        handle_retry=_RetryWithExponentialBackoff(),
+        request_timeout: Union[
+            float, Tuple[float, float]
+        ] = constants.DEFAULT_REQUEST_TIMEOUT,
```
hashhar, line 444: The constant is a plain float. Where is the Tuple[float, float] coming from?

trino/client.py
```diff
@@ -562,7 +614,12 @@ def post(self, sql: str, additional_http_headers: Optional[Dict[str, Any]] = Non
             while http_response is not None and http_response.is_redirect:
                 location = http_response.headers["Location"]
                 url = self._redirect_handler.handle(location)
-                logger.info("redirect %s from %s to %s", http_response.status_code, location, url)
+                logger.info(
+                    "redirect %s from %s to %s",
+                    http_response.status_code,
+                    location,
+                    url,
+                )
```
hashhar, line 622: extract a commit

trino/client.py
```diff
@@ -606,11 +665,11 @@ def raise_response_error(self, http_response):
         raise exceptions.HttpError(
             "error {}{}".format(
                 http_response.status_code,
-                ": {}".format(http_response.content) if http_response.content else "",
+                ": {!r}".format(http_response.content) if http_response.content else "",
```
hashhar, line 668: what does this do?

trino/client.py
```diff
@@ -633,14 +692,18 @@ def process(self, http_response) -> TrinoStatus:
                 self._client_session.properties[key] = value
 
         if constants.HEADER_SET_CATALOG in http_response.headers:
-            self._client_session.catalog = http_response.headers[constants.HEADER_SET_CATALOG]
+            self._client_session.catalog = http_response.headers[
+                constants.HEADER_SET_CATALOG
+            ]
```
hashhar, line 697: extract commit

trino/client.py
```diff
@@ -633,14 +692,18 @@ def process(self, http_response) -> TrinoStatus:
...
         if constants.HEADER_SET_SCHEMA in http_response.headers:
-            self._client_session.schema = http_response.headers[constants.HEADER_SET_SCHEMA]
+            self._client_session.schema = http_response.headers[
+                constants.HEADER_SET_SCHEMA
+            ]
 
         if constants.HEADER_SET_ROLE in http_response.headers:
             for key, value in get_roles_values(
-                    http_response.headers, constants.HEADER_SET_ROLE
+                http_response.headers, constants.HEADER_SET_ROLE
```
hashhar, line 706: extract commit

trino/client.py
```diff
@@ -669,19 +732,23 @@ def process(self, http_response) -> TrinoStatus:
...
         """
         Verifies that key has ASCII only and non-whitespace characters.
         """
         key = header[0]
 
         if not _HEADER_EXTRA_CREDENTIAL_KEY_REGEX.match(key):
-            raise ValueError(f"whitespace or '=' are disallowed in extra credential '{key}'")
+            raise ValueError(
+                f"whitespace or '=' are disallowed in extra credential '{key}'"
+            )
```
hashhar, line 744: extract commit

trino/client.py
```diff
@@ -669,19 +732,23 @@ def process(self, http_response) -> TrinoStatus:
...
 
         try:
-            key.encode().decode('ascii')
+            key.encode().decode("ascii")
         except UnicodeDecodeError:
-            raise ValueError(f"only ASCII characters are allowed in extra credential '{key}'")
+            raise ValueError(
+                f"only ASCII characters are allowed in extra credential '{key}'"
+            )
```
hashhar, line 751: extract commit

trino/client.py
```diff
@@ -745,38 +812,41 @@ def __init__(
         self._sql = sql
         self._result: Optional[TrinoResult] = None
         self._legacy_primitive_types = legacy_primitive_types
-        self._row_mapper: Optional[RowMapper] = None
+        self._row_mapper: Optional[Union[RowMapper, NoOpRowMapper]] = None
```
hashhar, line 815: instead of a union we should define a common super type. That has the benefit that in future if we change RowMapper we also are reminded to change NoOpRowMapper as well.

trino/client.py
```diff
@@ -745,38 +812,41 @@ def __init__(
...
     @property
-    def columns(self):
+    def columns(self) -> Any:
         if self.query_id:
             while not self._columns and not self.finished and not self.cancelled:
                 # Columns are not returned immediately after query is submitted.
                 # Continue fetching data until columns information is available and push fetched rows into buffer.
-                self._result.rows += self.fetch()
+                if self._result:
+                    self._result.rows += self.fetch()
```
hashhar, line 824: this is a bugfix, separate commit

trino/client.py
```diff
@@ -846,7 +916,10 @@ def cancel(self) -> None:
 
     def is_finished(self) -> bool:
         import warnings
-        warnings.warn("is_finished is deprecated, use finished instead", DeprecationWarning)
+
+        warnings.warn(
+            "is_finished is deprecated, use finished instead", DeprecationWarning
+        )
```
hashhar, line 922: extract commit

trino/client.py
```diff
@@ -894,26 +974,26 @@ def map(self, value: Any) -> Optional[T]:
...
         if value is None:
             return None
-        if value == 'Infinity':
+        if value == "Infinity":
             return float("inf")
-        if value == '-Infinity':
+        if value == "-Infinity":
             return float("-inf")
-        if value == 'NaN':
+        if value == "NaN":
```
hashhar, line 996: extract commit

trino/client.py
```diff
@@ -1095,42 +1175,45 @@ def map(self, value) -> Optional[datetime]:
...
         self.mappers = mappers
 
-    def map(self, values: List[Any]) -> Optional[Tuple[Optional[Any], ...]]:
+    def map(self, values: Optional[List[Any]]) -> Optional[Tuple[Optional[Any], ...]]:
         if values is None:
             return None
-        return tuple(self.mappers[index].map(value) for index, value in enumerate(values))
+        return tuple(
+            self.mappers[index].map(value) for index, value in enumerate(values)
+        )
```
hashhar, line 1203: extract commit

trino/client.py
```diff
@@ -1095,42 +1175,45 @@ def map(self, value) -> Optional[datetime]:
...
         self.value_mapper = value_mapper
 
     def map(self, values: Any) -> Optional[Dict[Any, Optional[Any]]]:
         if values is None:
             return None
         return {
-            self.key_mapper.map(key): self.value_mapper.map(value) for key, value in values.items()
+            self.key_mapper.map(key): self.value_mapper.map(value)
+            for key, value in values.items()
         }
```
hashhar, line 1217: extract commit

trino/client.py
```diff
@@ -1191,31 +1305,8 @@ def _create_value_mapper(self, column) -> ValueMapper:
...
 
-    def _get_precision(self, column: Dict[str, Any]):
+    def _get_precision(self, column: Dict[str, Any]) -> int:
         args = column['arguments']
         if len(args) == 0:
             return 3
         return args[0]['value']
-
-
-class RowMapper:
```
hashhar, line 1201: why moved?

hashhar: extract just the movement to separate commit with reasoning in the msg

trino/dbapi.py
```diff
@@ -72,16 +84,6 @@
...
 
-def connect(*args, **kwargs):
-    """Constructor for creating a connection to the database.
-
-    See class :py:class:`Connection` for arguments.
-
-    :returns: a :py:class:`Connection` object.
-    """
-    return Connection(*args, **kwargs)
-
```
hashhar, line 83: extract no-op movements to a single commit

trino/dbapi.py
```diff
@@ -151,50 +153,53 @@ def __init__(
...
-    def start_transaction(self):
+    def start_transaction(self) -> Transaction:
         self._transaction = Transaction(self._create_request())
         self._transaction.begin()
         return self._transaction
 
-    def commit(self):
-        if self.transaction is None:
+    def commit(self) -> None:
+        if self._transaction is None:
```
hashhar, line 191: why go directly to field?

trino/dbapi.py
```diff
@@ -207,7 +212,7 @@ def _create_request(self):
             self.request_timeout,
         )
 
-    def cursor(self, legacy_primitive_types: bool = None):
+    def cursor(self, legacy_primitive_types: Optional[bool] = None) -> 'Cursor':
```
hashhar, line 215: `-> 'Cursor'`? Why no qualify the name?

hashhar: here and other places where we use "strings" as return types

trino/dbapi.py
```diff
@@ -503,6 +521,7 @@ def executemany(self, operation, seq_of_params):
         for parameters in seq_of_params[:-1]:
             self.execute(operation, parameters)
             self.fetchall()
+            assert self._query is not None
```
hashhar, line 524: potential bugfix, extract a commit

trino/dbapi.py
```diff
@@ -584,66 +603,22 @@ def describe(self, sql: str) -> List[DescribeOutput]:
 
         return list(map(lambda x: DescribeOutput.from_row(x), result))
 
-    def genall(self):
-        return self._query.result
+    def genall(self) -> Any:
+        if self._query:
+            return self._query.result
+        return None
```
hashhar, line 609: potential bugfix, extract a commit

trino/dbapi.py
```diff
@@ -584,66 +603,22 @@ def describe(self, sql: str) -> List[DescribeOutput]:
...
         self._query.cancel()
 
-    def close(self):
+    def close(self) -> None:
         self.cancel()
         # TODO: Cancel not only the last query executed on this cursor
         #  but also any other outstanding queries executed through this cursor.
-
-
-Date = datetime.date
```
hashhar, line 606: Why removed? These are standard DB-API types and users can inspect these to understand what they'll receive and what they need to pass.

hashhar (commented):
first quick pass

trino/client.py
damian3031: dictionary unpacking operator `**` is missing here, so `kwargs` parameter could be virtually anything. 
Parameters: `func`, `args`, `kwargs`, `err` are not used anywhere in this method anyway. Maybe I could just delete these not used parameters?

trino/client.py
damian3031: extracted to separate  commit

trino/dbapi.py
damian3031: extracted to separate  commit

trino/dbapi.py
damian3031: removal reverted

trino/dbapi.py
damian3031: Used `transaction` property, instead of directly `_transaction` attribute

trino/client.py
damian3031: `!r` calls repr() on the expression ([source](https://peps.python.org/pep-0498/#specification))
PEP498 mention, that this is redundant ([source](https://peps.python.org/pep-0498/#s-r-and-a-are-redundant))

but without it, mypy throws error:
```
error: On Python 3 formatting "b'abc'" with "{}" produces "b'abc'", not "abc"; use "{!r}" if this is desired behavior  [str-bytes-safe]
```

trino/client.py
hashhar: I think the retry method was written very optimistically. I agree to simplify it to cover current usages only.

trino/client.py
hashhar: Does mypy stop complaining if you add repr around the argument to format instead of changing format string? I think it'll be more obvious.

trino/client.py
damian3031: done

trino/client.py
damian3031: It works, changed according to you suggestion

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: Extracted to separate commit

trino/client.py
damian3031: corrected

trino/client.py
damian3031: corrected

trino/client.py
damian3031: changed type hint to `float` only

trino/dbapi.py
damian3031: Changed to `Optional[Iterator[List[Any]]]`

trino/client.py
damian3031: Reverted. Used [Postponed Evaluation of Annotations](https://peps.python.org/pep-0563/) instead of reordering class definition. Reason was that type hint contained names that have not been defined yet.

trino/dbapi.py
damian3031: Reverted. Used [Postponed Evaluation of Annotations](https://peps.python.org/pep-0563/) instead of reordering class definition. Reason was that type hint contained names that have not been defined yet.

trino/dbapi.py
damian3031: Fixed to use [Postponed Evaluation of Annotations](https://peps.python.org/pep-0563/) when type hint contains names that have not been defined yet.

trino/client.py
damian3031: Defined common super type `RowMapperBase` [here](https://github.com/trinodb/trino-python-client/blob/3de613f7314803e42544fd886a73b9212e6a3097/trino/client.py#L1220)

trino/dbapi.py
damian3031: mypy throws error then:
```trino/dbapi.py:615: error: Argument 1 to "list" has incompatible type "Optional[TrinoResult]"; expected "Iterable[List[Any]]"  [arg-type]```

changing to `Iterable[List[Any]]` throws error elsewhere, and so on

damian3031:
@hashhar I've adjusted to all your comments, PTAL

PR #332, opened 2023-02-14
[WIP] Configurable targetResultSize

DanielMorales9:
## Release notes

```markdown
```


mdesmet:
Could you elaborate a bit on your use case? What changes do you make to this property and what are the results? What type of queries require changing this property?

mdesmet:
We discussed this change request and would advise filing this request on Trino itself, ideally it should be managed automatically by Trino or through a setting instead of driven by the client.

hashhar:
Closing this because it's stale and we think the value needs to be configurable on the server instead of in every single client.

PR #335, opened 2023-02-17
Make names of ROW datatype available in result set

mdesmet:
## Description

```
cur.execute("SELECT CAST(ROW(1, 2e0) AS ROW(x BIGINT, y DOUBLE))")
rows = cur.fetchall()

assert rows[0][0] == (1, 2.0)
assert rows[0][0].x == 1
assert rows[0][0].y == 2.0

cur.execute("SELECT CAST(ROW(1, ROW(1, 2e0)) AS ROW(x BIGINT, y ROW(x BIGINT, y DOUBLE)))")
rows = cur.fetchall()

assert rows[0][0] == (1, (1, 2.0))
assert rows[0][0].x == 1
assert rows[0][0].y.x == 1
assert rows[0][0].y.y == 2.0
```

## Release notes

```markdown
* Make names of ROW datatype available in result set
```

Fixes https://github.com/trinodb/trino-python-client/issues/338

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -745,6 +745,49 @@ def test_named_row(trino_connection):
     rows = cur.fetchall()
 
     assert rows[0][0] == (1, 2.0)
+    assert rows[0][0][0] == 1
+    assert rows[0][0][1] == 2.0
+    assert rows[0][0].x == 1
+    assert rows[0][0].y == 2.0
+
+    assert rows[0][0].__annotations__["names"] == ['x', 'y']
+    assert rows[0][0].__annotations__["types"] == ['bigint', 'double']
+
+
+def test_named_row_duplicate_names(trino_connection):
+    cur = trino_connection.cursor()
+    cur.execute("SELECT CAST(ROW(1, 2e0) AS ROW(x BIGINT, x DOUBLE))")
+    rows = cur.fetchall()
+
+    assert rows[0][0] == (1, 2.0)
+    with pytest.raises(ValueError, match="Ambiguous row field reference: x"):
+        rows[0][0].x
+
+    assert rows[0][0].__annotations__["names"] == ['x', 'x']
+    assert rows[0][0].__annotations__["types"] == ['bigint', 'double']
+    assert str(rows[0][0]) == "(1, 2.0)"
+
+
+def test_nested_named_row(trino_connection):
+    cur = trino_connection.cursor()
+    cur.execute("SELECT CAST(ROW(DECIMAL '2.3', ROW(1, 'test')) AS ROW(x DECIMAL(3,2), y ROW(x BIGINT, y VARCHAR)))")
+    rows = cur.fetchall()
+
+    assert rows[0][0] == (Decimal('2.3'), (1, 'test'))
+    assert rows[0][0][0] == Decimal('2.3')
+    assert rows[0][0][1] == (1, 'test')
+    assert rows[0][0][1][0] == 1
+    assert rows[0][0][1][1] == 'test'
+    assert rows[0][0].x == Decimal('2.3')
+    assert rows[0][0].y.x == 1
+    assert rows[0][0].y.y == 'test'
+
+    assert rows[0][0].__annotations__["names"] == ['x', 'y']
+    assert rows[0][0].__annotations__["types"] == ['decimal', 'row']
+
+    assert rows[0][0].y.__annotations__["names"] == ['x', 'y']
+    assert rows[0][0].y.__annotations__["types"] == ['bigint', 'varchar']
+    assert str(rows[0][0]) == "(x: Decimal('2.30'), y: (x: 1, y: 'test'))"
 
 
 def test_float_query_param(trino_connection):
--- a/trino/client.py
+++ b/trino/client.py
@@ -1128,14 +1128,48 @@ class ArrayValueMapper(ValueMapper[List[Optional[Any]]]):
         return [self.mapper.map(value) for value in values]
 
 
+class NamedRowTuple(tuple):
+    """Custom tuple class as namedtuple doesn't support missing or duplicate names"""
+    def __new__(cls, values, names: List[str], types: List[str]):
+        return super().__new__(cls, values)
+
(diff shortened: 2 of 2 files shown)


mdesmet:
@hashhar: i didn't change the `repr` it still uses tuple's implementation. Do you think we should also adapt to show the attributes?

trino/client.py
```diff
@@ -1128,14 +1128,26 @@ def map(self, values: List[Any]) -> Optional[List[Any]]:
         return [self.mapper.map(value) for value in values]
 
 
+class NamedRowTuple(tuple):
+    """Custom tuple class as namedtuple doesn't support missing or duplicate names"""
```
hashhar, line 1132: IMO the behaviour should be:

- If duplicate fields exist return a plain tuple/throw if named tuple is requested
  - Current behaviour is bad because it leads to making some fields return "wrong" values when accessed by name
- If a field name is missing generate one as `field + str(n)` where `n` is a counter
- If all fields have unique names return a namedtuple

tests/integration/test_dbapi_integration.py
```diff
@@ -745,6 +745,28 @@ def test_named_row(trino_connection):
...
+
+def test_nested_named_row(trino_connection):
+    cur = trino_connection.cursor()
+    cur.execute("SELECT CAST(ROW(1, ROW(1, 2e0)) AS ROW(x BIGINT, y ROW(x BIGINT, y DOUBLE)))")
+    rows = cur.fetchall()
+
+    assert rows[0][0] == (1, (1, 2.0))
+    assert rows[0][0].x == 1
+    assert rows[0][0].y.x == 1
+    assert rows[0][0].y.y == 2.0
```
hashhar, line 769: Also assert that the fields can still be accessed as a plain tuple.

trino/client.py
```diff
@@ -1128,14 +1128,26 @@ def map(self, values: List[Any]) -> Optional[List[Any]]:
...
+    def __new__(cls, values, names):
+        return super().__new__(cls, values)
+
+    def __init__(self, values, names):
+        for name, value in zip(names, values):
+            if name is not None and not hasattr(self, name):
+                setattr(self, name, value)
```
abhiagarwal01, line 1139: Could type information be added as well for each row? We often convert the results of our trino queries into polars dataframes, and by providing type information polars can infer the schema so we can stream the data back to arrow IPC. Should be possible by updating `self.__annotations__[name]` with the relevant type. Can open a PR for this functionality as well.

mdesmet: Interesting! I'll give that a try.

trino/client.py
mdesmet: I think it should behave like Trino for ambiguous field names. 

```
trino> SELECT CAST(ROW(1, 2.0) AS ROW(x BIGINT, x DOUBLE)).x;
Query 20230306_185114_07832_kzrg9 failed: line 1:53: Ambiguous row field reference: x
```

trino/client.py
mdesmet: I added names and types annotations, please have a look at the tests.

mdesmet: I didn't want to go with `self.__annotations__[name]` as that doesn't support duplicate column names.

abhiagarwal01: Hmmm yeah, that's a big problem... based on what I just read in the source of pola-rs/polars#5057, they don't support duplicate keys either since annotations are assumed to be a `dict` per [python guidance on __annotations__](https://docs.python.org/3/howto/annotations.html). We don't write queries with duplicate columns by design, so would likely require some processing on our end.

mdesmet: On second thought I prefer to remove the annotations and first merge the attributes support. Can't you extract the information from `cursor.description` property?

abhiagarwal01: That's what we've been doing in the past and it's worked perfectly fine, just saw this PR and figured it might be decent to have it implemented directly in Trino :) but it seems like it may be more work than I first believe wrt handled duplicated column names

trino/client.py
```diff
@@ -1128,14 +1128,39 @@ def map(self, values: List[Any]) -> Optional[List[Any]]:
...
 
 
+class NamedRowTuple(tuple):
+    """Custom tuple class as namedtuple doesn't support missing or duplicate names"""
+    def __new__(cls, values, names: List[str], types: List[str]):
+        return super().__new__(cls, values)
+
+    def __init__(self, values, names: List[str], types: List[str]):
+        self._names = names
+        self.__annotations__ = dict()
```
hashhar, line 1138: I think a comment here/commit message about how annotations are meant to be used (or why they are useful) would be useful.

hashhar (approved):

hashhar:
@mdesmet Did you forget to push some changes? I don't see any after your comment.

Raghav-254:
Hi when this latest version 0.323.0 will be published in pypi.org?
Currently I am not able to see this version there.

hashhar:
I plan to do a release sometime this week. There's a few PRs we're waiting for.

Raghav-254:
Great
Currently at Linkedin we are using trino-python-client version 0.306.0 and for one of the column we are getting it's value as below
Screenshot 2023-04-10 at 3 49 33 PM

But now after upgrading it to 0.322.0, the output doesn't seem to be correct.
Screenshot 2023-04-10 at 3 45 43 PM

It seems that it is not handling the special characters properly. Can you please check this issue as well?

It is also giving the below error sometimes
`Unsupported UTF-8 sequence length when encoding string`

hashhar:
@Raghav-254 Could you file an issue with some reproduction steps? At-least having the table DDL (or at-least data-type for the problematic column) + some snippet of how you use the client would be helpful.

Raghav-254:
Hi @hashhar, thanks for responding so quickly.
I have opened up this issue: https://github.com/trinodb/trino-python-client/issues/355 including the steps to reproduce it.
Regards,
Raghav Mittal

Raghav-254:
Hi folks, can you please help me in one thing?
What could be the most efficient way to iterate over the NamedRowTuple which is there in this PR?
I have a usecase where I need access the values in the NamedRowTuple and do some processing over it.

PR #339, opened 2023-02-28
Support setting session properties on a individual statement

mdesmet:
## Description

Sometimes people want to set certain session properties on a single statement. This feature makes this much easier to manage as it avoids the state management involved with SET to the desired value and resetting back to the original value (which might not be the default value so RESET cannot be used).

Other DBMS have similar feature, eg. statement_params in Snowflake.

Some common use cases are setting different limits on query time or INSERT OVERWRITE in hive.

## Release notes

(*) Release notes are required, with the following suggested text:

```markdown
* Support setting session properties on individual statements
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,6 +373,29 @@ conn = trino.dbapi.connect(
 )
 ```
 
+## Session properties
+
+Session properties can be set on the connection
+
+```python
+import trino
+conn = trino.dbapi.connect(
+    ...,
+    session_properties={"query_max_run_time": "1d"}
+)
+```
+
+### Statement properties
+
+It's also possible to set a session property for a specific statement by setting it on the Cursor. This is especially handy in the case of hive partitions.
+
+```python
+import trino
+conn = trino.dbapi.connect()
+cur = conn.cursor(statement_properties={"hive.insert_existing_partitions_behavior": "OVERWRITE"})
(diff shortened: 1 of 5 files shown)


hashhar:
@ebyhr opinions? I'm not in favour of this since it makes it harder to reason about how session properties at connection and cursor level interact and what happens with cursor reuse.

Note that most other db-api clients follow the creating separate `Connection` per unique set of session properties mechanism.

ebyhr:
@hashhar I have the same opinion as you wrote. No strong opinion though.

mdesmet:
> I'm not in favour of this since it makes it harder to reason about how session properties at connection and cursor level interact and what happens with cursor reuse.

The counter argument is that this makes setting session properties for individual queries much easier for the end user. It also pretty transparent as the property is only valid for that cursor instance.

I don't think creating separate connections makes sense, I have the opposing view this actually adds complexity for the user. Also what if at some point we introduce transactions (Iceberg). It still makes sense to apply separate session properties onto a certain query within the same connection in that case. This is obviously already possible with SET and RESET session property. The advantage of this method is that the user doesn't have to keep track of the session property state (which is not necessarily the default when using RESET).

aalbu:
Trino has added syntactic support  for statement scoped session variables in https://github.com/trinodb/trino/pull/24889. Should we close this?

hashhar:
Yes.

PR #341, opened 2023-03-02
Add role parameter

damian3031:
Added role parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,6 +359,18 @@ conn = trino.dbapi.connect(
 )
 ```
 
+You could also pass `system` role without explicitly specifing "system" catalog:
+
+```python
+import trino
+conn = trino.dbapi.connect(
+    host='localhost',
+    port=443,
+    user='the-user',
+    roles="role1" # equivalent to {"system": "role1"}
+)
+```
+
 ## Timezone
 
 The time zone for the session can be explicitly set using the IANA time zone
--- a/trino/client.py
+++ b/trino/client.py
@@ -135,7 +135,7 @@ class ClientSession(object):
         transaction_id: str = None,
         extra_credential: List[Tuple[str, str]] = None,
         client_tags: List[str] = None,
-        roles: Dict[str, str] = None,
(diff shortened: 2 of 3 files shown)


trino/client.py
```diff
@@ -238,7 +239,13 @@ def timezone(self):
...
             return self._timezone
 
-    def _format_roles(self, roles):
+    def _format_roles(self, role, roles):
+        if role and roles:
+            raise ValueError("specify either 'role' or 'roles' parameter, but not both")
+        elif role:
+            roles = {"system": role}
+        elif role is None and roles is None:
+            return {}
```
mdesmet, line 248: I think this is no-op and can be removed.

hashhar (changes requested):
Why? Why not use the `roles` mechanism instead?

mdesmet:
> Why? Why not use the `roles` mechanism instead?

Because catalog roles are rather a legacy concept. Only hive standard SQL leverages the catalog roles. All modern access systems use a single SYSTEM role. Most users are not aware about the catalog roles concept.

Adding a single role will not confuse users but rather it will make the client more easy to use.

damian3031:
@hashhar @mdesmet @hovaesco I refactored PR to align with discussed approach - we could pass string with only role to `roles` parameter and it is transformed to system catalog role dict.

hashhar (approved):

README.md
```diff
@@ -359,6 +359,18 @@ conn = trino.dbapi.connect(
...
 
+You could also pass `system` role without explicitly specifing "system" catalog:
+
+```python
+import trino
+conn = trino.dbapi.connect(
+    host='localhost',
+    port=443,
+    user='the-user',
+    roles="roleC" # equivalent to {"system": "roleC"}
```
mdesmet, line 370: I would advise against a mixed case example unless we are full sure it works in all cases.

hashhar: good point.

hovaesco (approved):

hashhar (approved):

Issue #352, opened 2023-04-05
Installing trino[sqlalchemy] wont allow pandas.read_sql


GitToby:
### Expected behavior

Running a pandas.read_sql or similar fails with `AttributeError: 'OptionEngine' object has no attribute 'execute'`

### Actual behavior

Running a pandas.read_sql or similar should return dataframe with the table results. 

### Steps To Reproduce

install with `trino[sqlalchemy]` then run 

```python
engine = create_engine(
    "trino://my.instance:443/abcd",
    connect_args={
        "user": user_,
        "auth": BasicAuthentication(user_, pass_),
        "http_scheme": "https",
        "verify": "path/to/bundle.cer",
        # ...
    }
)
df = pd.read_sql("SELECT * FROM system.runtime.nodes", engine)
```

This is due to the optional dependancy not limiting the sqlalchemt version and hence falls victim to this bug: https://github.com/pandas-dev/pandas/issues/51015

A fix would be to limit the version to `sqlalchemy<2.0.0`

### Operating System

macos

### Trino Python client version

0.322.0

### Trino Server version

na

### Python version

3.11

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

GitToby:
a fix is to add the dependencies
- "trino>=0.322.0"
- "sqlalchemy<2.0.0"
after install this should push the sqlalchemy down and the pandas methods will work.

hashhar:
This is something that users of this package need to do. They can add explicit constraint for sqlalchemy. In case pandas doesn't work with sqlalchemy 2.x pandas should add the constraint and then pip would automatically resolve to a working version.

GitToby:
at the time of writing pandas 2.0 was released 3 days ago and is not really stable. I suppose this can be skipped internally to the Trino Python client, maybe documentation is a better approach for those who run into this issue.

PR #353, opened 2023-04-05
Update setup.py

GitToby:
## Description
Limit the sqlachemy version requires to sidestep pandas+sqlalchemy>2.0.0 issues with `pd.read_sql()`, `pd.read_sql_table()` and `pd.read_sql_query()`. See #352

## Non-technical explanation

Installing `trino[sqlalchemy]` installs a version that breaks methods such as `pd.read_sql()` and `pd.read_sql_table()`. This will limit the sqlalchemy version to sidestep these bugs.

## Release notes

```markdown
Limit the sqlachemy version requires to sidestep pandas+sqlalchemy>2.0.0 issues with `pd.read_sql()`, `pd.read_sql_table()` and `pd.read_sql_query()`. ({Installing trino[sqlalchemy] wont allow pandas.read_sql}`#352`)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ with open("trino/__init__.py", "rb") as f:
     version = str(ast.literal_eval(trino_version.group(1)))
 
 kerberos_require = ["requests_kerberos"]
-sqlalchemy_require = ["sqlalchemy >= 1.3"]
+sqlalchemy_require = ["sqlalchemy >= 1.3, <2.0.0"]
 external_authentication_token_cache_require = ["keyring"]
 
 # We don't add localstorage_require to all_require as users must explicitly opt in to use keyring.


mdesmet:
Hi @GitToby, we actually want to be compatible with sqlalchemy 2.0.0. What is exactly the issue that you are facing?

GitToby:
see issue #352 - ultimately this is an issue due to the bug in pandas where pandas 1.5.x doesn't work with sqlalchemy>2.0.0 because of a change in API. The default `trio[sqlalchemy]` install will hence fail when calling the pandas methods that leverage sqlalchemy tools.

hashhar:
This is something Pandas fixed in their 2.0 release. It should be improved upstream in Pandas (or by the user by pinning a specific version) in case of incompatibilities. This client itself works with SQLA 2.x fine so no changes need to be done here.

Issue #331, opened 2023-02-02
typeof(uuid) returns incorrecd result `varchar(32)` when using sqlalchemy 2.0


cpcloud:
### Expected behavior

I expected a result of `uuid`

### Actual behavior

I get a result of `varchar(32)`

### Steps To Reproduce

```
In [13]: import sqlalchemy as sa

In [14]: engine = sa.create_engine("trino://user@localhost/memory/default")
<ipython-input-14-456274705007>:1: SADeprecationWarning: The dbapi() classmethod on dialect classes has been renamed to import_dbapi().  Implement an import_dbapi() classmethod directly on class <class 'trino.sqlalchemy.dialect.TrinoDialect'> to remove this warning; the old .dbapi() classmethod may be maintained for backwards compatibility.
  engine = sa.create_engine("trino://user@localhost/memory/default")

In [15]: with engine.begin() as con:
    ...:     print(con.exec_driver_sql("SELECT typeof(uuid())").fetchone())
    ...:
('uuid',)

In [16]: with engine.begin() as con:
    ...:     print(con.execute(sa.select(sa.func.typeof(sa.func.uuid()))).fetchone())
    ...:
('uuid',)

In [17]: import uuid

In [18]: with engine.begin() as con:
    ...:     print(con.execute(sa.select(sa.func.typeof(sa.literal(uuid.uuid4())))).fetchone())
    ...:
('varchar(32)',)

In [19]: with engine.begin() as con:
    ...:     print(con.execute(sa.select(sa.func.typeof(sa.literal(uuid.uuid4(), type_=sa.UUID())))).fetchone())
    ...:
('varchar(32)',)

In [20]: sa.__version__
Out[20]: '2.0.1'

In [23]: import trino

In [24]: trino.__version__
Out[24]: '0.321.0'
```

### Operating System

NixOS

### Trino Python client version

0.321

### Trino Server version

406

### Python version

3.10.9

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

cpcloud:
Running with SQLAlchemy 1.4.46 I get:

```
In [6]: with engine.begin() as con:
   ...:     print(con.execute(sa.select(sa.func.typeof(sa.literal(uuid.uuid4())))).fetchone())
   ...:
('uuid',)
```

when using a `uuid.UUID` object.

mdesmet:
@cpcloud : Can you provide a PR for this issue? Let me know how I can assist you!

[...]

PR #354, opened 2023-04-06
Support UUID type

lpoulain:
## Description

Map Trino UUID type to Python's

## Non-technical explanation

Support Trino UUID type

## Release notes

(X) Release notes are required, with the following suggested text:

```markdown
* Maps Trino UUID type into Python's UUID
```

### DIFF
--- a/tests/integration/test_types_integration.py
+++ b/tests/integration/test_types_integration.py
@@ -1,5 +1,6 @@
 import math
 import re
+import uuid
 from datetime import date, datetime, time, timedelta, timezone, tzinfo
 from decimal import Decimal
 
@@ -774,7 +775,8 @@ def test_ipaddress(trino_connection):
 def test_uuid(trino_connection):
     SqlTest(trino_connection) \
         .add_field(sql="CAST(null AS UUID)", python=None) \
-        .add_field(sql="UUID '12151fd2-7586-11e9-8f9e-2a86e4085a59'", python='12151fd2-7586-11e9-8f9e-2a86e4085a59') \
+        .add_field(sql="UUID '12151fd2-7586-11e9-8f9e-2a86e4085a59'",
+                   python=uuid.UUID('12151fd2-7586-11e9-8f9e-2a86e4085a59')) \
         .execute()
 
 
--- a/trino/client.py
+++ b/trino/client.py
@@ -43,6 +43,7 @@ import random
 import re
 import threading
 import urllib.parse
+import uuid
 import warnings
 from dataclasses import dataclass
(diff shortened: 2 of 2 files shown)


trino/client.py
```diff
@@ -1185,6 +1186,14 @@ def map(self, values: Any) -> Optional[Dict[Any, Optional[Any]]]:
         }
 
 
+class UuidValueMapper(ValueMapper[uuid.UUID]):
+    def map(self, values: Any) -> Optional[uuid.UUID]:
+        try:
+            return uuid.UUID(values)
+        except:
+            return None
```
mdesmet, line 1194: Why the exception handling,  I don't think we would like to silently discard values.

```suggestion
        return uuid.UUID(values)
```

mdesmet (commented):
Missing prepared statement handling

trino/client.py
lpoulain: The test currently expects that `CAST(NULL AS UUID)` returns `None` in Python. Should we expect a different behavior or a specific case should be applied when `values` is `None`?

tests/integration/test_types_integration.py
```diff
@@ -1,5 +1,6 @@
 import math
 import re
+import uuid
```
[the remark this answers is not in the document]
lpoulain: The reason why I did this is because I saw elsewhere in the code was importing uuid instead of `uuid.UUID`. But I can change this.

trino/client.py
mdesmet: I get it but we should not let `None` go into UUID, similar like all the other mappers.

trino/client.py
```diff
@@ -1185,6 +1186,13 @@ def map(self, values: Any) -> Optional[Dict[Any, Optional[Any]]]:
         }
 
 
+class UuidValueMapper(ValueMapper[uuid.UUID]):
+    def map(self, values: Any) -> Optional[uuid.UUID]:
```
mdesmet, line 1190: change into `value` instead of `values`

mdesmet (commented):
LGTM % minor comment

trino/client.py

mdesmet (approved):

mdesmet (commented):
Please move all sqlalchemy related changes to a separate commit. Also I don't see any tests related to that change.

hashhar (approved):

hashhar:
does this work with sqla too? If yes then please add tests. Otherwise looks good to go.

lpoulain:
It does not work yet with sqla. And this does not fix #331 which seems to be a sqla issue (I've added comments to this issue to detail what I've found)

PR #359, opened 2023-04-14
Support UUID for SQLAlchemy

lpoulain:
## Description

Provide a SQLAlchemy mapping to Trino's UUID type (resolves issue #331 )

## Non-technical explanation

Provide UUID support for SQLAlchemy

## Release notes

(X) Release notes are required, with the following suggested text:

```markdown
* Support UUID for SQLAlchemy ({issue}`358`)
```

### DIFF
--- a/tests/integration/test_sqlalchemy_integration.py
+++ b/tests/integration/test_sqlalchemy_integration.py
@@ -9,6 +9,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License
+import uuid
+
 import pytest
 import sqlalchemy as sqla
 from sqlalchemy.sql import and_, not_, or_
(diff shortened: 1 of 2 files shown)


trino/sqlalchemy/datatype.py
```diff
@@ -86,6 +86,11 @@ def get_col_spec(self, **kw):
         return 'JSON'
 
 
+class UUID(UserDefinedType):
```
hovaesco, line 89: please add `__visit_name__`

hovaesco (approved):

trino/sqlalchemy/datatype.py
lpoulain: Done

tests/integration/test_sqlalchemy_integration.py
```diff
@@ -89,13 +91,15 @@ def test_define_and_create_table(trino_connection):
         sqla.Table('users',
                    metadata,
                    sqla.Column('id', sqla.Integer),
+                   sqla.Column('guid', UUID),
```
mdesmet, line 94: I think this should be sqla.Uuid instead. See https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Uuid

mdesmet: Uuid only works in v2 of sqlalchemy. So you will need to ensure it only works in that version.

trino/sqlalchemy/datatype.py
```diff
@@ -86,6 +86,13 @@ def get_col_spec(self, **kw):
...
+    __visit_name__ = "UUID"
+
+    def get_col_spec(self, **kw):
+        return "UUID"
+
+
```
mdesmet, line 95: I think this code block should be removed as we don't need to add a custom `UserDefinedType`

trino/sqlalchemy/datatype.py
```diff
@@ -122,7 +129,7 @@ def get_col_spec(self, **kw):
     #
     # === Mixed ===
     # 'ipaddress': IPADDRESS
-    # 'uuid': UUID,
+    "uuid": UUID,
```
mdesmet, line 132: This should be mapped to the sqlalchemy uuid type if it's available

trino/sqlalchemy/compiler.py
```diff
@@ -196,6 +196,9 @@ def visit_TIME(self, type_, **kw):
     def visit_JSON(self, type_, **kw):
         return 'JSON'
 
+    def visit_UUID(self, type, **kw):
+        return "UUID"
+
```
mdesmet, line 201: This should only be added if the uuid type is available.

tests/unit/sqlalchemy/test_datatype_parse.py
```diff
@@ -192,3 +192,8 @@ def test_parse_row(type_str: str, sql_type: ARRAY, assert_sqltype):
 def test_parse_datetime(type_str: str, sql_type: TypeEngine, assert_sqltype):
     actual_type = datatype.parse_sqltype(type_str)
     assert_sqltype(actual_type, sql_type)
+
+
+def test_parse_uuid():
+    actual_type = datatype.parse_sqltype("UUID")
+    assert type(actual_type) == UUID
```
mdesmet, line 199: Test should only be succesful if type is available.

tests/integration/test_sqlalchemy_integration.py
```diff
@@ -133,6 +135,60 @@ def test_insert(trino_connection):
         metadata.drop_all(engine)
 
 
+@pytest.mark.skipif(
+    sqlalchemy_version() < "2.0",
+    reason="columns argument to select() must be a Python list or other iterable"
```
mdesmet, line 140: reason seems not correct.

mdesmet (commented):
LGTM except reason

trino/sqlalchemy/datatype.py
```diff
@@ -122,13 +123,16 @@ def get_col_spec(self, **kw):
     #
     # === Mixed ===
     # 'ipaddress': IPADDRESS
-    # 'uuid': UUID,
```
mdesmet, line 125: Can you revert this change

mdesmet:
Please also add that this resolves https://github.com/trinodb/trino-python-client/issues/331 in the PR description.

mdesmet (approved):

hashhar (approved):

PR #365, opened 2023-04-28
Perform prepared statements on client-side

mdesmet:
## Release notes

```markdown
```

### DIFF
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -373,10 +373,11 @@ class Cursor(object):
         :param statement: sql to be executed.
         :param name: name that will be assigned to the prepared statement.
         """
-        sql = f"PREPARE {name} FROM {statement}"
-        query = trino.client.TrinoQuery(self.connection._create_request(), query=sql,
-                                        legacy_primitive_types=self._legacy_primitive_types)
-        query.execute()
+        # sql = f"PREPARE {name} FROM {statement}"
+        # query = trino.client.TrinoQuery(self.connection._create_request(), query=sql,
+        #                                 legacy_primitive_types=self._legacy_primitive_types)
+        # query.execute()
+        self.connection._client_session.prepared_statements[name] = statement
 
     def _execute_prepared_statement(
         self,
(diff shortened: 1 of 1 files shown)


martint:
Please provide more details about the purpose and motivation for this change.

mdesmet:
> Please provide more details about the purpose and motivation for this change.

This is just a PoC for reducing the amount of http calls involved in prepared statement handling. 

This PR was to demonstrate that this approach conceals invalid SQL as HTTP errors.

`trino.exceptions.HttpError: error 400: b"Invalid X-Trino-Prepared-Statement header: line 2:22: mismatched input 'FRMO'. Expecting: ',', 'EXCEPT', 'FETCH', 'FROM', 'GROUP', 'HAVING', 'INTERSECT', 'LIMIT', 'OFFSET', 'ORDER', 'UNION', 'WHERE', 'WINDOW', <EOF>"`

trino/dbapi.py
```diff
@@ -373,10 +373,11 @@ def _prepare_statement(self, statement: str, name: str) -> None:
...
         """
-        sql = f"PREPARE {name} FROM {statement}"
-        query = trino.client.TrinoQuery(self.connection._create_request(), query=sql,
-                                        legacy_primitive_types=self._legacy_primitive_types)
-        query.execute()
+        # sql = f"PREPARE {name} FROM {statement}"
+        # query = trino.client.TrinoQuery(self.connection._create_request(), query=sql,
+        #                                 legacy_primitive_types=self._legacy_primitive_types)
+        # query.execute()
+        self.connection._client_session.prepared_statements[name] = statement
```
hashhar, line 380: if we want to reduce http calls the alternative is to implement client side parameter handling and not use server side prepared statements at all (which I'm not a fan of).

An option is to remove the `DEALLOCATE` call and do a deallocate for all tracked statements during connection close. That has the benefit that you can avoid a `PREPARE` as well for already prepared statements (since we know the already prepared statements on the client side).

However this entire thing feels like a workaround.

lpoulain: I know this is Python, but is there some concurrency risk by modifying `_client_session`?

Issue #366, opened 2023-04-28
Timestamp with time zone not mapped correctly


john-bodley:
### Expected behavior

Per the [Trino documentation](https://trino.io/docs/current/functions/datetime.html#time-zone-conversion) the Trino DB-API should correctly map timestamps with time zone to Python `datetime` objects.

### Actual behavior

Per the example provided in the Trino documentation, 

```python
from trino.dbapi import connect

cursor = connect(host="localhost").cursor()
cursor.execute("SELECT timestamp '2012-10-31 01:00 UTC' AT TIME ZONE 'America/Los_Angeles'")
rows = cursor.fetchall()
print(rows)
```

returns 

```
[[datetime.datetime(2012, 10, 30, 18, 0, tzinfo=<DstTzInfo 'America/Los_Angeles' LMT-1 day, 16:07:00 STD>)]]
```

### Steps To Reproduce

See above.

### Log output

The root cause is the `TimeWithTimeZoneValueMapper` which incorrectly maps the Trino result '2012-10-30 18:00:00 America/Los_Angeles'.. The current implementation encodes the timezone incorrectly, i.e., per [here](https://stackoverflow.com/a/50613134),

> "A pytz timezone class does not represent a single offset from UTC, it represents a geographical area ...

thus [this](https://github.com/trinodb/trino-python-client/blob/master/trino/client.py#L1076) line, 

```python
time.fromisoformat(whole_python_temporal_value).replace(tzinfo=_create_tzinfo(timezone_part))
```

returns `2012-10-30 18:00:00-07:53` with a 07:53 as opposed to an 07:00 offset. The correct implementation is to adhere to the suggestion in the linked StackOverflow post, i.e., 

```python
_create_tzinfo(timezone_part).localize(datetime.fromisoformat(whole_python_temporal_value)
```

which returns  the `2012-10-30 18:00:00-07:00` output.

### Operating System

Ubuntu

### Trino Python client version

0.321.0

### Trino Server version

393.0

### Python version

3.9.16

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

hashhar:
Fixed via https://github.com/trinodb/trino-python-client/pull/368

PR #367, opened 2023-04-28
fix(dbapi): Address timestamp with time zone localization issue

john-bodley:
## Description

This PR addresses https://github.com/trinodb/trino-python-client/issues/366.

Note @mdesmet et al. it doesn't address the time zone conundrum with [fractional times](https://github.com/trinodb/trino-python-client/blob/226b216c946025a19560e51f4794daa027963ba0/trino/client.py#L1017) or times (as opposed to timestamps). Additionally I sense there's an issue with how the [`TimeWithTimeZoneValueMapper`](https://github.com/trinodb/trino-python-client/blob/226b216c946025a19560e51f4794daa027963ba0/trino/client.py#L1068-L1078) works (or doesn't work). 

Specifically, the following Trino query, 

```sql
SELECT TIME '01:00' AT TIME ZONE 'America/Los_Angeles'
```

returns `18:00:00-07:00` when run on 2023-04-28 when daylights savings is in effect for said time zone. In order to resolve this query (I presume) the Trino server has had to taken has taken into account the current date on the server to "anchor" the time, i.e., if one were to run the same query on 2023-01-01 when daylights savings in not in effect the result would be `17:00:00-08:00`. 

The issue is the mapping that @mdesmet outlined is running on the Trino client—which is pseudo agnostic* of the server's timezone—and may be in a different timezone (and thus date) than the Trino server. The problem seems ill-defined, i.e., when mapping this to a Python object localized to a timezone without a numerical offset, i.e., America/Los_Angeles, one first needs to know the underlying date. 

\* Granted the client isn't completely agnostic of the current date on the server, i.e., it can be determined via `SELECT current_date`, however this requires an additional query and there's no guarantee this will be executed on the same date as the primary query.

## Release notes

```markdown
* Fix localization of timestamp with time zone. ({[366](https://github.com/trinodb/trino-python-client/issues/366)})
```

### DIFF
--- a/tests/integration/test_types_integration.py
+++ b/tests/integration/test_types_integration.py
@@ -283,14 +283,19 @@ def test_time(trino_connection):
     ).execute()
 
 
-@pytest.mark.skipif(trino_version() == '351', reason="time not rounded correctly in older Trino versions")
-def test_time_with_timezone(trino_connection):
-    query_time_with_timezone(trino_connection, '-08:00')
-    query_time_with_timezone(trino_connection, '+08:00')
-    query_time_with_timezone(trino_connection, '+05:30')
-
-
-def query_time_with_timezone(trino_connection, tz_str: str):
+@pytest.mark.skipif(
+    trino_version() == '351',
+    reason="time not rounded correctly in older Trino versions",
+)
+@pytest.mark.parametrize(
+    'tz_str',
+    [
+        '-08:00',
+        '+08:00',
+        '+05:30',
+    ]
+)
+def test_time_with_timezone(trino_connection, tz_str: str):
     tz = create_timezone(tz_str)
     (
         SqlTest(trino_connection)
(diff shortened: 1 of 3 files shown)


tests/integration/test_types_integration.py
```diff
@@ -569,133 +579,129 @@ def query_timestamp_with_timezone(trino_connection, tz_str):
...
         .add_field(
             sql=f"TIMESTAMP '2001-08-22 23:59:59.9999999 {tz_str}'",
-            python=datetime(2001, 8, 23, 0, 0, 0, 0).replace(tzinfo=tz))
+            python=_localize(datetime(2001, 8, 23, 0, 0, 0, 0), tz_str))
         .add_field(
             sql=f"TIMESTAMP '2001-08-22 23:59:59.999999999 {tz_str}'",
-            python=datetime(2001, 8, 23, 0, 0, 0, 0).replace(tzinfo=tz))
-        # ce
-        .add_field(
-            sql=f"TIMESTAMP '0001-01-01 01:23:45.123 {tz_str}'",
```
john-bodley, line 685: This is a non-representative date for the US/Eastern and Asia/Kolkata time zones.

john-bodley:
Per [this](https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html) article I wonder if there's merit in removing `pytz` in favor of `dateutil.tz` which since Python 3.6 remedies ambiguous datetimes due to daylight saving time transition.

john-bodley:
Closing in favor of https://github.com/trinodb/trino-python-client/pull/368.

PR #368, opened 2023-04-30
Address time zone localization issue

john-bodley:
## Description

This PR addresses https://github.com/trinodb/trino-python-client/issues/366. 

Per [this](https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html) article, there seemed to be merit in removing `pytz` in favor of `dateutil.tz` given that in Python 3.6 the `datetime` module remedied the ambiguous datetimes due to daylight saving time transition. 

Regrettably though one cannot obtain the IANA name from a `dateutil.tz` time zone—which is required when transpiling `datetime.time`/`datetime.datetime` objects to Trino SQL when provided as operation parameters. Instead we use the `zoneinfo` package (added to Python’s standard library in Python 3.9 and back ported) which does provide a mechanism of obtaining the specified IANA name from a `datetime.datetime`/`datetime.time` object, i.e., 

```python
>>> from datetime import time
>>> from zoneinfo import ZoneInfo
>>>
>>> midnight = time(0, tzinfo=ZoneInfo("America/Los_Angeles"))
>>> midnight.tzinfo.key 
'America/Los_Angeles'
```

The fix is rather simple as simply switching out the `pytz` package for the `zoneinfo` package remedies the issue:

```
>>> from datetime import datetime
>>> import pytz
>>>
>>> datetime(2023, 1, 1, tzinfo=pytz.timezone("America/Los_Angeles")).isoformat()
'2023-01-01T00:00:00-07:53' 
```

```
>>> from datetime import datetime
>>> import zoneinfo
>>>
>>> datetime(2023, 1, 1, tzinfo=zoneinfo.ZoneInfo("America/Los_Angeles")).isoformat()
'2023-01-01T00:00:00-08:00'
```

## Release notes

```markdown
* Fixes time zone localization. The `pytz` package for encoding/decoding time zones has been deprecated in favor of `zoneinfo`. Time zones should now be encoded (when specifying execution parameters) using the `zoneinfo.ZoneInfo` method as opposed to `pytz.timezone` method. ({[issue](https://github.com/trinodb/trino-python-client/issues/366.)}`366`)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,12 @@ setup(
         "Topic :: Database :: Front-Ends",
     ],
     python_requires='>=3.7',
-    install_requires=["pytz", "requests", "tzlocal"],
+    install_requires=[
+        "backports.zoneinfo;python_version<'3.9'",
+        "pytz",
+        "requests",
+        "tzlocal",
+    ],
     extras_require={
         "all": all_require,
         "kerberos": kerberos_require,
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,7 @@ setup(
     python_requires='>=3.7',
     install_requires=[
         "backports.zoneinfo;python_version<'3.9'",
+        "python-dateutil",
         "pytz",
         "requests",
         "tzlocal",
--- a/tests/integration/test_types_integration.py
+++ b/tests/integration/test_types_integration.py
@@ -4,8 +4,12 @@ import uuid
 from datetime import date, datetime, time, timedelta, timezone, tzinfo
 from decimal import Decimal
 
+try:
+    from zoneinfo import ZoneInfo
+except ModuleNotFoundError:
+    from backports.zoneinfo import ZoneInfo
+
 import pytest
-import pytz
 
 import trino
 from tests.integration.conftest import trino_version
@@ -729,7 +733,7 @@ def create_timezone(timezone_str: str) -> tzinfo:
         else:
             return timezone(-timedelta(hours=hours, minutes=minutes))
     else:
-        return pytz.timezone(timezone_str)
+        return ZoneInfo(timezone_str)
 
 
 def test_interval(trino_connection):
--- a/tests/integration/test_types_integration.py
+++ b/tests/integration/test_types_integration.py
@@ -302,114 +302,114 @@ def test_time_with_timezone(trino_connection, tz_str: str):
         # min supported time(3)
         .add_field(
             sql=f"TIME '00:00:00 {tz_str}'",
-            python=time(0, 0, 0).replace(tzinfo=tz))
+            python=time(0, 0, 0, tzinfo=tz))
         .add_field(
             sql=f"TIME '12:34:56.123 {tz_str}'",
-            python=time(12, 34, 56, 123000).replace(tzinfo=tz))
+            python=time(12, 34, 56, 123000, tzinfo=tz))
         # max supported time(3)
         .add_field(
             sql=f"TIME '23:59:59.999 {tz_str}'",
-            python=time(23, 59, 59, 999000).replace(tzinfo=tz))
+            python=time(23, 59, 59, 999000, tzinfo=tz))
         # min value for each precision
         .add_field(
             sql=f"TIME '00:00:00 {tz_str}'",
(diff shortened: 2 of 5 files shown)


tests/integration/test_types_integration.py
```diff
@@ -284,13 +288,15 @@ def test_time(trino_connection):
...
 
 @pytest.mark.skipif(trino_version() == '351', reason="time not rounded correctly in older Trino versions")
-def test_time_with_timezone(trino_connection):
-    query_time_with_timezone(trino_connection, '-08:00')
-    query_time_with_timezone(trino_connection, '+08:00')
-    query_time_with_timezone(trino_connection, '+05:30')
-
-
-def query_time_with_timezone(trino_connection, tz_str: str):
+@pytest.mark.parametrize(
```
john-bodley, line 291: Same logic as before but more aligned with how `pytest` was intended to be used.

trino/client.py
```diff
@@ -51,21 +51,18 @@
...
 from tzlocal import get_localzone_name  # type: ignore
 
 import trino.logging
 from trino import constants, exceptions
 
-try:
-    from zoneinfo import ZoneInfo  # type: ignore
-
-except ModuleNotFoundError:
-    from backports.zoneinfo import ZoneInfo  # type: ignore
```
john-bodley, line 66: Previously there was no requirement in `setup.py` for the `backports.zoneinfo` package.

tests/integration/test_types_integration.py
```diff
@@ -300,114 +306,114 @@ def query_time_with_timezone(trino_connection, tz_str: str):
         # min supported time(3)
         .add_field(
             sql=f"TIME '00:00:00 {tz_str}'",
-            python=time(0, 0, 0).replace(tzinfo=tz))
+            python=time(0, 0, 0, tzinfo=tz))
```
john-bodley, line 309: Clearer/cleaner to specify the time zone when creating the object.

mdesmet (approved):
LGTM

john-bodley:
@mdesmet I've addressed your comments.

hashhar:
I'll take a look tomorrow. I'm leaving some quick editorial comments for now.

Since this is a correctness issue I'd like to do a release as soon as we have this merged.

hashhar (commented):
Commit "[preparatory] Use @pytest.mark.parameterize".

The project avoids prefixing commit messages, so something just like `Parameterize time and timestamp with time zone tests` is good enough.

tests/integration/test_types_integration.py
hashhar: are there any other occurrences of this in production code as well?

hashhar (commented):
commit "[preparatory] Specify tzinfo during datetime/time object creation" -> "Specify tzinfo during datetime/time creation in tests"

setup.py
```diff
@@ -78,7 +78,12 @@
         "Topic :: Database :: Front-Ends",
     ],
     python_requires='>=3.7',
-    install_requires=["pytz", "requests", "tzlocal"],
+    install_requires=[
+        "backports.zoneinfo;python_version<'3.9'",
```
hashhar, line 82: missing zoneinfo require should be a separate commit, it's not related to the fix itself.

trino/client.py
```diff
@@ -1041,7 +1037,7 @@ def new_instance(self, value: datetime, fraction: Decimal) -> TimestampWithTimeZ
         return TimestampWithTimeZone(value, fraction)
 
     def normalize(self, value: datetime) -> datetime:
-        if isinstance(self._whole_python_temporal_value.tzinfo, BaseTzInfo):
+        if tz.datetime_ambiguous(value):
```
hashhar, line 1040: nice.

hashhar (commented):
commit " [fix] Address timezone localization issue ".

Can you add a short explanation in the commit message body about the issue? Alternatively I think it'd also be self-explanatory if you can add a test which showcases the issue and then in next commit add the fix and adjust the test.

john-bodley:
@hashhar I've addressed your comments.

hashhar (approved):

PR #371, opened 2023-05-07
Add lazy evaluation of `server_version_info`

mdesmet:
## Release notes

```markdown
```

### DIFF
--- a/tests/integration/test_sqlalchemy_integration.py
+++ b/tests/integration/test_sqlalchemy_integration.py
@@ -15,6 +15,7 @@ import pytest
 import sqlalchemy as sqla
 from sqlalchemy.sql import and_, not_, or_
 
+from tests.integration.conftest import trino_version
 from tests.unit.conftest import sqlalchemy_version
 from trino.sqlalchemy.datatype import JSON
 
@@ -497,3 +498,24 @@ def test_get_view_names_raises(trino_connection):
 
     with pytest.raises(sqla.exc.NoSuchTableError):
         sqla.inspect(engine).get_view_names(None)
+
+
+@pytest.mark.parametrize('trino_connection', ['system'], indirect=True)
+@pytest.mark.skipif(trino_version() == '351', reason="version() not supported in older Trino versions")
+def test_version_is_lazy(trino_connection):
+    _, conn = trino_connection
+    result = conn.execute(sqla.text("SELECT 1"))
+    result.fetchall()
+    num_queries = _num_queries_containing_string(conn, "SELECT version()")
+    assert num_queries == 0
+    version_info = conn.dialect.server_version_info
+    assert isinstance(version_info, tuple)
+    num_queries = _num_queries_containing_string(conn, "SELECT version()")
+    assert num_queries == 1
+
+
+def _num_queries_containing_string(connection, query_string):
+    statement = sqla.text("select query from system.runtime.queries order by query_id desc offset 1 limit 1")
+    result = connection.execute(statement)
+    rows = result.fetchall()
+    return len(list(filter(lambda rec: query_string in rec[0], rows)))
--- a/trino/sqlalchemy/dialect.py
+++ b/trino/sqlalchemy/dialect.py
@@ -336,15 +336,20 @@ class TrinoDialect(DefaultDialect):
         """Trino has no support for sequence. Returns False indicate that given sequence does not exists."""
         return False
 
-    def _get_server_version_info(self, connection: Connection) -> Any:
-        query = "SELECT version()"
-        try:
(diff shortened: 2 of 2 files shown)


aalbu (commented):
That's cool.  Have you checked whether the getter is ever called in the process of creating a connection/running some SQL/closing the connection?  It seemed to me like the property is never accessed, but I'm not positive.

hashhar:
This is something that the SQLA engine interface provides so user scripts/programs can make use of it. Core itself doesn't use it other than for populating a `server_version_info` attribute.

Some dialects also use it internally for version dependent logic, an example is Oracle - https://github.com/zzzeek/sqlalchemy/blob/671647c8574bd6ff4c39fe503a2b1958a802772d/lib/sqlalchemy/dialects/oracle/base.py#L1487

hashhar:
How much does this really help, it's only called once per connection. So only if someone is using `NullPool` or creating new `engine` everytime.

tests/integration/test_sqlalchemy_integration.py
```diff
@@ -497,3 +497,23 @@ def test_get_view_names_raises(trino_connection):
...
+    result.fetchall()
+    num_queries = _num_queries_containing_string(conn, "SELECT version()")
+    assert num_queries == 0
+    version_info = conn.dialect.server_version_info
+    assert isinstance(version_info, tuple)
+    num_queries = _num_queries_containing_string(conn, "SELECT version()")
+    assert num_queries == 1
+
+
+def _num_queries_containing_string(connection, query_string, num=3):
```
hashhar, line 515: no default limits would be safer IMO.

trino/sqlalchemy/dialect.py
```diff
@@ -336,15 +336,20 @@ def has_sequence(self, connection: Connection, sequence_name: str, schema: str =
...
+            query = "SELECT version()"
+            try:
+                res = connection.execute(sql.text(query))
+                version = res.scalar()
+                return tuple([version])
+            except exc.ProgrammingError as e:
+                logger.debug(f"Failed to get server version: {e.orig.message}")
+                return None
+
+        # We make the server_version_info be evaluated lazily
```
hashhar, line 351: what is obvious - the why needs to be commented in detail.

hashhar (commented):
LGTM % comment.

hashhar:
> Well, it is called once per connection and not used in 99.999% of the cases, so making it lazily evaluated makes definitely a difference as it removes these series of requests from EVERY sqlalchemy usage.

I was thinking in terms of number of requests.
This function isn't executed per-query, so it only reduces 1 call for entire application/connection (which is generally much lower than number of queries).

i.e. for someone who uses only a few connections there is no benefit of this lazy evaluation.

trino/sqlalchemy/dialect.py
```diff
@@ -336,15 +336,20 @@ def has_sequence(self, connection: Connection, sequence_name: str, schema: str =
...
+        cls.server_version_info = property(get_server_version_info, lambda instance, value: None)
```
nineinchnick, line 352: Why not use the decorator `@property`?

nineinchnick:
When working with a server that's geographically far away, or over a bad network connection, the overhead of the extra query can be significant, especially when executing short/simple queries. IMO, it is very valuable to avoid this unnecessary query.

nineinchnick:
Note that I tested today that getting the server version info is done only once after creating the engine, even if connections are not reused (using a `NullPool` and closing the connection after every query).

trino/sqlalchemy/dialect.py
mdesmet: Because `server_version_info` is not in our codebase but in sqlalchemy's, so we need to resort to this type of instantiation.

nineinchnick: I mean why use the `property()` function instead of a decorator.

mdesmet: Have a look at [sqlalchemy's code](https://github.com/sqlalchemy/sqlalchemy/blob/eb286c15f096771dbb128acbe8fe03e94aa72f6a/lib/sqlalchemy/engine/default.py#L514). Simply adding the property wouldn't work as it would be overwritten.

nineinchnick: Thanks, I get it know. But it seems we're doing a hack here where we override the server info property with one that has a getter for lazy evaluation, but a noop setter. Then we return a None, which sqlalchemy tries to assign to that prop, but that'll get ignored because go the setter. Is it worth adding a comment explaining this, if I got it right?

hashhar (approved):

PR #375, opened 2023-05-17
Use `EXECUTE IMMEDIATE` for parameterized statement executions

aalbu:
## Description

Make parametrized statement executions more efficient by taking advantage of https://github.com/trinodb/trino/pull/17341.

When running against older versions of Trino that don't support the `EXECUTE IMMEDIATE` statement, the old behavior can be restored by passing `legacy_prepared_statements=True` when creating a connection.

Draft PR until the back-end change is released.

## Release notes

(X) Release notes are required, please propose a release note for me.

```markdown
```

### DIFF
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -21,7 +21,10 @@ import binascii
 import datetime
 import math
 import uuid
+from collections import OrderedDict
 from decimal import Decimal
+from threading import Lock
+from time import time
 from typing import Any, Dict, List, NamedTuple, Optional  # NOQA for mypy types
 from urllib.parse import urlparse
 
@@ -78,6 +81,41 @@ paramstyle = "qmark"
 logger = trino.logging.get_logger(__name__)
 
 
+class TimeBoundLRUCache:
+    """A bounded LRU cache which expires entries after a configured number of seconds.
+    Note that expired entries will be evicted only on an attempted access (or through
+    the LRU policy)."""
+    def __init__(self, capacity: int, ttl_seconds: int):
+        self.capacity = capacity
+        self.ttl_seconds = ttl_seconds
+        self.cache = OrderedDict()
+        self.lock = Lock()
+
+    def get(self, key):
+        with self.lock:
+            if key not in self.cache:
+                return None
+            value, timestamp = self.cache[key]
+            if time() - timestamp > self.ttl_seconds:
+                self.cache.pop(key)
+                return None
+            self.cache.move_to_end(key)
+            return value
+
+    def put(self, key, value):
+        with self.lock:
+            self.cache[key] = value, time()
+            self.cache.move_to_end(key)
+            if len(self.cache) > self.capacity:
+                self.cache.popitem(last=False)
+
+    def __repr__(self):
+        return f"LRUCache(capacity: {self.capacity}, ttl: {self.ttl_seconds} seconds, {self.cache})"
+
+
+must_use_legacy_prepared_statements = TimeBoundLRUCache(1024, 3600)
+
+
 def connect(*args, **kwargs):
     """Constructor for creating a connection to the database.
 
@@ -117,6 +155,7 @@ class Connection(object):
         http_session=None,
         client_tags=None,
         legacy_primitive_types=False,
+        legacy_prepared_statements=None,
         roles=None,
         timezone=None,
     ):
@@ -162,6 +201,7 @@ class Connection(object):
         self._request = None
         self._transaction = None
         self.legacy_primitive_types = legacy_primitive_types
+        self.legacy_prepared_statements = legacy_prepared_statements
 
     @property
     def isolation_level(self):
@@ -228,9 +268,28 @@ class Connection(object):
         return Cursor(
             self,
             request,
-            # if legacy_primitive_types is not explicitly set in Cursor, take from Connection
+            # if legacy params are not explicitly set in Cursor, take them from Connection
             legacy_primitive_types if legacy_primitive_types is not None else self.legacy_primitive_types
         )
+
+    def _use_legacy_prepared_statements(self):
+        if self.legacy_prepared_statements is not None:
+            return self.legacy_prepared_statements
+
+        value = must_use_legacy_prepared_statements.get((self.host, self.port))
+        if value is None:
+            try:
+                query = trino.client.TrinoQuery(
+                    self._create_request(),
+                    query="EXECUTE IMMEDIATE 'SELECT 1'")
+                query.execute()
+            except Exception as e:
+                logger.warning(
+                    "EXECUTE IMMEDIATE not available for %s:%s; defaulting to legacy prepared statements (%s)",
+                    self.host, self.port, e)
+                value = True
+                must_use_legacy_prepared_statements.put((self.host, self.port), value)
+        return value
 
 
 class DescribeOutput(NamedTuple):
@@ -280,7 +339,11 @@ class Cursor(object):
 
     """
 
-    def __init__(self, connection, request, legacy_primitive_types: bool = False):
+    def __init__(
+            self,
+            connection,
+            request,
+            legacy_primitive_types: bool = False):
         if not isinstance(connection, Connection):
             raise ValueError(
                 "connection must be a Connection object: {}".format(type(connection))
@@ -391,6 +454,18 @@ class Cursor(object):
         sql = 'EXECUTE ' + statement_name + ' USING ' + ','.join(map(self._format_prepared_param, params))
         return trino.client.TrinoQuery(self._request, query=sql, legacy_primitive_types=self._legacy_primitive_types)
 
+    def _execute_immediate_statement(self, statement: str, params):
+        """
+        Binds parameters and executes a statement in one call.
+
+        :param statement: sql to be executed.
+        :param params: parameters to be bound.
+        """
+        sql = "EXECUTE IMMEDIATE '" + statement.replace("'", "''") + \
+              "' USING " + ",".join(map(self._format_prepared_param, params))
+        return trino.client.TrinoQuery(
+            self.connection._create_request(), query=sql, legacy_primitive_types=self._legacy_primitive_types)
+
     def _format_prepared_param(self, param):
         """
         Formats parameters to be passed in an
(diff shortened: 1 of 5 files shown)


hashhar:
> When running against older versions of Trino that don't support the EXECUTE IMMEDIATE statement, the old behavior can be restored by passing legacy_prepared_statements=True when creating a connection.

Since client knows server version can we do this automatically instead of user having to opt-into it? We should only have the toggle so that if someone wants to manually revert to old PREPARE + EXECUTE + DEALLOCATE they can do so.
The default behaviour should be to auto-cofigured.

cc: @findepi is this reasonable?

trino/dbapi.py
```diff
@@ -391,6 +395,16 @@ def _execute_prepared_statement(
...
         return trino.client.TrinoQuery(self._request, query=sql, legacy_primitive_types=self._legacy_primitive_types)
 
+    def _execute_immediate_statement(self, statement: str, params):
+        """
+        Binds parameters and executes a statement in one call.
+
+        :param statement: sql to be executed.
+        :param params: parameters to be bound.
+        """
+        sql = 'EXECUTE IMMEDIATE ' + self._format_prepared_param(statement) + ' USING ' + ','.join(map(self._format_prepared_param, params))
```
hashhar, line 405: why do you call `_format_prepared_param` on the `statement`? Shouldn't it just be `statement.replace("'", "''")`?

hashhar (commented):
do we have existing tests with query text which needs escaping? Since EXECUTE IMMEDIATE accepts query text as a string literal and hence quotes need to be escaped.

hovaesco:
>Since client knows server version can we do this automatically instead of user having to opt-into it?

How client knows server version? I assume client need to execute `SELECT version()` to know it.

aalbu:
> do we have existing tests with query text which needs escaping? Since EXECUTE IMMEDIATE accepts query text as a string literal and hence quotes need to be escaped.

We do: https://github.com/trinodb/trino-python-client/blob/master/tests/integration/test_dbapi_integration.py#L107-L120

trino/dbapi.py
aalbu: I just wanted to make sure I am using the same logic for formatting string literals as elsewhere.  I realize now it's just escaping single quotes, but I was looking at https://github.com/trinodb/trino/blob/4ab52e4a5e2abd31d7b28aa506dd788e11f62091/core/trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java#L1054 and was thinking that maybe at one point we'll augment the Python client to handle Unicode, as well.

hovaesco:
Some checks are failing, please run `pre-commit install` to run these checks locally before commiting.

trino/dbapi.py
hashhar: hmmm, I don't think we have tests for unicode (or byte-array) literals. But I think it's the server who "knows" that a `U&` is a unicode literal - client passes it as a plain string in other clients too as well.

aalbu: My impression was that Unicode characters needed to be encoded as `U&` in string literals, but I guess I was wrong.

mdesmet (approved):

findepi:
> Since client knows server version can we do this automatically instead of user having to opt-into it? We should only have the toggle so that if someone wants to manually revert to old PREPARE + EXECUTE + DEALLOCATE they can do so.
The default behaviour should be to auto-cofigured.

I agree that -- if this feature is important -- it should auto-configure.
People should typically be expected not to know to turn a thing on.

Alternatively,  we enable it by default after some time, once we believe only compliant server versions are still in use.
(Even with auto-configure, we should remove the auto-configuration code after some time).


> > Since client knows server version can we do this automatically instead of user having to opt-into it?
> 
> How client knows server version? I assume client need to execute `SELECT version()` to know it.

This would work. Of course, server can return any version string, so we need to guess whether this looks like an official Trino number and not fail if it doesn't.

Also note that it would be good to avoid additional round-trip to the server to configure the feature. Can we piggy-back on some pre-existing call? Can we auto-enable the feature eg starting from second query on a connection?

aalbu:
> I agree that -- if this feature is important -- it should auto-configure. People should typically be expected not to know to turn a thing on.
> 
> Alternatively, we enable it by default after some time, once we believe only compliant server versions are still in use. (Even with auto-configure, we should remove the auto-configuration code after some time).

Perhaps my approach of enabling by default is overly aggressive.  I was thinking that by causing a failure, we encourage people to upgrade to a new version of Trino, though I realize it would be a frustrating user experience, because it would not be clear from the exception.  I was thinking our release notes could document the behavior, but nobody RTFM.

> > How client knows server version? I assume client need to execute `SELECT version()` to know it.
> 
> This would work. Of course, server can return any version string, so we need to guess whether this looks like an official Trino number and not fail if it doesn't.
> 
> Also note that it would be good to avoid additional round-trip to the server to configure the feature. Can we piggy-back on some pre-existing call? Can we auto-enable the feature eg starting from second query on a connection?

Right, the issue we've been trying to address with this change is reducing the number of round-trips.  I am not sure how we could piggy-back on an existing call.  We could get the version before executing the first statement on a connection and cache the result.  Is there a smarter way to do it?  Note that we've seen some usage patterns that don't reuse connections (e.g. because they execute statements with different roles or on behalf of different users).  Such situations would not benefit from caching the version.

But to summarize, is the recommendation to use an auto-detection mechanism, but still provide a toggle to allow users to configure behavior?

trino/dbapi.py
```diff
@@ -228,10 +232,28 @@ def cursor(self, legacy_primitive_types: bool = None):
         return Cursor(
             self,
             request,
-            # if legacy_primitive_types is not explicitly set in Cursor, take from Connection
+            # if legacy params are not explicitly set in Cursor, take them from Connection
             legacy_primitive_types if legacy_primitive_types is not None else self.legacy_primitive_types
         )
 
+    @functools.lru_cache(maxsize=512)
```
mdesmet, line 239: Adding `@functools.lru_cache` on an object instance method will leak memory. as the instance can't be garbage collected.

aalbu: Thank you for catching that.

aalbu:
@findepi, @hashhar, @mdesmet PTAL.

trino/dbapi.py
```diff
@@ -78,6 +81,39 @@
...
+        self.ttl_seconds = ttl_seconds
+        self.cache = OrderedDict()
+        self.lock = Lock()
+
+    def get(self, key):
+        with self.lock:
+            if key not in self.cache:
+                return None
+            value, timestamp = self.cache[key]
+            if time() - timestamp > self.ttl_seconds:
```
hashhar, line 96: the time isn't enforced until an access is made, deserves to be a comment where the cache is instantiated.

trino/dbapi.py
```diff
@@ -78,6 +81,39 @@
...
+            self.cache[key] = value, time()
+            self.cache.move_to_end(key)
+            if len(self.cache) > self.capacity:
+                self.cache.popitem(last=False)
+
+    def __repr__(self):
+        return f"LRUCache(capacity: {self.capacity}, ttl: {self.ttl_seconds} seconds, {self.cache})"
+
+
+legacyPreparedStatementsCache = TimeBoundLRUCache(1024, 3600)
```
hashhar, line 114: the name is very misleading. It's caching hosts which only support legacy prepared statements.

trino/dbapi.py
```diff
@@ -228,10 +266,31 @@ def cursor(self, legacy_primitive_types: bool = None):
...
+        if self.legacy_prepared_statements is not None:
+            return self.legacy_prepared_statements
+
+        value = legacyPreparedStatementsCache.get((self.host, self.port))
+        if value is None:
+            try:
+                query = trino.client.TrinoQuery(
+                    self._create_request(),
+                    # The version() function was introduced in Trino version 352
+                    query="SELECT node_version FROM system.runtime.nodes WHERE coordinator = true")
```
hashhar, line 283: the user may not always have access to `system.runtime` in which case the query will fail and our logic here would assume it supports EXECUTE IMMEDIATE.

aalbu: Ugh, I meant to return `False` there.

trino/dbapi.py
aalbu: Done.

trino/dbapi.py
aalbu: Renamed.

trino/dbapi.py
```diff
@@ -228,10 +268,31 @@ def cursor(self, legacy_primitive_types: bool = None):
...
+                rows = query.execute().rows
+                if rows:
+                    version = rows[0][0]
+                    value = (version <= "417")
+                    mustUseLegacyPreparedStatements.put((self.host, self.port), value)
+            except Exception:
+                # not updating the cache
+                value = True
```
hashhar, line 293: in case of failure with the query we'd not cache the failure and keep retrying on every new connection creation.

Is the failure expected to be transient? IMO we should cache failure as well and revert to legacy statements in such cases - there's no downside to doing so.

hashhar (commented):
LGTM % one last comment

mdesmet (approved):
LGTM except small style remark.

trino/dbapi.py
aalbu: Yeah, I was thinking the failure would be transient, but your suggestion makes sense.

aalbu: Given this, what do you think about changing the query used for checking the capability.  Instead of relying on the version check, just see if the new statement works.  Basically run `EXECUTE IMMEDIATE 'SELECT 1'` and see if it succeeds.

hashhar: It makes sense to me.

I also see that SELECT can be constrained via resource groups/access control, I'm not sure if `EXECUTE` has such restrictions as well.

hashhar:
Please squash after applying https://github.com/trinodb/trino-python-client/pull/375#discussion_r1202432037 (if you want to add that) or squash otherwise too.

hashhar (approved):
LGTM.

cc: @electrum on similar caching approach we can follow in JDBC driver when switching prepared statements to EXECUTE IMMEDIATE.

hashhar:
@aalbu please check if the release notes at https://github.com/trinodb/trino-python-client/issues/377#issuecomment-1562370072 properly capture the change.

PR #376, opened 2023-05-23
Bump min requests version to 2.31.0

hovaesco:
## Description
Bump min requests version to 2.31.0

## Release notes

```markdown
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,8 @@ setup(
         "backports.zoneinfo;python_version<'3.9'",
         "python-dateutil",
         "pytz",
-        "requests",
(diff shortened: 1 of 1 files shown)


hashhar (approved):
seems like a no-op practically since the only reason an older version would be used is if it's not available for your platform.
And this doesn't affect people who've already installed the package.

LGTM anyway

hashhar:
add some context to the commit message and a comment in the requirements.txt, we know it's about a CVE today but no-one else in future will remember.

hovaesco:
Done.

PR #378, opened 2023-05-25
Release notes for 0.325.0

### LINKED ISSUE #377: Release notes for 0.325.0
~~~
```markdown
* <release note text>
  ([#NNN](https://github.com/trinodb/trino-python-client/issues/NNN))
```
~~~

hashhar:
## Description

Fixes #377

## Release notes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,15 @@ list](https://github.com/trinodb/trino-python-client/tags), the
 [README](https://github.com/trinodb/trino-python-client/blob/master/README.md)
 and the [PyPI page](https://pypi.org/project/trino/).
 
+## Release 0.325.0
+
+* Queries with parameters now uses `EXECUTE IMMEDIATE` instead of `PREPARE`,
+  `EXECUTE` and `DEALLOCATE` if the server supports `EXECUTE IMMEDIATE`. If
+  `EXECUTE IMMEDIATE` is not support the client falls back to using `PREPARE`,
+  `EXECUTE` and `DEALLOCATE`. If you want to always use `PREPARE`, `EXECUTE`
+  and `DEALLOCATE` set the kwarg `legacy_prepared_statements` to `True`.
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -6,6 +6,16 @@ list](https://github.com/trinodb/trino-python-client/tags), the
 [README](https://github.com/trinodb/trino-python-client/blob/master/README.md)
 and the [PyPI page](https://pypi.org/project/trino/).
 
+## Release 0.325.0
+
+* Queries with parameters now uses `EXECUTE IMMEDIATE` instead of `PREPARE`,
+  `EXECUTE` and `DEALLOCATE` if the server supports `EXECUTE IMMEDIATE`. If you
```
ebyhr, line 12: > if the server supports `EXECUTE IMMEDIATE`

Do you mean python client automaticaly falls back to PREPARE when the server doesn't support EXECUTE IMMEDIATE statement?

CHANGES.md
```diff
@@ -6,6 +6,16 @@ list](https://github.com/trinodb/trino-python-client/tags), the
...
+  want to revert to older mechanism set the kwarg `legacy_prepared_statements`
+  to `True`.
+  ([#375](https://github.com/trinodb/trino-python-client/issues/375))
+* Update minimum required `requests` library version to `2.31.0`.
```
ebyhr, line 16: nit: We didn't mention the library bump in the past release notes. Do we start mentioning such changes from this version?

ebyhr (approved):

CHANGES.md
hashhar: Yes.

CHANGES.md
hashhar: I agree. Let me remove this. I included this initially because if anyone has a constraint in their own `requirements.txt` which doesn't satisfy `request>=2.31.0` then the install would fail. However I think that's something for the user to fix since pip gives proper error message in such case.

PR #380, opened 2023-05-26
Fix handling of legacy_prepared_statements parameter

aalbu:
## Release notes

```markdown
```

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -10,6 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import math
+import time as t
 import uuid
 from datetime import date, datetime, time, timedelta, timezone
 from decimal import Decimal
@@ -27,7 +28,7 @@ from tzlocal import get_localzone_name  # type: ignore
 import trino
 from tests.integration.conftest import trino_version
 from trino import constants
-from trino.dbapi import Cursor, DescribeOutput
+from trino.dbapi import Cursor, DescribeOutput, TimeBoundLRUCache
 from trino.exceptions import NotSupportedError, TrinoQueryError, TrinoUserError
 from trino.transaction import IsolationLevel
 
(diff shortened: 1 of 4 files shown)


tests/integration/test_dbapi_integration.py
```diff
@@ -10,6 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import math
+import time as t
```
hashhar, line 13: there's already `time` imported at L15.

```suggestion
```

tests/integration/test_dbapi_integration.py
```diff
@@ -1782,6 +1783,43 @@ def test_rowcount_insert(trino_connection):
...
+        pytest.param(False, marks=pytest.mark.skipif(
+            trino_version() <= '417',
+            reason="EXECUTE IMMEDIATE was introduced in version 418")),
+        None
+    ]
+)
+def test_prepared_statement_capability_autodetection(legacy_prepared_statements, run_trino):
+    # start with an empty cache
+    trino.dbapi.must_use_legacy_prepared_statements = TimeBoundLRUCache(1024, 3600)
+    user_name = f"user_{t.monotonic_ns()}"
```
hashhar, line 1799: (time is already imported)
```suggestion
    user_name = f"user_{time.monotonic_ns()}"
```

tests/integration/test_dbapi_integration.py
```diff
@@ -1782,6 +1783,43 @@ def test_rowcount_insert(trino_connection):
...
+        host=host,
+        port=port,
+        user=user_name,
+        legacy_prepared_statements=legacy_prepared_statements,
+    )
+    cur = connection.cursor()
+    cur.execute("SELECT ?", [1])
+    cur.fetchall()
+    cur.execute("SELECT ?", [42])
+    cur.fetchall()
```
hashhar, line 1812: Is 2 queries on single connection testing something? We do test with 2 cursors just after this so we do test cache working properly just with that.

```suggestion
    cur.execute("SELECT ?", [42])
    cur.fetchall()
```

hashhar (approved):
LGTM % comments

tests/integration/test_dbapi_integration.py
hashhar: If i missed something which the two queries are testing then please add a comment since it's not obvious

hovaesco (approved):

tests/integration/test_dbapi_integration.py
aalbu: I need the `time` package here.  The other import is the time `datetime.time` class.

tests/integration/test_dbapi_integration.py
aalbu: No, I started with the first 2 queries and then I realized I should test different connections and didn't go back to remove one of the first 2.

hashhar (approved):

PR #393, opened 2023-07-18
Wrap all requests.exceptions.RequestException as TrinoConnectionError

Shaheer-rossoneri14:
## Description

Wrap all requests.exceptions.RequestException as TrinoConnectionError.
Fixes #364.

## Non-technical explanation

Before this change if a user needed to catch connection errors (e.g. connection refused) they needed to add a dependency on requests and import the relevant exception from requests.exceptions module.

After this change all requests exception get rethrown as a TrinoConnectionError instead.

## Release notes

( ) This is not user-visible or docs only and no release notes are required.
(x) Release notes are required, please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
* Fix some things. ({issue}`issuenumber`)
```

PR #394, opened 2023-07-18
Add defer_connect config to allow eagerly verifying connection

Shaheer-rossoneri14:
## Description

Add defer_connect config to allow eagerly verifying connection.

Depends on #393.

## Non-technical explanation

This commit adds a new connection parameter `defer_connect` which can be
set to False to force creating a connection when `trino.dbapi.connect`
is called. Any connection errors as a result of that get rewrapped into
`trino.exceptions.TrinoConnectionError`.

This doesn't end up actually executing a query on the server because
after the initial POST request the nextUri in the response is not
followed which leaves the query in QUEUED state. This is not documented
in the Trino REST API but the server does behave like this today. The
benefit is that we can very cheaply verify if the connection is valid
without polluting the server's query history or adding queries to queue.

## Release notes

```markdown
```


hovaesco:
Please drop https://github.com/trinodb/trino-python-client/pull/394/commits/9d417d60c85eac7534f3d2b743da1ee976d74e70 from here since it's being added here https://github.com/trinodb/trino-python-client/pull/393

Shaheer-rossoneri14:
@hovaesco Dropped the commit https://github.com/trinodb/trino-python-client/commit/9d417d60c85eac7534f3d2b743da1ee976d74e70.

hashhar:
@ebyhr @mdesmet @hovaesco what are your opinions? I like this because it's a relatively small change and addresses a complaint I've heard quite often. The downside is there are already enough parameters for the `connect` method.

hovaesco:
It looks like a useful change from my perspective, note that PyMySql sets `defer_connect` to False by default.

trino/dbapi.py
```diff
@@ -157,6 +159,7 @@ def __init__(
         legacy_prepared_statements=None,
         roles=None,
         timezone=None,
+        defer_connect=True,
```
hashhar, line 162: Please change the default to `False`.

trino/dbapi.py
```diff
@@ -201,6 +204,36 @@ def __init__(
...
+        )
+        try:
+            test_response = connection_test_request.post("<not-going-to-be-executed>")
+            if not test_response.ok:
+                raise trino.exceptions.TrinoConnectionError(
+                    "error {}{}".format(
+                        test_response.status_code,
+                        ": {}".format(test_response.content)
+                        if test_response.content
+                        else "",
```
hashhar, line 231: please simplify the format string, maybe using simple concatenation instead

right now it's using nested format like `error {}{: {}}`

hovaesco:
Please rebase on master.

trino/dbapi.py
```diff
@@ -201,6 +204,31 @@ def __init__(
...
+            self._client_session,
+            self._http_session,
+            self.http_scheme,
+            self.auth,
+            self.max_attempts,
+            self.request_timeout,
+            verify=self._http_session.verify,
+        )
+        try:
+            test_response = connection_test_request.post("<not-going-to-be-executed>")
```
hashhar, line 223: @electrum The server doesn't seem to "execute" the query until the first `nextUri` is followed. i.e. if you send just a POST /v1/statement and then do nothing then query execution doesn't happen.

Is this expected from the protocol? Asking because this change relies on this behaviour to allow verifying that connection is indeed correct - e.g. proper auth instead of failing when the first query is submitted.

This is a UX issue which people complain about when using the Trino CLI or JDBC as well for example.

hashhar: I discussed this with David offline and he said that's it's expected behaviour.

However if we don't send a DELETE the query might occupy resources on the server. So in this change we should send a subsequent DELETE as well for the query.

Additionally we also discussed about making the protocol more "synchronous" in future so that connection verification happens eagerly always.

@Shaheer-rossoneri14 Can you adjust this code to send a DELETE for the fake query we executed as well.

trino/dbapi.py
```diff
@@ -201,6 +204,31 @@ def __init__(
         self.legacy_primitive_types = legacy_primitive_types
         self.legacy_prepared_statements = legacy_prepared_statements
 
+        if not defer_connect:
```
hashhar, line 207:
```suggestion
        if not defer_connect and auth != constants.DEFAULT_AUTH:
```

@hovaesco This would mean we'll only do eager connection if some authentication is actually provided. WDYT? Or do you think existing code is better (simpler and we don't assume when someone might want eager connections)?

hovaesco: I don't see any production use cases when auth is `constants.DEFAULT_AUTH` but might be helpful in some local testing.

Issue #397, opened 2023-08-01
getting can't render element of type MAP exception


dudu-upstream:
### Expected behavior

Im trying to describe some table schema that contains a column that is MAP data type but getting the following exception:

sqlalchemy.exc.CompileError: (in table 'MY_TABLE', column 'SOME_COLUMN'): Compiler <trino.sqlalchemy.compiler.TrinoTypeCompiler object at 0x110fb7d30> can't render element of type MAP

AttributeError: 'TrinoTypeCompiler' object has no attribute 'visit_MAP'

### Actual behavior

AttributeError: 'TrinoTypeCompiler' object has no attribute 'visit_MAP'

### Steps To Reproduce

create a new db connection and try to get a table schema

### Operating System

MacOS

### Trino Python client version

0.326.0

### Trino Server version

419

### Python version

Python 3.9.17

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #398, opened 2023-08-02
Add support to data type MAP in TrinoTypeCompiler

dudub12:
Will my attempt to integrate with LangChain and Trino using this library, I countered the following exception:

```
sqlalchemy.exc.CompileError: (in table 'MY_TABLE', column 'SOME_COLUMN'): Compiler <trino.sqlalchemy.compiler.TrinoTypeCompiler object at 0x110fb7d30> can't render element of type MAP

AttributeError: 'TrinoTypeCompiler' object has no attribute 'visit_MAP'
```

After some research, I found out that TrinoTypeCompiler doesn't support MAP data type.
I have edited the compiler so it will support `visit_MAP` function and it worked on my local machine.

Fixes https://github.com/trinodb/trino-python-client/issues/397

### DIFF
--- a/trino/sqlalchemy/compiler.py
+++ b/trino/sqlalchemy/compiler.py
@@ -196,6 +196,9 @@ class TrinoTypeCompiler(compiler.GenericTypeCompiler):
     def visit_JSON(self, type_, **kw):
         return 'JSON'
 
+    def visit_MAP(self, type_, **kw):
+        return 'MAP'
+
 
 class TrinoIdentifierPreparer(compiler.IdentifierPreparer):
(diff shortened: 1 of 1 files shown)


hashhar:
Please add a test where you create a table with MAP type and try to do some operation via SQLAlchemy which fails without your change.

hashhar:
@dudub12 Can you please add a test? This is ready to be merged otherwise.

hashhar:
Superseded by https://github.com/trinodb/trino-python-client/pull/448

Thanks for showing the way @dudub12

PR #400, opened 2023-08-08
Deferred fetch

dungdm93:
fix apache/superset#24858


giftig (approved):
One small suggestion

giftig:
This probably needs unit test coverage as well.

dungdm93:
Test added

tests/unit/test_client.py
```diff
@@ -1016,6 +1016,59 @@ def json(self):
         assert isinstance(result, TrinoResult)
 
 
+def test_trino_query_deferred_fetch(sample_get_response_data):
+    """
+    Validates that the `TrinoQuery.execute` function deferred_fetch and non-block execution
+    """
+
+    class MockResponse(mock.Mock):
```
giftig, line 1024: Looks like this is copied and pasted from another method, so it would be nice to extract it out and stay DRY.

tests/unit/test_client.py
```diff
@@ -1016,6 +1016,59 @@ def json(self):
...
+                'X-Trino-Fake-1': 'one',
+                'X-Trino-Fake-2': 'two',
+            }
+
+        def json(self):
+            return sample_get_response_data
+
+    rows = sample_get_response_data['data']
+    sample_get_response_data['data'] = []
+    sql = 'execute my_stament using 1, 2, 3'
```
giftig, line 1038: statement?

tests/unit/test_client.py
```diff
@@ -1016,6 +1016,59 @@ def json(self):
...
+        client_session=ClientSession(
+            user="test",
+            source="test",
+            catalog="test",
+            schema="test",
+            properties={},
+        ),
+        http_scheme="http",
+    )
+    query = TrinoQuery(
```
giftig, line 1051: Not sure if there's a formatting standard in this project, but this could go comfortably on one line.

giftig (commented):
A couple of comments on the test

tests/unit/test_client.py
giftig: That's something which could just be passed int othe constructor though, right? There's no reason for it to refer to local scope. It's a whole class which has been copied and pasted and defined in a local function, I think now is the right time to refactor it and stay DRY personally.

trino/client.py
```diff
@@ -700,14 +700,13 @@ def rownumber(self) -> int:
...
-        while not self._query.finished or self._rows is not None:
-            next_rows = self._query.fetch() if not self._query.finished else None
+        while not self._query.finished or self._rows:
             for row in self._rows:
                 self._rownumber += 1
                 logger.debug("row %s", row)
                 yield row
 
-            self._rows = next_rows
+            self._rows = self._query.fetch() if not self._query.finished else []
```
mdesmet, line 709: Through this change the above comment is not valid anymore as the `next_uri` is only acknowledged after exposing all rows from `self._rows` through dbapi. The result is the same but now the query will be hanging until the dbapi method `fetch_one` or `fetch_many` has processed all rows in the buffer.

Also in the future the fetch could be made async, done in a separate thread, which allows for continuously looping through rows and fetching the next resultset.

dungdm93:
Hi @mdesmet, @hashhar, could you help me review this PR.

tests/unit/conftest.py
```diff
@@ -14,8 +14,10 @@
 
 import pytest
 
+from tests.unit.oauth_test_utils import SERVER_ADDRESS
```
mdesmet, line 17: Shouldn't we put SERVER_ADDRESS in a more central place if we want to use it in multiple places. here the usage is not related to oauth.

mdesmet:
@dungdm93 : Can you handle the open comments?

PR #406, opened 2023-09-08
add Trino Python Client version to user agent

aksakalli:
## Description

Trino introduces some breaking changes for old client versions but there is no way to find out the client version of our users to warn them before upgrading to a newer version since the python client is using the default user agent from `requests` that results `python-requests/2.31.0`. However, the JDBC driver [sets](https://github.com/trinodb/trino/blob/423/client/trino-jdbc/src/main/java/io/trino/jdbc/NonRegisteringTrinoDriver.java#L123) a more descriptive user agent header value such as `Trino JDBC Driver/371` that helps us to identify users with outdated client versions.

This PR aims to specify the user agent to pass the current driver version to the coordinator to be logged later in query created and completed events as [`io.trino.spi.eventlistener.QueryContext.userAgent`](https://github.com/trinodb/trino/blob/423/core/trino-spi/src/main/java/io/trino/spi/eventlistener/QueryContext.java#L39C1-L39C1).

## Non-technical explanation

This change will allow us to spot outdated clients accessing our Trino cluster.

## Release notes

```markdown
* Add Trino Python Client version to User-Agent header. ({issue}`406`)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,18 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import ast
-import re
+import os
 import textwrap
+from codecs import open
+from typing import Any
 
 from setuptools import find_packages, setup
 
-_version_re = re.compile(r"__version__\s+=\s+(.*)")
-
-with open("trino/__init__.py", "rb") as f:
-    trino_version = _version_re.search(f.read().decode("utf-8"))
-    assert trino_version is not None
-    version = str(ast.literal_eval(trino_version.group(1)))
+about = {}  # type: dict[str, Any]
+here = os.path.abspath(os.path.dirname(__file__))
+with open(os.path.join(here, "trino", "_version.py"), "r", "utf-8") as f:
+    exec(f.read(), about)
 
 kerberos_require = ["requests_kerberos"]
 sqlalchemy_require = ["sqlalchemy >= 1.3"]
@@ -44,14 +43,14 @@ tests_require = all_require + [
 ]
 
 setup(
-    name="trino",
-    author="Trino Team",
-    author_email="python-client@trino.io",
-    version=version,
-    url="https://github.com/trinodb/trino-python-client",
+    name=about["__title__"],
+    author=about["__author__"],
+    author_email=about["__author_email__"],
+    version=about["__version__"],
+    url=about["__url__"],
     packages=find_packages(include=["trino", "trino.*"]),
     package_data={"": ["LICENSE", "README.md"]},
-    description="Client for the Trino distributed SQL Engine",
+    description=about["__description__"],
     long_description=textwrap.dedent(
         """
     Client for Trino (https://trino.io), a distributed SQL engine for
@@ -59,7 +58,7 @@ setup(
     a DBAPI 2.0 implementation.
     """
     ),
-    license="Apache 2.0",
+    license=about["__license__"],
     classifiers=[
         "Development Status :: 4 - Beta",
         "Intended Audience :: Developers",
--- a/trino/__init__.py
+++ b/trino/__init__.py
@@ -11,7 +11,28 @@
 # limitations under the License.
 
 from . import auth, client, constants, dbapi, exceptions, logging
+from ._version import (
+    __author__,
+    __author_email__,
+    __description__,
+    __license__,
+    __title__,
+    __url__,
+    __version__,
+)
 
-__all__ = ['auth', 'dbapi', 'client', 'constants', 'exceptions', 'logging']
-
-__version__ = "0.326.0"
+__all__ = [
+    "auth",
+    "client",
+    "constants",
+    "dbapi",
+    "exceptions",
+    "logging",
+    "__author__",
+    "__author_email__",
+    "__description__",
+    "__license__",
+    "__title__",
+    "__url__",
+    "__version__",
+]
--- a/trino/client.py
+++ b/trino/client.py
@@ -62,6 +62,7 @@ from tzlocal import get_localzone_name  # type: ignore
 
 import trino.logging
 from trino import constants, exceptions
+from trino._version import __version__
 
 __all__ = ["ClientSession", "TrinoQuery", "TrinoRequest", "PROXIES"]
 
@@ -447,7 +448,7 @@ class TrinoRequest(object):
 
     @property
     def http_headers(self) -> Dict[str, str]:
-        headers = {}
+        headers = requests.structures.CaseInsensitiveDict()
 
         headers[constants.HEADER_CATALOG] = self._client_session.catalog
         headers[constants.HEADER_SCHEMA] = self._client_session.schema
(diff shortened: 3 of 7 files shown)


hashhar (commented):
Please squash the commits.

This is a very nice change. Thank you @aksakalli.

trino/client.py
```diff
@@ -455,6 +455,7 @@ def http_headers(self) -> Dict[str, str]:
         headers[constants.HEADER_USER] = self._client_session.user
         headers[constants.HEADER_TIMEZONE] = self._client_session.timezone
         headers[constants.HEADER_CLIENT_CAPABILITIES] = 'PARAMETRIC_DATETIME'
+        headers["User-Agent"] = f"Trino Python Client/{__version__}"
```
hashhar, line 458: is there a constant that request provides for the header name?

It may not matter if requests normalises the keys of the dict when merging the headers.

hashhar: Also it seems we can test this via the mocked client in `test_client.py`?

aksakalli: There is no constant for that in `requests` (also see [examples](https://requests.readthedocs.io/en/latest/user/quickstart/#custom-headers)) but I think it's more readable to have it directly in dictionary. I can create a constant for the text `Trino Python Client`, shall I?

I will add the test 👌

aksakalli: tests are added.

aksakalli:
> Please squash the commits.

Is it possible to squash merge? I would rather like to keep my history for reference for fixing issues.

trino/__init__.py
```diff
@@ -12,6 +12,14 @@
 
 from . import auth, client, constants, dbapi, exceptions, logging
 
-__all__ = ['auth', 'dbapi', 'client', 'constants', 'exceptions', 'logging']
```
aksakalli, line 15: sorting the list

trino/client.py
```diff
@@ -447,14 +448,15 @@ def transaction_id(self, value):
 
     @property
     def http_headers(self) -> Dict[str, str]:
-        headers = {}
```
aksakalli, line 450: It might cause issues to use the default dict if the user tries to override any default header with a different case like `User-Agent` instead of `user-agent`, since the headers are case insensitive.

The requests is also [using](https://github.com/psf/requests/blob/v2.31.0/requests/utils.py#L898) this utility struct for creating the default header.

setup.py
```diff
@@ -12,18 +12,17 @@
...
-_version_re = re.compile(r"__version__\s+=\s+(.*)")
-
-with open("trino/__init__.py", "rb") as f:
-    trino_version = _version_re.search(f.read().decode("utf-8"))
-    assert trino_version is not None
-    version = str(ast.literal_eval(trino_version.group(1)))
+about = {}  # type: dict[str, Any]
+here = os.path.abspath(os.path.dirname(__file__))
+with open(os.path.join(here, "trino", "_version.py"), "r", "utf-8") as f:
+    exec(f.read(), about)
```
aksakalli, line 25: taken from [requests/setup.py](https://github.com/psf/requests/blob/147c8511ddbfa5e8f71bbf5c18ede0c4ceb3bba4/setup.py#L76-L79)

trino/client.py
```diff
@@ -62,6 +62,7 @@
 
 import trino.logging
 from trino import constants, exceptions
+from trino._version import __version__
```
aksakalli, line 65: having version in `trino/__init__.py` would cause a circular dependency for the following import statement:

```python
from trino import __version__
```

So I ended up moving to another submodule called `_version`. Much more convenient to be accessed by `setup.py` and insider `trino`.

aksakalli:
Hi @aalbu @hashhar, I made the requested changes and added some explanation comments. Could you review and merge this PR?

trino/client.py
hovaesco: why exactly name `_version` ? could you please point to some docs that describes that rule?

aksakalli: it could be `__version__.py` as well later to be overwrite in `__init__.py` with `from .__version__ import __version__` like how `requests` does it.

I choose `_version.py` as a "private" package path. This is kinda pseudo convention and it's not a strict rule. I [searched](https://github.com/search?q=path%3A**%2F_version.py+language%3APython&type=code&ref=advsearch) on Github for `_version.py` and found 18k matches. 

I can change it to `__version__.py` if you prefer so.

hovaesco: Thanks, I don't have a strong preference here, just was looking for some "official" rule. 

@hashhar WDYT? I believe these changes should be extracted to a separate commit.

hashhar: Yes, this setup.py makes sense to extract.

Sorry I didn't notice initially.

hashhar: Either filename is fine by me. The _version_.py does however give the feeling it's something special.

trino/__init__.py
```diff
@@ -11,7 +11,28 @@
...
+    "constants",
+    "dbapi",
+    "exceptions",
+    "logging",
+    "__author__",
+    "__author_email__",
+    "__description__",
+    "__license__",
+    "__title__",
+    "__url__",
```
hashhar, line 36: can you remove these?

User's of the module generally should have no use for these properties.

aksakalli: everything else other than `__version__`? It's kind of an old convention but it provides package related information after importing it. For examples when you run:

```python
import requests
help(requests)
```
gives the following output:

```
...

DATA
    __author_email__ = 'me@kennethreitz.org'
    __build__ = 141313
    __cake__ = '✨ 🍰 ✨'
    __copyright__ = 'Copyright 2022 Kenneth Reitz'
    __description__ = 'Python HTTP for Humans.'
    __license__ = 'Apache 2.0'
    __title__ = 'requests'
    __url__ = 'https://requests.readthedocs.io'
    chardet_version = '4.0.0'
    charset_normalizer_version = '2.1.1'
    codes = <lookup 'status_codes'>

VERSION
    2.28.1

AUTHOR
    Kenneth Reitz

...
```

It's not super essential but no harm keeping package information accessible with dunder.

hashhar: makes sense. I can see a few other packages following the convention.

Can you however extract all `setup.py` related preparatory changes (including this) to a separate commit?

hashhar: i.e. every change except in `test_client.py`, `client.py` and `constants.py` should be a preparatory commit.

It's safer to keep refactors separate from functional changes - helps pay better attention in reviews + easier to bisect or understand in future.

aksakalli: I cannot make the change in `client.py` before refactoring that includes separating `__version__` from `trino/__init__.py`. Shall we merge the refactoring first?

aksakalli: I created a refactoring PR https://github.com/trinodb/trino-python-client/pull/410 I can rebase this one after merging the refactoring.

hashhar:
superseded by https://github.com/trinodb/trino-python-client/pull/411

@aksakalli Should this be closed?

aksakalli:
closing this one in favour of #411

PR #410, opened 2023-09-15
separate __version__ from __init__

aksakalli:
## Description

As part of adding a user-agent to client task, `trino.__version__` should be accessible within the package itself otherwise a statement such as

```python
from trino import __version__
```

would cause a circular dependency. Everything within a package should be able to imported correctly before importing the root level `__init__.py`. 

Having a "private" package `trino._version` to store meta info about this package is a  pseudo convention  that followed by many popular python libraries. GitHub contains 18.8k repos with such convention according to [this search](https://github.com/search?q=path%3A**%2F_version.py+language%3APython&type=code&ref=advsearch).  

For more explanation:

* https://www.python.org/dev/peps/pep-0396/
* https://packaging.python.org/guides/single-sourcing-package-version/

## Non-technical explanation

This change will make `trino.__version__` accessible from a central place within the package itself.

## Release notes

```markdown
```


hovaesco (approved):

aksakalli:
Can we merge this? @hashhar @hovaesco

hashhar (commented):
Good to go but I need to adjust https://github.com/trinodb/trino-python-client/blob/master/.github/workflows/release.yml to this.

That's why I've not merged it yet.

aksakalli:
> Good to go but I need to adjust https://github.com/trinodb/trino-python-client/blob/master/.github/workflows/release.yml to this.
> 
> That's why I've not merged it yet.

Sorry, I missed that out. I didn't know that the next version is generated from CI.

https://github.com/trinodb/trino-python-client/blob/d6dee794eeecf59ab5725671094ec7e5aa741fff/.github/workflows/release.yml#L30

I can simply replace `trino/__init__.py` with `trino/_version.py`?

hashhar (approved):

hashhar:
I've just reworded the commit message + squashed (otherwise there would be one commit at which point automated releases wouldn't work).

Will merge once CI is finished.

aksakalli:
> I've just reworded the commit message + squashed (otherwise there would be one commit at which point automated releases wouldn't work).
> 
> Will merge once CI is finished.

I don't think force push to a feature branch is the best approach, you can always squash merge to avoid that happening. That's the convention most repos are following including `trindb/trino`.

hashhar:
> squash merge to avoid that happening. 

Needs to be enabled per repo. We don't have it enabled on this. It's enabled on trinodb/trino but don't always squash merge since in most cases the history and smaller commits are useful.

PR #412, opened 2023-09-27
add markdown long description

aksakalli:
## Description

The `README.md` contains the full long description such as documentation. The current text is not comprehensive as `README.md`. This file is ignored during installation and the content is lost when it's published.

For example https://pypi.org/project/trino/ shows 

Screenshot 2023-09-27 at 13 01 50

After this change it will be like the `requests` package https://pypi.org/project/requests/ which [defines](https://github.com/psf/requests/blob/ee93fac6b2f715151f1aa9a1a06ddba9f7dcc59a/setup.py#L88) a long markdown description.

## Non-technical explanation

The readme content will be published in the package.

## Release notes

```markdown
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import os
-import textwrap
 from codecs import open
 from typing import Any
 
@@ -23,6 +22,9 @@ about = {}  # type: dict[str, Any]
 here = os.path.abspath(os.path.dirname(__file__))
 with open(os.path.join(here, "trino", "_version.py"), "r", "utf-8") as f:
     exec(f.read(), about)
+
+with open(os.path.join(here, "README.md"), "r", "utf-8") as f:
+    readme = f.read()
 
 kerberos_require = ["requests_kerberos"]
 sqlalchemy_require = ["sqlalchemy >= 1.3"]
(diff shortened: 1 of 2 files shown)


nineinchnick (approved):

aksakalli:
> while I can see how it is useful it also introduces two different places for reading documentation with no clarity to users where the information on pypi is coming from or how to ask questions or send feedback.
> 
> We should instead add a link to the README in the existing long_description text.
> 
> @hovaesco opinions?


PyPi is just an example but it's about keeping the documentation content in the published package. It's also kind of a convention. [Here](https://github.com/search?q=long_description_content_type+language%3APython+&type=code) you can see different python projects following it.

hashhar:
In all of those examples the readme is super small and nowhere near as exhaustive as the readme for this project.

Can you explain a bit more about "about keeping the documentation content in the published package"? What's the motivation for this - maybe I'm just not understanding what you are trying to solve.

hovaesco:
Standard convention in PyPi uses README as long_description. It might be beneficial to see documentation in PyPi directly, but I usually follow Github for it. Anyway, I don't have a strong opinion on it.

hashhar (approved):

setup.py
```diff
@@ -51,13 +53,8 @@
     packages=find_packages(include=["trino", "trino.*"]),
     package_data={"": ["LICENSE", "README.md"]},
     description=about["__description__"],
-    long_description=textwrap.dedent(
-        """
-    Client for Trino (https://trino.io), a distributed SQL engine for
-    interactive and batch big data processing. Provides a low-level client and
-    a DBAPI 2.0 implementation.
```
hashhar, line 58: @aksakalli Can you change the leading statement of existing `README.md` to include this information.

So something like:

```
Client for Trino (https://trino.io), a distributed SQL engine for interactive and batch big data processing.
Provides a low-level client and a DBAPI 2.0 implementation and a SQLAlchemy adapter.
It supports Python>=3.7 and pypy.
```

aksakalli: done in c521f75 let me know if it's ok, I can squash the commits

hashhar (approved):

setup.py
hashhar: Looks good, please squash and I'll merge.

aksakalli: squashed as 78fb9b0

PR #420, opened 2023-10-18
Check cursor _query is None before columns

ammarar:
## Description
Fixes an issue where if `cursor.description` is called before `execute`, then it returns `AttributeError` for `columns`. This make this client not work with [Ray Data](https://github.com/ray-project/ray/blob/master/python/ray/data/datasource/sql_datasource.py#L71-L81) because it checks if `description` exists.

## Non-technical explanation
Guard against error state when calling a function

## Release notes


hashhar (approved):

hashhar:
@cla-bot check

hashhar:
@ammarar Can you fix flake8 failure?

Please amend the commit and force push instead of adding a new commit.
I'll merge after CI is done.

ammarar:
@hashhar It's done! Its all yours now.

hashhar (approved):

PR #421, opened 2023-10-25
Default to https scheme when port is 443 (sqlalchemy)

alexfricker:
## Description

Attempting to use `sqlalchemy` with HTTPS scheme and default port `443` results in a `TrinoConnectionError: RemoteDisconnected('Remote end closed connection without response')` as the underlying dialect's `http_scheme` is set to `http` not `https`. This PR will automatically set the `http_scheme` to `https` which alleviates the connection issue.

Trino server: 430
Trino client: 0.327.0
SQLAlchemy: 2.0.22

## Non-technical explanation

Using sqlalchemy with a Trino server accessed via an `https` endpoint and no custom port results in a connection error. 

## Release notes

```markdown
```

### DIFF
--- a/tests/unit/sqlalchemy/test_dialect.py
+++ b/tests/unit/sqlalchemy/test_dialect.py
@@ -39,7 +39,14 @@ class TestTrinoDialect:
                 )),
                 'trino://user@localhost:443/?source=trino-sqlalchemy',
                 list(),
-                dict(host="localhost", port=443, catalog="system", user="user", source="trino-sqlalchemy"),
+                dict(
+                    host="localhost",
+                    port=443,
+                    catalog="system",
+                    user="user",
+                    source="trino-sqlalchemy",
+                    http_scheme="https",
+                ),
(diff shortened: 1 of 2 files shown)


hashhar (changes requested):
Note that the port being `443` doesn't mean it's necessarily HTTPS so you MUST NOT unconditionally override the `http_scheme`.

Either way I'm more inclined towards similar logic as the JDBC driver - https://github.com/trinodb/trino/blob/68a5ffa7dc1d2562b4c696d2684efbc9c80d9ff2/client/trino-client/src/main/java/io/trino/client/uri/TrinoUri.java#L427-L428

- if no port is provided:
  - if the url scheme is `https` or if `ssl` is enabled then use port 443
  - in other cases use port 80
- if port is explicitly provided:
  - use the provided http_scheme and port as it is

This requires removing the default values from the dbapi and dialect constructor and instead assigning the default values within the body of the method.

Issue #423, opened 2023-10-31
JSON support


d-gorodnov:
### Describe the feature

Could you please consider adding JSON support when using SQLAlchemy? Currently, it seems to be treated as a string.
https://github.com/trinodb/trino-python-client/blob/master/trino/sqlalchemy/datatype.py#L77-L87

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

Issue #387, opened 2023-07-05
JSON functionality


dwolfeu:
### Describe the feature

Can the [standard JSON operations](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.JSON) such as `some_table.c.json_column['some key']` and `some_table.c.json_column[('key_1', 'key_2', 5, ..., 'key_n')]` be implemented?

### Describe alternatives you've considered

[I sought the advice of the community.](https://stackoverflow.com/questions/76613959/json-parsing-with-trino-and-sqlalchemy-orm/)

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

hashhar:
Fixed via https://github.com/trinodb/trino-python-client/pull/425

PR #425, opened 2023-11-28
Fix JSON support in SQLAlchemy

hovaesco:
## Description
Resolves https://github.com/trinodb/trino-python-client/issues/423 and https://github.com/trinodb/trino-python-client/issues/387

## Release notes

```markdown
```

### DIFF
--- a/trino/sqlalchemy/datatype.py
+++ b/trino/sqlalchemy/datatype.py
@@ -9,15 +9,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import json
 import re
 from typing import Any, Dict, Iterator, List, Optional, Tuple, Type, Union
 
 import sqlalchemy
-from sqlalchemy import util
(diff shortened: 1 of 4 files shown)


trino/sqlalchemy/datatype.py
```diff
@@ -75,16 +74,58 @@ def __init__(self, precision=None, timezone=False):
...
+            value = '$.%s' % value
+        return value
+
+
+class JSONPathType(_FormatTypeMixin, sqltypes.JSON.JSONPathType):
+    def _format_value(self, value):
+        return "$%s" % (
+            "".join(
+                [
+                    "[%s]" % elem if isinstance(elem, int) else '.%s' % elem
```
hashhar, line 124: why not quote with `[` always? e.g. if `elem` has a dash then also you need to quote.

hashhar: also maybe extract a well-named static method for this

hovaesco: good catch, changed

hashhar (approved):

Issue #424, opened 2023-11-27
The order of loading for trino, and the __init__.py means that there's no way of setting a different log level than INFO


mb-m:
### Expected behavior

I'm running the trino client through airflow through a custom version of the trino operator, and we run our workers in DEBUG log level. We should see requests and responses happening (this helps when mocking the operator in tests too)

### Actual behavior

No log is seen in normal operation.

### Steps To Reproduce

I tried:
```python
import logging
from trino import logging as trino_logging


def get_logger(name, level=None):
    # we use the default (inherited) level rather than having trino set its own
    return logging.getLogger(name)


trino_logging.get_logger = get_logger

import trino
```
and
```python
import logging
import trino.logging


def get_logger(name, level=None):
    # we use the default (inherited) level rather than having trino set its own
    return logging.getLogger(name)


trino.logging.get_logger = get_logger

import trino
```
but because of the `trino/__init__.py` and the `__logger__` at the module roots (load time), you can't influence the log level.

### Operating System

linux

### Trino Python client version

0.322.0

### Trino Server version

N/A

### Python version

3.10

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

mb-m:
I was wondering about using environment or similar as a global dict to set the default log-level for trino (then I could do the import with the environment wrapping it). I'd be happy to submit a PR for this.

mb-m:
Specifically, I'm thinking about a diff like:
```python
diff --git a/trino/logging.py b/trino/logging.py
index 866c3e9..eafb57c 100644
--- a/trino/logging.py
+++ b/trino/logging.py
@@ -10,13 +10,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import logging
+from typing import Optional
 
-LEVEL = logging.INFO
+_level = os.environ.get('TRINO_LOGLEVEL', 'INFO')
+if _level.upper() == _level and hasattr(logging, _level) and isinstance(getattr(logging, _level), int):
+    LEVEL = getattr(logging, _level)
+elif _level == 'NONE':
+    LEVEL = None
+else:
+    LEVEL = logging.INFO
 
 
 # TODO: provide interface to use ``logging.dictConfig``
-def get_logger(name: str, log_level: int = LEVEL) -> logging.Logger:
+def get_logger(name: str, log_level: Optional[int] = LEVEL) -> logging.Logger:
     logger = logging.getLogger(name)
[...]

PR #432, opened 2023-12-19
Allow to set different log level

guerremdq:
## Description
Allow users to set a different log level for trino client. 

* Fixes [424](https://github.com/trinodb/trino-python-client/issues/424)

Credits go to @mb-m

### DIFF
--- a/trino/logging.py
+++ b/trino/logging.py
@@ -11,12 +11,21 @@
 # limitations under the License.
 
 import logging
+import os
+from typing import Optional
 
-LEVEL = logging.INFO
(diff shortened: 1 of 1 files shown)


john-bodley:
@guerremdq I sense @hashhar or @mdesmet are likely better reviewers, so I added them and removed myself.

hashhar:
This was addressed via https://github.com/trinodb/trino-python-client/pull/434.
You can now change log levels using the standard logging mechanism.

```python
trino_logger = logging.getLogger("trino")
trino_logger.setLevel(DEBUG)
```

The loggers are named same as the module name. `trino` is the root logger.

PR #434, opened 2023-12-22
Only set log level for the root logger

### LINKED ISSUE #424: The order of loading for trino, and the __init__.py means that there's no way of setting a different log level than INFO

hashhar:
## Description

Before this change we called `setLevel` on each logger with level set to INFO. This meant that it was not possible to change log level for all child modules by doing something like
`logging.getLogger('trino').setLevel(logging.DEBUG)` because the child loggers had explicit levels set already. It instead required us to change log levels for each module (`trino.client`, `trino.dbapi`, `trino.auth` etc.) separately.

After this change only the root logger `trino` has a default level set. Other child loggers inherit from it. So now the default log level for all modules can be changed by doing
`logging.getLogger('trino').setLevel(logging.DEBUG)` for example.

## Non-technical explanation

Make it easy to configure log level for all modules at once.

## Release notes

```markdown
* Make it possible to configure log level for all modules via the root logger. ({issue}`424`)
```

Fixes #424

### DIFF
--- a/trino/logging.py
+++ b/trino/logging.py
@@ -11,12 +11,21 @@
 # limitations under the License.
 
 import logging
+from typing import Optional
 
 LEVEL = logging.INFO
 
 
 # TODO: provide interface to use ``logging.dictConfig``
-def get_logger(name: str, log_level: int = LEVEL) -> logging.Logger:
+def get_logger(name: str, log_level: Optional[int] = None) -> logging.Logger:
     logger = logging.getLogger(name)
-    logger.setLevel(log_level)
+    # We must not call setLevel by default except on the root logger otherwise
+    # we cannot change log levels for all modules by changing level of the root
+    # logger
+    if log_level is not None:
+        logger.setLevel(log_level)
     return logger
+
+
+# set default log level to LEVEL
+trino_root_logger = get_logger('trino', LEVEL)


hashhar:
FYI @guerremdq

hovaesco (approved):

ebyhr (approved):

Issue #430, opened 2023-12-19
Support for oauth2 token per user per host (vs per host only)


gerilya:
### Describe the feature

Hello,

While using this package in Sagemaker Notebook I noticed that OAuth2 tokens are cached per host making it hard for multiple users to share a single notebook.
I think tokens should be cached per (host,user) pair and not just per host for the following reasons:
- Security: token is user-specific and can't (and should not) be shared between users.
- Granular access: there are use cases with granular access where different resources (schemas, tables) are accessed with different usernames/credentials. Having just 1 token per host makes it more difficult for developer to access different databases/tables within an application.
- web service: similar to Sagemaker Notebook's use case, using this package within a webservice would require a workaround.
This request is somewhat similar to [Custom cache for oauth2 tokens](https://github.com/trinodb/trino-python-client/issues/223), but offers a different approach to address it: rather than implementing a custom cache, extend official caching mechanism to support multi user token caching.
Thank you

### Describe alternatives you've considered

currently using a keyring with file-based backend which can be configured with a different filename per user. 
 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

Issue #431, opened 2023-12-19
Having no keyring backends should be treated as no keyring module and fall back to using memory-based cache


gerilya:
### Expected behavior

If an [attempt to load keyring module](https://github.com/trinodb/trino-python-client/blob/f99ef245f28b4e1784a829b7935a31d0215ace4d/trino/auth.py#L242) fails, there is a fall back to [_OAuth2TokenInMemoryCache](https://github.com/trinodb/trino-python-client/blob/master/trino/auth.py#L278).
I think no backends or having `keyring.backends.null.Keyring` backend should be treated the same way.
Using `keyring.backends.null.Keyring` backend is the recommended way to disable keyring. See `Disabling Keyring` at https://pypi.org/project/keyring/ 

### Actual behavior

When there is `keyring` package installed but there are no available backends, the following error is generated:
```
Although keyring module is installed no backend has been detected, check https://pypi.org/project/keyring/ for more information.
```
### Steps To Reproduce

Run the following code on Sagemaker notebook instance:
```
import trino

connect = trino.dbapi.connect(
                              host=config['HOST'],
                              port=config['PORT'],
                              http_scheme=config['PROTOCOL'],
                              auth=trino.auth.OAuth2Authentication(),
                              user=config['USERNAME'],
                              )
cursor=connect.cursor()
query="some query"
cursor.execute(query)
```

### Log output

[ 2023-12-14T09:53:41.370244 | Although keyring module is installed no backend has been detected, check https://pypi.org/project/keyring/ for more information. ]

### Operating System

AWS Sagemaker Notebook

### Trino Python client version

0.327.0

### Trino Server version

can't connect - not relevant

### Python version

Python 3.10.13

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

hashhar:
@hovaesco can this be closed after #435?

hovaesco:
@hashhar Yes.

hovaesco:
Resolved by https://github.com/trinodb/trino-python-client/pull/435

PR #435, opened 2023-12-29
OAuth cache improvements

hovaesco:
## Description
Closes https://github.com/trinodb/trino-python-client/issues/430 and https://github.com/trinodb/trino-python-client/issues/431

## Release notes

```markdown
```


trino/auth.py
```diff
@@ -382,18 +390,26 @@ def _get_token(self, token_server: str, response: Response, **kwargs: Any) -> st
...
     def _determine_host(url: Optional[str]) -> Any:
         return urlparse(url).hostname
 
+    @staticmethod
+    def _determine_user(headers: Mapping[Any, Any]) -> Optional[Any]:
+        return headers.get(HEADER_USER)
+
+    @staticmethod
+    def _construct_cache_key(host: Optional[str], user: Optional[str]) -> str:
+        return f"{host}@{user}"
```
hashhar, line 411: usually people expect stuff to be `user@host` but it doesn't matter here since it's an internal cache key (unless we log it somewhere)

hashhar (commented):
first commit is unclear to me, can you explain a bit about it?

2nd looks good.

hovaesco:
If there is no backend available then keyring backend is being initialized as `backends.fail.Keyring` so first commit adds a check which validates and treats it as no keyring avaiable and fall backs to `_OAuth2TokenInMemoryCache()` https://github.com/trinodb/trino-python-client/pull/435/commits/0517c65c396c2e09787263952b0d91524e3b857a#diff-6db3a24c52bc43426976f516411e7de756f5ee0f8f3904ca036da682aeaa840cL278

trino/auth.py
```diff
@@ -382,18 +390,26 @@ def _get_token(self, token_server: str, response: Response, **kwargs: Any) -> st
...
-            self._token_cache.store_token_to_cache(host, token)
+            self._token_cache.store_token_to_cache(key, token)
 
     @staticmethod
     def _determine_host(url: Optional[str]) -> Any:
         return urlparse(url).hostname
 
+    @staticmethod
+    def _determine_user(headers: Mapping[Any, Any]) -> Optional[Any]:
+        return headers.get(HEADER_USER)
```
lukasz-walkiewicz, line 407: `X-Trino-User` header is optional

hovaesco: You're right, is there any other way to get username?

hashhar: maybe we can use the client provided principal/user instead of user which it gets resolved to on the server to bypass this problem entirely?

trino/auth.py
```diff
@@ -208,32 +209,33 @@ class _OAuth2TokenCache(metaclass=abc.ABCMeta):
...
+    In-memory token cache implementation. The token is stored per host and user pair,
+    so multiple clients can share the same cache.
     """
 
     def __init__(self) -> None:
         self._cache: Dict[Optional[str], str] = {}
 
-    def get_token_from_cache(self, host: Optional[str]) -> Optional[str]:
-        return self._cache.get(host)
+    def get_token_from_cache(self, key: Optional[str]) -> Optional[str]:
```
hashhar, line 229: if you extract the `host` -> `key` rename to it's own commit functional changes will be more visible and easier to see.

trino/auth.py
```diff
@@ -208,32 +209,33 @@ class _OAuth2TokenCache(metaclass=abc.ABCMeta):
...
-    def store_token_to_cache(self, host: Optional[str], token: str) -> None:
+    def store_token_to_cache(self, key: Optional[str], token: str) -> None:
         pass
 
 
 class _OAuth2TokenInMemoryCache(_OAuth2TokenCache):
     """
-    In-memory token cache implementation. The token is stored per host, so multiple clients can share the same cache.
+    In-memory token cache implementation. The token is stored per host and user pair,
+    so multiple clients can share the same cache.
```
hashhar, line 223: comment is not actually matching implementation.

probably should say something like:

```
Multiple clients can share the same cache only if each connection explicitly specifies a user otherwise the first cached token will be used to authenticate all other users.
```

hashhar (commented):
please update docs to add warning what happens when you don't specify `user` when connecting (token cache gets shared)

hashhar (approved):
LGTM.

@lukasz-walkiewicz do you want to take a look?

We now do like JDBC driver, if username was specified we use host+user cache key otherwise we use only host as the cache key and tokens get shared (same as JDBC driver).

PR #441, opened 2024-01-22
Fix mapping of MAP types with boolean or integer keys

hashhar:
## Description

Fix mapping of MAP types with boolean or integer keys.

This is on top of https://github.com/trinodb/trino-python-client/pull/440 so only last 3 commits need to be reviewed.

## Release notes

```markdown
* Fix mapping of MAP types with boolean or integer keys.
```

### DIFF
--- a/tests/integration/test_types_integration.py
+++ b/tests/integration/test_types_integration.py
@@ -799,11 +799,9 @@ def test_map(trino_connection):
         SqlTest(trino_connection)
         .add_field(sql="CAST(null AS MAP(VARCHAR, INTEGER))", python=None)
         .add_field(sql="MAP()", python={})
-        # TODO: boolean map keys get returned as strings
-        # .add_field(sql="MAP(ARRAY[true, false], ARRAY[false, true])", python={True: False, False: True})
-        # .add_field(sql="MAP(ARRAY[true, false], ARRAY[true, null])", python={True: True, False: None})
-        # TODO: integer map keys get returned as strings
-        # .add_field(sql="MAP(ARRAY[1, 2], ARRAY[1, null])", python={1: 1, 2: None})
+        .add_field(sql="MAP(ARRAY[true, false], ARRAY[false, true])", python={True: False, False: True})
+        .add_field(sql="MAP(ARRAY[true, false], ARRAY[true, null])", python={True: True, False: None})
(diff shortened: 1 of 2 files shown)


trino/mapper.py
```diff
@@ -31,6 +31,29 @@ def map(self, value: Any) -> Optional[T]:
...
+class BooleanValueMapper(ValueMapper[bool]):
+    def map(self, value: Any) -> Optional[bool]:
+        if value is None:
+            return None
+        if isinstance(value, bool):
+            return value
+        if str(value).lower() == 'true':
+            return True
+        if str(value).lower() == 'false':
+            return False
```
hovaesco, line 43: is there a way to avoid this and return map keys as boolean instead?

tests/integration/test_types_integration.py
```diff
@@ -794,14 +794,92 @@ def test_array(trino_connection):
...
 
+    # structural types - note that none of these below tests work in the Trino JDBC Driver either.
+    # Unhashable types like lists and dicts cannot be used as keys so these values cannot be represented as Python
+    #  objects at all.
+    # .add_field(sql="MAP(ARRAY[ARRAY[1, 2]], ARRAY[null])", python={[1, 2]: None})
+    # .add_field(sql="MAP(ARRAY[MAP(ARRAY[1], ARRAY[2])], ARRAY[null])", python={{1: 2}: None})
+
+    # TODO: fails because server sends [[{"[1, 2]":null}]] as response whereas it sends [[[1,2]]] as response for ROW
+    #  types that are not enclosed in a MAP while the RowValueMapper expects values to be lists.
+    # .add_field(sql="MAP(ARRAY[ROW(1, 2)], ARRAY[CAST(null AS VARCHAR)])", python={(1, 2): None})
```
hovaesco, line 879: follow-up issue for the TODO?

hovaesco (commented):
looks good

trino/mapper.py
hashhar: Technically yes, but it's much harder to do on server. There's already some code `FixJsonDataUtils` which tries to handle some of this but it doesn't work well or is easy to do for structural types.

ebyhr (approved):

tests/integration/test_types_integration.py
hashhar: https://github.com/trinodb/trino-python-client/issues/442

PR #443, opened 2024-01-30
Add support for 'None' on response data

avibhstarburst:
## Description
Add support for 'None' on response data 

## Release notes

```markdown
```

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -186,6 +186,29 @@ def sample_get_response_data():
 
 
 @pytest.fixture(scope="session")
+def sample_get_response_data_none():
+    """
+    This is the response to the second HTTP request (a GET) from an actual
+    Trino session. It is deliberately not truncated to document such response
+    and allow to use it for other tests. After doing the steps above, do:
+
+    ::
+        >>> cur.fetchall()
+
+    """
+    yield {
+        "id": "20210817_140827_00000_arvdv",
+        "nextUri": "coordinator:8080/v1/statement/20210817_140827_00000_arvdv/2",
+        "data": None,
+        "columns": [],
+        "taskDownloadUris": [],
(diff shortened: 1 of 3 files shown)


trino/client.py
```diff
@@ -633,6 +633,9 @@ def process(self, http_response) -> TrinoStatus:
 
         self._next_uri = response.get("nextUri")
 
+        data = response.get("data", [])
```
hashhar, line 636: Probably would be more readable/less surprising if you remove the 2nd empty list arg to response.get.

I understand what happens is that the "data" key exists with null value so instead of returning empty list it returns None.

hashhar (commented):
@hovaesco can you guide with how to add a unit test for this (using mocking since not easy to have isolated test case).

trino/client.py
hashhar: Or maybe:

```
data = if response.get("data") else []
```

hovaesco:
To write a unit test here please create mock like https://github.com/trinodb/trino-python-client/blob/ca38d2908533d891eedf5ee80f38dbdcfbbdb564/tests/unit/conftest.py#L64 and then see how it's used in tests https://github.com/trinodb/trino-python-client/blob/ca38d2908533d891eedf5ee80f38dbdcfbbdb564/tests/unit/test_client.py#L682

tests/unit/conftest.py
```diff
@@ -184,6 +184,126 @@ def sample_get_response_data():
...
+                "name": "state",
+                "type": "varchar",
+                "typeSignature": {
+                    "typeArguments": [],
+                    "arguments": [{"kind": "LONG_LITERAL", "value": 2147483647}],
+                    "literalArguments": [],
+                    "rawType": "varchar",
+                },
+            },
+        ],
```
hashhar, line 253: shorten this to a single column for example, our test doesn't depend on so many being present.

hashhar (approved):
LGTM % comment.

hashhar (approved):
Thanks. Will merge after CI is done.

PR #445, opened 2024-01-31
Fixed OAuth2Authentication redirect_auth_url_handler callback not firing

matialm:
Remove the Bearer from x_redirect_server, with the count remove only the first coincidence and it make fails the redirect auth url callback.

## Release notes

(X ) Release notes are required, with the following suggested text:

```markdown
* Fix the non firing auth url callback issue when oauth2 is used ({issue}`444`)
```

### DIFF
--- a/trino/auth.py
+++ b/trino/auth.py
@@ -320,7 +320,7 @@ class _OAuth2TokenBearer(AuthBase):
             raise exceptions.TrinoAuthError(f"Error: header info didn't match {auth_info}")
 
         auth_info_headers = parse_dict_header(
-            _OAuth2TokenBearer._BEARER_PREFIX.sub("", auth_info, count=1))  # type: ignore
+            _OAuth2TokenBearer._BEARER_PREFIX.sub("", auth_info))  # type: ignore
 
         auth_server = auth_info_headers.get('x_redirect_server')
         token_server = auth_info_headers.get('x_token_server')


matialm:
Contributor License Agreement signed and sended

hashhar:
cc: @lukasz-walkiewicz

lukasz-walkiewicz:
Thanks @matialm for this PR but I don't understand what you're trying to fix here. Could you please try to explain it in the description and/or ideally add a test for the issue you're trying to fix?

matialm:
> Thanks @matialm for this PR but I don't understand what you're trying to fix here. Could you please try to explain it in the description and/or ideally add a test for the issue you're trying to fix?

Sorry for the late response, it fix the issue in this discussion:
https://github.com/trinodb/trino-python-client/discussions/437

When I debug it I find it that the oauth2 redirect url callback wasn't been fired because it wasn't retriving the x_redirect_server parameter correctly. Long story short, when they were trying to remove the prefix "Bearer" from the headers because the "count=1" parameter it was removing that prefix only for the parameter x_token_server and not from x_redirect_server.

mahic:
I can confirm an issue with callback not firing because of `count=1` with Trino 427 with a basic OAuth setup verified working with DBeaver. It seems like the response from the server does not contain count=1 in the header for that particular version.

By removing `count=1` on a debug install of 0.328 the callback fires. What does the counter do and was it added to Trino in a recent release?

hashhar:
It's entirely unexpected. The auth challenge header format is standardised - see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate.

In general it's of the form `<auth-scheme> auth-param1=token1, ..., auth-paramN=auth-paramN-token` where `<auth-scheme>` = `Bearer`, `x_redirect_server` and `x_token_server` are `auth-param`s.

Can you try to log the value of the headers?

hashhar:
Figured out the possible issue but will need you to share the values of the WWW-Authenticate headers that you see.

One thing which can happen is that the IdP supports multiple auth mechanisms and hence repsonds with multiple challenges - either within same `WWW-Authenticate` header or multiple occurrences of that header. Neither case is handled correctly today.

We need to change the parsing logic for www-authenticate header. This proposed fix is incorrect either way.

mahic:
@hashhar 

Our Idp is ForgeRock
This is the 401 Headers returned from the IdP:

```
HTTP/1.1 401 Unauthorized
date: Fri, 08 Mar 2024 15:56:27 GMT
content-type: text/plain;charset=utf-8
www-authenticate: Basic realm="Trino"
www-authenticate: Bearer realm="Trino", token_type="JWT"
www-authenticate: Bearer x_redirect_server="https://trino.ourdomain.com/oauth2/token/initiate/9e7e5766b1603b9151aa6712f995915658b990c18a6f9a7b350de889651f1ad6", x_token_server="https://trino.ourdomain.com/oauth2/token/c3ade6f9-2a19-1e26-a4a7-0102dbbe8022"
content-length: 12
x-envoy-upstream-service-time: 1
server: envoy
```

This works fine with Trino JDBC (used in DBeaver) 🤷‍♂️

matialm:
This days I'm out but next week I will share an example of the problem I'm refering to.

hashhar:
> Seems you're right about the multiple auth-mechanisms, but this works fine with Trino JDBC (used in DBeaver) 🤷‍♂️

Yes, the JDBC driver uses an HTTP library (OkHttp) which is able to return all list of challenges. Python requests collapses multi-valued HTTP headers into a single header with all values comma-separated. That's why changing the `count` for `re.sub` works. It'll also however break stuff if the sequence of characters `bearer` happens to appear anywhere in the URLs or tokens too.

Thanks for confirming the issue. That itself makes it much easier to test the fixes.

hashhar:
Fixed by #467

PR #446, opened 2024-02-03
Type hints for types and mapper modules

hashhar:
## Description

Add type hints for mapper and types modules and enables mypy there.

This uncovered some dead code / possibly broken code regarding handling of ambiguous date-time values. This will be fixed separately in a follow-up.

## Release notes

### DIFF
--- a/trino/mapper.py
+++ b/trino/mapper.py
@@ -56,7 +56,7 @@ class IntegerValueMapper(ValueMapper[int]):
 
 
 class DoubleValueMapper(ValueMapper[float]):
-    def map(self, value) -> Optional[float]:
+    def map(self, value: Any) -> Optional[float]:
         if value is None:
             return None
         if value == 'Infinity':
@@ -69,7 +69,7 @@ class DoubleValueMapper(ValueMapper[float]):
 
 
 class DecimalValueMapper(ValueMapper[Decimal]):
-    def map(self, value) -> Optional[Decimal]:
+    def map(self, value: Any) -> Optional[Decimal]:
         if value is None:
             return None
         return Decimal(value)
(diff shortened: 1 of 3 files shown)


trino/types.py
```diff
@@ -101,16 +102,18 @@ def new_instance(self, value: datetime, fraction: Decimal) -> TimestampWithTimeZ
 
     def normalize(self, value: datetime) -> datetime:
         if tz.datetime_ambiguous(value):
-            return self._whole_python_temporal_value.tzinfo.normalize(value)
+            # This appears to be dead code since tzinfo doesn't actually have a `normalize` method.
+            # TODO: Fix this or remove.
```
hovaesco, line 106: Please create an issue.

trino/mapper.py
```diff
@@ -175,37 +176,38 @@ class ArrayValueMapper(ValueMapper[List[Optional[Any]]]):
...
         self.mappers = mappers
         self.names = names
         self.types = types
 
-    def map(self, values: List[Any]) -> Optional[Tuple[Optional[Any], ...]]:
-        if values is None:
+    def map(self, value: Optional[List[Any]]) -> Optional[Tuple[Optional[Any], ...]]:
+        if value is None:
             return None
+        # TODO: Fix typing for self.names
```
hovaesco, line 207: another TODO

hovaesco (approved):
LGTM % what's the plan for TODOs?

trino/types.py
hashhar: I plan to send a follow-up PR for this.

Looks like it'll be a bugfix. Seems that the ambiguous date-time handling hasn't been working since the beginning.
This code became "broken" (ie. calling non-existent method) in https://github.com/trinodb/trino-python-client/commit/2b9ca0c2e59496e168907fec1df7a50f26976dce#diff-cf9b945b8cb541d38680c39caa2de6568793a6d53644d6f2659de6a77ff96127R1040

I checked both before and after that commit/PR and the code isn't called by any tests.

Will create an issue to track.

trino/mapper.py
hashhar: oh, this is easy to fix in the commit which adds typing for types.py.

Let me fix and remove TODO instead. I left it as a note for myself but forgot.

trino/types.py
hashhar: https://github.com/trinodb/trino-python-client/issues/449

hashhar:
@hovaesco Addressed comments. Only last commit has changed.

hovaesco (approved):

PR #453, opened 2024-02-16
CI updates

hashhar:
## Description

- Update GitHub action versions
- Add pip caching for CI workflows, release workflow doesn't use caching still just for safety/clean build reasons

## Release notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Checkout the source code"
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: "Install Python"
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: "3.11"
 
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -20,12 +20,13 @@ jobs:
...
       - name: "Checkout the source code"
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: "Install Python"
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: "3.11"
+          cache: "pip"
```
nineinchnick, line 29: I'm not sure if caching can be used without `requirements.txt`, maybe it's not worth anyway?

hashhar: it works with setup.py too but I forgot to add the params. Thanks.

hashhar: Eventually we plan to move to pyproject.toml now that setup.py is going to be deprecated soon.

hashhar: saves ~1m per job for total of 11 mins across a run. I'll take what I can get.

hovaesco (approved):

PR #454, opened 2024-02-21
Add GSSAPIAuthentication authentication class.

gkarg:
## Description

Add yet another Authentication method, via `requests-gssapi`. The provided class, `GSSAPIAuthentication` is complete copy-paste of `KerberosAuthentication`, but uses a much nicer library (`requests-gssapi`) underneath.

## Non-technical explanation

Added support for alternative method of kerberos authentication.

## Release notes

```markdown
* Add requests-gssapi based authentication via `GSSAPIAuthentication`. ({issue}`454`)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ with open(os.path.join(here, "README.md"), "r", "utf-8") as f:
     readme = f.read()
 
 kerberos_require = ["requests_kerberos"]
-gssapi_require = ["requests_kerberos"]
+gssapi_require = ["requests_gssapi"]
 sqlalchemy_require = ["sqlalchemy >= 1.3"]
 external_authentication_token_cache_require = ["keyring"]
 
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,7 @@ with open(os.path.join(here, "README.md"), "r", "utf-8") as f:
     readme = f.read()
 
 kerberos_require = ["requests_kerberos"]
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -27,6 +27,7 @@
     readme = f.read()
 
 kerberos_require = ["requests_kerberos"]
+gssapi_require = ["requests_kerberos"]
```
hashhar, line 30: this is same as `kerberos_require`. Why is this needed separately?

hashhar (commented):
skimmed, Kerberos needs more attention so I'll revisit this.

setup.py
gkarg: spot on, this is a typo. fixed

setup.py
```diff
@@ -27,6 +27,7 @@
     readme = f.read()
 
 kerberos_require = ["requests_kerberos"]
+gssapi_require = ["requests_gssapi"]
```
hovaesco, line 30: add it to `all_require` maybe? just for sake of completeness

hovaesco (approved):

setup.py
gkarg: that's easy to add, but just to be clear: although those libraries do not conflict with each other, I would assume an end-user would want either one either another, but rarely both.

hashhar (approved):

hashhar:
@cla-bot check

PR #455, opened 2024-03-01
Add retry on 429 error code

hovaesco:
## Description

Add retry on 429 error code

## Release notes

```markdown
```

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -507,9 +507,9 @@ class TrinoRequest(object):
             self._handle_retry,
             handled_exceptions=self._exceptions,
             conditions=(
-                # need retry when there is no exception but the status code is 502, 503, or 504
+                # need retry when there is no exception but the status code is 429, 502, 503, or 504
                 lambda response: getattr(response, "status_code", None)
-                in (502, 503, 504),
+                in (429, 502, 503, 504),
             ),
             max_attempts=self._max_attempts,
         )


trino/client.py
```diff
@@ -507,9 +507,9 @@ def max_attempts(self, value) -> None:
             self._handle_retry,
             handled_exceptions=self._exceptions,
             conditions=(
-                # need retry when there is no exception but the status code is 502, 503, or 504
+                # need retry when there is no exception but the status code is 429, 502, 503, or 504
                 lambda response: getattr(response, "status_code", None)
-                in (502, 503, 504),
+                in (429, 502, 503, 504),
```
lshrinivas, line 512: Do we need to add any delays for the retry? Immediate retry will just make the problem worse (not sure if that's already built in)

electrum (dismissed):
This isn’t part of the Trino protocol. Before we add this here, it needs to be specified in the protocol, which will require more details. How long should the client wait? Is there a standard header for this?

electrum:
This behavior needs to be specified in https://trino.io/docs/current/develop/client-protocol.html

There is a standard `Retry-After` header.

trino/client.py
```diff
@@ -887,7 +895,11 @@ def decorated(*args, **kwargs):
                 try:
                     result = func(*args, **kwargs)
                     if any(guard(result) for guard in conditions):
-                        handle_retry.retry(func, args, kwargs, None, attempt)
+                        if result.status_code == 429 and "Retry-After" in result.headers:
+                            handle_retry_sleep = _RetryAfterSleep(result.headers.get("Retry-After"))
```
hashhar, line 899: Add a method to convert header value to seconds. The header can be a point-in-time too. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After#syntax in which case this code will throw.

hovaesco: I've added a method `_parse_retry_after_header` for parsing `Retry-After` header.

hashhar (approved):
LGTM for the client.

As David said earlier we should document the expectation from other clients for HTTP 429 in the Trino client protocol docs. Please address that in a follow-up.

hashhar (approved):
Trino protocol docs were updated. So this is now good to merge.

PR #462, opened 2024-05-07
Support object as value in extra_credential

huw0:
## Description

I want `extra_credential` to accept an object as a value. This is helpful because it allows pass through of short lived tokens (JWT or otherwise) that can be refreshed by the client without having to generate a new engine/client object.

Presently this is not possible because `parse_quote` does a [check](https://github.com/python/cpython/blob/de1428f8c234a8731ced99cbfe5cd6c5c719e31d/Lib/urllib/parse.py#L916) on the object type.

Therefore it is necessary to call `str()` on the `extra_credential` value during serialisation.

## Non-technical explanation

Support object as value in `extra_credential` to support refreshing tokens.

## Release notes

```markdown
* Support object as value in `extra_credential` to support refreshing tokens.
```

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -867,6 +867,41 @@ def test_extra_credential_value_encoding(mock_get_and_post):
     assert headers[constants.HEADER_EXTRA_CREDENTIAL] == "foo=bar+%E7%9A%84"
 
 
+def test_extra_credential_value_object(mock_get_and_post):
+    _, post = mock_get_and_post
+
+    class TestCredential(object):
+        value = "initial"
+
+        def __str__(self):
+            return self.value
+
+    credential = TestCredential()
+
+    req = TrinoRequest(
+        host="coordinator",
+        port=constants.DEFAULT_TLS_PORT,
+        client_session=ClientSession(
+            user="test",
+            extra_credential=[("foo", credential)]
+        )
+    )
+
+    req.post("SELECT 1")
+    _, post_kwargs = post.call_args
+    headers = post_kwargs["headers"]
+    assert constants.HEADER_EXTRA_CREDENTIAL in headers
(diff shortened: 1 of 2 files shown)


hovaesco (approved):

huw0:
Thanks for approving.

@hovaesco, @hashhar is this good to merge?

tests/unit/test_client.py
```diff
@@ -864,6 +864,29 @@ def test_extra_credential_value_encoding(mock_get_and_post):
...
+    class TestCredential(object):
+        def __str__(self):
+            return "bar"
+
+    req = TrinoRequest(
+        host="coordinator",
+        port=constants.DEFAULT_TLS_PORT,
+        client_session=ClientSession(
+            user="test",
+            extra_credential=[("foo", TestCredential())]
```
hashhar, line 879: I get why you made this change, makes it easier to use object as cred values. But how is this different from below? Just want to understand what problem it actually solves.

```
extra_credential=[("foo"), str(TestCredential())]
```

hashhar (approved):
Change looks good, just had a question.

tests/unit/test_client.py
huw0: If `TestCredential()` provides a time limited credential (JWT or otherwise) then the object can refresh the underlying credential. This has two benefits:

1. No need to regenerate TrinoClient object.
1. If the time-limited credential expiry is hit, the automatic retry should include the refreshed credential

huw0: Somehow I missed that your comment was on the test! I've added a second assertion which hopefully is both a more thorough test and shows the difference clearly.

huw0:
I've rebased the branch and improved the unit test.

huw0:
@hashhar @hovaesco - is this ok to merge? :)

hashhar (approved):

hashhar:
Sorry for the delay here. Thanks for the contribution.

Issue #444, opened 2024-01-31
Cannot connect to Trino via OAUTH2 with SSO credentials


matialm:
### Discussed in https://github.com/trinodb/trino-python-client/discussions/437

<div type='discussions-op-text'>

<sup>Originally posted by **lejanco** January 18, 2024</sup>
I am trying to connect to trino via oauth2 using the externalauthentication instructions in the README(e.g. keyring). I have tried (what seems like) everything. You can see the commented out efforts here as well.  I have done the pip install 'trino[<keyring>]'

` trino_connection = trino.dbapi.connect(host=host,
                                               port=port,
                                               user=username,
                                               catalog=catalog,
                                               schema=schema,
                                               auth=trino.auth.OAuth2Authentication(
                                                   # redirect_auth_url_handler=redirect_auth_url_handler
                                               ),
                                               http_scheme=http_scheme,
                                               # verify=False,
                                               # extra_credential=[('user', username),
                                               #                   ('password', password),
                                               #                   ("SSL", "true"),
                                               #                   ("SSLVerification", "NONE"),
                                               #                   ("externalAuthentication", "true")
                                               #                   ]
                                               )`

The debug suggests the token is being retrieved but not being authenticated. Help is greatly appreciated, but please limit suggestions that require admin rights to trino as I do not have that authorization. Thanks:

DEBUG:
</div>

matialm:
PR created
https://github.com/trinodb/trino-python-client/pull/445

hashhar:
Useful discussion about what the root cause was can be found at https://github.com/trinodb/trino-python-client/pull/445#issuecomment-1982806130

PR #467, opened 2024-06-27
Apply new logic for parsing WWW-Authenticate header

hovaesco:
## Description
Resolves https://github.com/trinodb/trino-python-client/issues/444

## Release notes

```markdown
```

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -539,6 +539,8 @@ def test_oauth2_authentication_missing_headers(header, error):
     'Bearer x_token_server="{token_server}", x_redirect_server="{redirect_server}"',
     'Basic realm="Trino", Bearer x_redirect_server="{redirect_server}", x_token_server="{token_server}"',
     'Bearer x_redirect_server="{redirect_server}", x_token_server="{token_server}", Basic realm="Trino"',
+    'Basic realm="Trino", Bearer realm="Trino", token_type="JWT", Bearer x_redirect_server="{redirect_server}", '
+    'x_token_server="{token_server}"'
 ])
 @httprettified
 def test_oauth2_header_parsing(header, sample_post_response_data):
--- a/trino/auth.py
+++ b/trino/auth.py
@@ -22,7 +22,6 @@ from urllib.parse import urlparse
 
 from requests import PreparedRequest, Request, Response, Session
 from requests.auth import AuthBase, extract_cookies_to_jar
-from requests.utils import parse_dict_header
 
 import trino.logging
 from trino.client import exceptions
@@ -421,10 +420,13 @@ class _OAuth2TokenBearer(AuthBase):
         if not _OAuth2TokenBearer._BEARER_PREFIX.search(auth_info):
             raise exceptions.TrinoAuthError(f"Error: header info didn't match {auth_info}")
 
-        auth_info_headers = parse_dict_header(
-            _OAuth2TokenBearer._BEARER_PREFIX.sub("", auth_info, count=1))  # type: ignore
+        # Example www-authenticate header value:
+        # 'Basic realm="Trino", Bearer realm="Trino", token_type="JWT",
+        # Bearer x_redirect_server="https://trino.com/oauth2/token/uuid4",
+        # x_token_server="https://trino.com/oauth2/token/uuid4"'
+        auth_info_headers = self._parse_authenticate_header(auth_info)
 
-        auth_server = auth_info_headers.get('x_redirect_server')
+        auth_server = auth_info_headers.get('bearer x_redirect_server', auth_info_headers.get('x_redirect_server'))
         token_server = auth_info_headers.get('x_token_server')
         if token_server is None:
             raise exceptions.TrinoAuthError("Error: header info didn't have x_token_server")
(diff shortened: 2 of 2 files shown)


trino/auth.py
```diff
@@ -421,10 +420,21 @@ def _attempt_oauth(self, response: Response, **kwargs: Any) -> None:
         if not _OAuth2TokenBearer._BEARER_PREFIX.search(auth_info):
             raise exceptions.TrinoAuthError(f"Error: header info didn't match {auth_info}")
 
-        auth_info_headers = parse_dict_header(
-            _OAuth2TokenBearer._BEARER_PREFIX.sub("", auth_info, count=1))  # type: ignore
+        split_challenge = auth_info.split(" ", 1)
```
hashhar, line 423: a comment here with an example header value would help. Consider extracting a method/class for this and unit-testing that against some headers you've seen while local testing (and the one mentioned in the issue).

hashhar: The example from the other PR is (comma to join each header value):

```
www-authenticate: Basic realm="Trino", Bearer realm="Trino", token_type="JWT", Bearer x_redirect_server="https://trino.ourdomain.com/oauth2/token/initiate/9e7e5766b1603b9151aa6712f995915658b990c18a6f9a7b350de889651f1ad6", x_token_server="https://trino.ourdomain.com/oauth2/token/c3ade6f9-2a19-1e26-a4a7-0102dbbe8022"
```

So I see this logic would set:

```
split_challenge = ['Basic', 'realm="Trino", Bearer realm="Trino", token_type="JWT", Bearer x_redirect_server="https://trino.ourdomain.com/oauth2/token/initiate/9e7e5766b1603b9151aa6712f995915658b990c18a6f9a7b350de889651f1ad6", x_token_server="https://trino.ourdomain.com/oauth2/token/c3ade6f9-2a19-1e26-a4a7-0102dbbe8022"']
self.scheme = 'Basic'
trimmed_challenge = 'realm="Trino", Bearer realm="Trino", token_type="JWT", Bearer x_redirect_server="https://trino.ourdomain.com/oauth2/token/initiate/9e7e5766b1603b9151aa6712f995915658b990c18a6f9a7b350de889651f1ad6", x_token_server="https://trino.ourdomain.com/oauth2/token/c3ade6f9-2a19-1e26-a4a7-0102dbbe8022"'
```

This confuses me because scheme is actually 3 values here - `Basic`, `Bearer` and `Bearer` but we only pick the first one.

trino/auth.py
```diff
@@ -421,10 +420,21 @@ def _attempt_oauth(self, response: Response, **kwargs: Any) -> None:
...
-        auth_server = auth_info_headers.get('x_redirect_server')
+        auth_info_headers = {}
+
+        for item in trimmed_challenge.split(","):
+            comps = item.split("=")
+            if len(comps) == 2:
+                key = comps[0].strip(' "')
+                value = comps[1].strip(' "')
+                if key:
+                    auth_info_headers[key.lower()] = value
```
hashhar, line 435: the dict would overwrite if same key is present (e.g. if there are multiple `Bearer realm` entries or multiple `Bearer x_redirect_server`).

trino/auth.py
hovaesco: Added an example header. Extracted logic to a method and added some testing. Scheme is not used, so I dropped it from the code.

trino/auth.py
hovaesco: I don't think it's possible based on https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

hashhar: The entire challenge cannot be a duplicate I think but here the `key` is the part split on `=`.

So the example at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#digest_authentication_with_sha-256_and_md5 has two challenges for which `key = 'Digest realm'`.

(Sorry I'm AFK and did not test the static method against that example 🙏).

tests/unit/test_client.py
```diff
@@ -539,6 +539,8 @@ def test_oauth2_authentication_missing_headers(header, error):
     'Bearer x_token_server="{token_server}", x_redirect_server="{redirect_server}"',
     'Basic realm="Trino", Bearer x_redirect_server="{redirect_server}", x_token_server="{token_server}"',
     'Bearer x_redirect_server="{redirect_server}", x_token_server="{token_server}", Basic realm="Trino"',
+    'realm="Trino", Bearer realm="Trino", token_type="JWT", Bearer x_redirect_server="{redirect_server}", '
```
hashhar, line 542:
```suggestion
    'Basic realm="Trino", Bearer realm="Trino", token_type="JWT", Bearer x_redirect_server="{redirect_server}", '
```

did you mean this?

hashhar (commented):
Looks good % question about duplicate keys

tests/unit/test_client.py
hovaesco: yes, changed

hashhar (approved):
works for our use-case where we just want `x_redirect_server` and `x_token_server` value. Sadly Python is not enterprise enough to have good quality libraries for HTTP 401 handling.

PR #478, opened 2024-08-16
Fix `cert` and `key` kwargs parsing from URL in sqlachemy's `create_connect_args` function

a-narsudinov:
## Description

Previous version of code contained an always true condition that may lead to unintended behaviour. This bug was captured by flake8 linter, so it may be a good idea to include it as a part of CI process.

## Release notes

```markdown
* Fix `cert` and `key` kwargs parsing from URL in sqlachemy's `create_connect_args` function
```

### DIFF
--- a/tests/unit/sqlalchemy/test_dialect.py
+++ b/tests/unit/sqlalchemy/test_dialect.py
@@ -298,6 +298,23 @@ def test_trino_connection_certificate_auth():
     assert cparams['auth']._key == key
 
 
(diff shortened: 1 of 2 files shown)


hovaesco (approved):

hashhar:
@cla-bot check

hashhar (approved):
I took the liberty to add a test. Thanks for the fix.

I'll merge once CLA is processed. It gets processed manually every 2 weeks.

PR #495, opened 2024-11-12
Fix parsing authentication header

mdesmet:
## Description

Currently the function to parse the authorization header doesn't support urls containing parameters. This PR addresses this.

## Release notes

```markdown
* Fix authentication parsing to allow urls containing url parameters
```

### DIFF
--- a/trino/auth.py
+++ b/trino/auth.py
@@ -546,17 +546,17 @@ class _OAuth2TokenBearer(AuthBase):
 
     @staticmethod
     def _parse_authenticate_header(header: str) -> Dict[str, str]:
-        split_challenge = header.split(" ", 1)
-        trimmed_challenge = split_challenge[1] if len(split_challenge) > 1 else ""
+        logger.debug(f"Authentication header: {header}")
+        components = header.split(",")
         auth_info_headers = {}
 
-        for item in trimmed_challenge.split(","):
-            comps = item.split("=")
-            if len(comps) == 2:
-                key = comps[0].strip(' "')
-                value = comps[1].strip(' "')
-                if key:
-                    auth_info_headers[key.lower()] = value
+        for component in components:
+            component = component.strip()
+            if "=" in component:
+                key, value = component.split("=", 1)
+                if value[0] == '"' and value[-1] == '"':
+                    value = value[1:-1]
+                auth_info_headers[key.lower()] = value
         return auth_info_headers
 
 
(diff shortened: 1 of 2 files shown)


trino/auth.py
```diff
@@ -546,17 +546,15 @@ def _construct_cache_key(host: Optional[str], user: Optional[str]) -> Optional[s
...
-                value = comps[1].strip(' "')
-                if key:
-                    auth_info_headers[key.lower()] = value
+        for component in components:
+            if "=" in component:
+                key, value = component.split("=", 1)
+                if value[0] == '"' and value[-1] == '"':
+                    value = value[1:-1]
+                auth_info_headers[key.lower()] = value
         return auth_info_headers
```
mdesmet, line 558: We could add a bit more exception handling here as well in case the headers don't return the expected values.

trino/auth.py
```diff
@@ -546,17 +546,15 @@ def _construct_cache_key(host: Optional[str], user: Optional[str]) -> Optional[s
 
     @staticmethod
     def _parse_authenticate_header(header: str) -> Dict[str, str]:
-        split_challenge = header.split(" ", 1)
-        trimmed_challenge = split_challenge[1] if len(split_challenge) > 1 else ""
+        components = header.split(", ")
```
hashhar, line 549: is the whitespace guaranteed to be present?

trino/auth.py
```diff
@@ -546,17 +546,15 @@ def _construct_cache_key(host: Optional[str], user: Optional[str]) -> Optional[s
...
-            if len(comps) == 2:
-                key = comps[0].strip(' "')
-                value = comps[1].strip(' "')
-                if key:
-                    auth_info_headers[key.lower()] = value
+        for component in components:
+            if "=" in component:
+                key, value = component.split("=", 1)
+                if value[0] == '"' and value[-1] == '"':
+                    value = value[1:-1]
```
hashhar, line 556: nice, much safer than older strip-based.

trino/auth.py
hashhar: in the worst case we should make sure enabling DEBUG logs will show the header values (raw values before we try extraction).

hashhar (commented):
looks good % question about whitespace

trino/auth.py
mdesmet: In the context of the WWW-Authenticate header as defined by RFC 7235 (which governs HTTP Authentication) and RFC 2616 for HTTP/1.1, the comma followed by a space is not strictly required. Both a simple comma (,) and a comma followed by a space (, ) are acceptable, though the latter is commonly used for readability.

trino/auth.py
```diff
@@ -546,17 +546,17 @@ def _construct_cache_key(host: Optional[str], user: Optional[str]) -> Optional[s
...
-        for item in trimmed_challenge.split(","):
-            comps = item.split("=")
-            if len(comps) == 2:
-                key = comps[0].strip(' "')
-                value = comps[1].strip(' "')
-                if key:
-                    auth_info_headers[key.lower()] = value
+        for component in components:
+            component = component.strip()
+            if "=" in component.strip():
```
hashhar, line 555: why `strip` twice?

hashhar (approved):

PR #496, opened 2024-11-22
Add support for the Trino client spooling protocol

mdesmet:
## Release notes

(*) Release notes are required, with the following suggested text:

```markdown
* Add support for the Trino client spooling protocol
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,6 +469,30 @@ conn = connect(
 )
 ```
 
+## Spooled protocol
+
+The client spooling protocol requires [a Trino server with spooling protocol support](https://trino.io/docs/current/client/client-protocol.html#spooling-protocol).
+
+Enable the spooling protocol by specifying a supported encoding in the `encoding` parameter:
+
+```python
+from trino.dbapi import connect
+
+conn = connect(
+    encoding="json+zstd"
+)
+```
+
+or a list of supported encodings:
+
+```python
+from trino.dbapi import connect
+
+conn = connect(
+    encoding=["json+zstd", "json"]
+)
+```
+
 ## Transactions
 
 The client runs by default in *autocommit* mode. To enable transactions, set
--- a/etc/config.properties
+++ b/etc/config.properties
@@ -1,0 +1,17 @@
+node.id=coordinator
+node.environment=test
+
+coordinator=true
+experimental.concurrent-startup=true
+node-scheduler.include-coordinator=true
+http-server.http.port=8080
+query.max-memory=1GB
+discovery.uri=http://localhost:8080
+
+# spooling protocol settings
+protocol.spooling.enabled=true
+protocol.spooling.shared-secret-key=jxTKysfCBuMZtFqUf8UJDQ1w9ez8rynEJsJqgJf66u0=
+protocol.spooling.retrieval-mode=coordinator_proxy
+
+# Disable http request log
+http-server.log.enabled=false
--- a/etc/spooling-manager.properties
+++ b/etc/spooling-manager.properties
@@ -1,0 +1,8 @@
+spooling-manager.name=filesystem
+fs.s3.enabled=true
+fs.location=s3://spooling/
+s3.endpoint=http://localstack:4566/
+s3.region=us-east-1
+s3.aws-access-key=test
+s3.aws-secret-key=test
+s3.path-style-access=true
--- a/tests/development_server.py
+++ b/tests/development_server.py
@@ -1,0 +1,127 @@
+import os
+import time
+from contextlib import contextmanager
+from pathlib import Path
+
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.network import Network
+from testcontainers.core.waiting_utils import wait_for_logs
+from testcontainers.localstack import LocalStackContainer
+
+from trino.constants import DEFAULT_PORT
+
+MINIO_ROOT_USER = "minio-access-key"
+MINIO_ROOT_PASSWORD = "minio-secret-key"
+
+TRINO_VERSION = os.environ.get("TRINO_VERSION") or "latest"
+TRINO_HOST = "localhost"
+
+
+def create_bucket(s3_client):
+    bucket_name = "spooling"
+    try:
+        print("Checking for bucket existence...")
+        response = s3_client.list_buckets()
+        buckets = [bucket["Name"] for bucket in response["Buckets"]]
+        if bucket_name in buckets:
+            print("Bucket exists!")
+            return
+    except s3_client.exceptions.ClientError as e:
+        if not e.response['Error']['Code'] == '404':
+            print("An error occurred:", e)
+            return
+
+    try:
+        print("Creating bucket...")
+        s3_client.create_bucket(
+            Bucket=bucket_name,
+        )
+        print("Bucket created!")
+    except s3_client.exceptions.ClientError as e:
+        print("An error occurred:", e)
+
+
+@contextmanager
+def start_development_server(port=None, trino_version=TRINO_VERSION):
+    network = None
+    localstack = None
+    trino = None
+
+    try:
+        network = Network().create()
+        supports_spooling_protocol = TRINO_VERSION == "latest" or int(TRINO_VERSION) >= 466
+        if supports_spooling_protocol:
+            localstack = LocalStackContainer(image="localstack/localstack:latest", region_name="us-east-1") \
+                .with_name("localstack") \
+                .with_network(network) \
+                .with_bind_ports(4566, 4566) \
+                .with_bind_ports(4571, 4571) \
+                .with_env("SERVICES", "s3")
+
+            # Start the container
+            print("Starting LocalStack container...")
+            localstack.start()
+
+            # Wait for logs indicating MinIO has started
+            wait_for_logs(localstack, "Ready.", timeout=30)
+
+            # create spooling bucket
+            create_bucket(localstack.get_client("s3"))
+
+        trino = DockerContainer(f"trinodb/trino:{trino_version}") \
+            .with_name("trino") \
+            .with_network(network) \
+            .with_env("TRINO_CONFIG_DIR", "/etc/trino") \
+            .with_bind_ports(DEFAULT_PORT, port)
+
+        root = Path(__file__).parent.parent
+
+        trino = trino \
+            .with_volume_mapping(str(root / "etc/catalog"), "/etc/trino/catalog")
+
+        # Enable spooling config
+        if supports_spooling_protocol:
+            trino \
+                .with_volume_mapping(
+                    str(root / "etc/spooling-manager.properties"),
+                    "/etc/trino/spooling-manager.properties", "rw") \
+                .with_volume_mapping(str(root / "etc/jvm.config"), "/etc/trino/jvm.config") \
+                .with_volume_mapping(str(root / "etc/config.properties"), "/etc/trino/config.properties")
+        else:
+            trino \
+                .with_volume_mapping(str(root / "etc/jvm-pre-466.config"), "/etc/trino/jvm.config") \
+                .with_volume_mapping(str(root / "etc/config-pre-466.properties"), "/etc/trino/config.properties")
+
+        print("Starting Trino container...")
+        trino.start()
+
+        # Wait for logs indicating the service has started
+        wait_for_logs(trino, "SERVER STARTED", timeout=60)
+
+        # Otherwise some tests fail with No nodes available
+        time.sleep(2)
+
+        yield localstack, trino, network
+    finally:
+        # Stop containers when exiting the context
+        if trino:
+            print("Stopping Trino container...")
+            trino.stop()
+        if localstack:
+            print("Stopping LocalStack container...")
+            localstack.stop()
+        if network:
+            network.remove()
+
+
+def main():
+    """Run Trino setup independently from pytest."""
+    with start_development_server(port=DEFAULT_PORT):
+        print(f"Trino started at {TRINO_HOST}:{DEFAULT_PORT}")
+
+        # Keep the process running so that the containers stay up
+        input("Press Enter to stop containers...")
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -40,7 +40,7 @@ from trino.transaction import IsolationLevel
 
 @pytest.fixture
 def trino_connection(run_trino):
-    _, host, port = run_trino
+    host, port = run_trino
 
     yield trino.dbapi.Connection(
         host=host, port=port, user="test", source="test", max_attempts=1
@@ -49,7 +49,7 @@ def trino_connection(run_trino):
 
 @pytest.fixture
 def trino_connection_with_transaction(run_trino):
-    _, host, port = run_trino
+    host, port = run_trino
 
     yield trino.dbapi.Connection(
         host=host,
@@ -63,7 +63,7 @@ def trino_connection_with_transaction(run_trino):
 
 @pytest.fixture
 def trino_connection_in_autocommit(run_trino):
-    _, host, port = run_trino
+    host, port = run_trino
 
     yield trino.dbapi.Connection(
         host=host,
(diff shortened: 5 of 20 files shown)


mdesmet:
Currently tests are failing because of https://github.com/trinodb/trino/pull/24226

trino/client.py
```diff
@@ -925,3 +948,103 @@ def _parse_retry_after_header(retry_after):
...
+
+
+# Trino Spooled protocol transfer objects
+SpooledSegmentMetadata = TypedDict('SpooledSegmentMetadata', {'uncompressedSize': str})
+SpooledSegment = TypedDict(
+    'SpooledSegment',
+    {
+        'type': str,
+        'uri': str,
+        'ackUri': NotRequired[str],
```
aalbu, line 960: I believe that in the final version of the protocol, this has become required (I know that it could be missing in some earlier versions of Trino).

wendigo: yes, it is required

trino/client.py
```diff
@@ -46,10 +48,13 @@
...
 from email.utils import parsedate_to_datetime
 from time import sleep
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterator, List, Literal, Optional, Tuple, Union, cast
 from zoneinfo import ZoneInfo
 
+import lz4.block
 import requests
+import zstandard
+from typing_extensions import TypedDict
```
damian3031, line 57: `TypedDict` was added to the `typing` module in Python 3.8. We can use standard library instead of introducing a new dependency on `typing_extensions`

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+        super().__init__(segment)
+        self._segment = cast(_InlineSegmentTO, segment)
+        self._row_mapper = row_mapper
+
+    @property
+    def rows(self) -> List[List[Any]]:
+        inline_segment = self._segment
+        data = inline_segment["data"]
+        decoded_string = base64.b64decode(data)
+        rows = self._row_mapper.map(json.loads(decoded_string))
```
wendigo, line 1056: we should probably combine _row_mapper with the decoder logic so that it accepts

_row_mapper.map(encoding: str, data: str)

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+    """
+    def __init__(
+        self,
+        response: _SpooledProtocolResponseTO,
+        segment: _SpooledSegmentTO,
+        row_mapper: RowMapper,
+        request: TrinoRequest
+    ) -> None:
+        super().__init__(segment)
+        self._segment = cast(_SpooledSegmentTO, segment)
```
wendigo, line 1086: do we need to cast?

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+        self._row_mapper = row_mapper
```
wendigo, line 1087: we should probably construct row_mapper for a given encoding

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+    @property
+    def ack_uri(self) -> str:
+        return self._segment["ackUri"]
+
+    @property
+    def headers(self) -> Dict[str, List[str]]:
+        return self._segment.get("headers", {})
+
+    def acknowledge(self):
+        http_response = self._send_spooling_request(self._segment["ackUri"], self._segment)
```
wendigo, line 1109: is this async? it should be, as this can block and we don't want that (also it's fine for this request to fail, this is best effort)

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+    def _load_spooled_segment(self) -> str:
+        segment = self._segment
+        http_response = self._send_spooling_request(segment["uri"], segment)
+        if not http_response.ok:
+            self._request.raise_response_error(http_response)
+        content = http_response.content
+
+        encoding = self._encoding
+        if encoding == "json+zstd":
+            zstd_decompressor = zstandard.ZstdDecompressor()
```
wendigo, line 1122: move this logic to the row_mapper

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+            return zstd_decompressor.decompress(content).decode('utf-8')
+        elif encoding == "json+lz4":
+            expected_size = segment["metadata"]["uncompressedSize"]
```
wendigo, line 1125: not every json+lz4 is compressed, if the data is below certain threshold, it can stay uncompressed. the algorithm is:

```
if exists metadata["uncompressedSize"]:
   decompress(data)
else
   data
```

you should also validate whether the data after decompression is `uncompressedSize`
and whether the retrieved data is `segmentSize`

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+            "GET",
+            uri,
+            headers=list(map(lambda item: [(item[0], value) for value in item[1]], segment.get("headers", {}).items()))
+        )
+        prepared = self._request._prepare_request(req)
+        return self._request._send(prepared)
+
+    def __repr__(self):
+        return (
+            f"SpooledSegment(metadata={self.metadata}, uri={self.uri}), "
```
wendigo, line 1143: do not log `uri` and `ackUri` and `headers` - these are sensitive and we should avoid surfacing these

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+            try:
+                return next(self._rows)
+            except StopIteration:
+                if self._current_segment and isinstance(self._current_segment, SpooledSegment):
+                    self._current_segment.acknowledge()
+                if self._finished:
+                    raise StopIteration
+                self._load_next_row_set()
+
+    def _load_next_row_set(self):
```
wendigo, line 1201: nit: _load_next_segment

tests/integration/test_dbapi_integration.py
```diff
@@ -1852,6 +1852,29 @@ def test_select_query_spooled_segments(trino_connection):
         assert isinstance(row[13], str), f"Expected string for shipinstruct, got {type(row[13])}"
 
 
+@pytest.mark.skipif(
+    trino_version() <= '464',
```
wendigo, line 1856: let's make it 466 as it has a stable/final protocol definition

wendigo (commented):
Overall LGTM except for decoding logic that needs to be improved

etc/config.properties
```diff
@@ -0,0 +1,22 @@
+node.id=coordinator
+node.environment=test
+
+coordinator=true
+experimental.concurrent-startup=true
```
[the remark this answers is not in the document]
wendigo: @mosabua spooling protocol

tests/integration/test_dbapi_integration.py
```diff
@@ -1821,8 +1822,61 @@ def test_prepared_statement_capability_autodetection(legacy_prepared_statements,
     assert statements.count("EXECUTE IMMEDIATE 'SELECT 1'") == (1 if legacy_prepared_statements is None else 0)
 
 
+@pytest.mark.skipif(
+    trino_version() <= '464',
+    reason="spooled protocol was introduced in version 464"
```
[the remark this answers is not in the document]
wendigo: Yeah, let's go with 466

tests/development_server.py
```diff
@@ -0,0 +1,118 @@
...
+        print(f"Error occurred: {e}")
+
+
+@contextmanager
+def start_development_server(port=None, trino_version=TRINO_VERSION):
+    minio = None
+    trino = None
+
+    try:
+        if TRINO_VERSION >= "466":
```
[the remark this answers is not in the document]
mosabua: I guess we could go with a integer number comparison ... since we will just keep counting up the Trino release number.. hahah

etc/spooling-manager.properties
```diff
@@ -0,0 +1,13 @@
+spooling-manager.name=filesystem
+fs.s3.enabled=true
+fs.location=s3://spooling/
+s3.endpoint=http://172.17.0.1:9000/
```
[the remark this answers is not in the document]
mdesmet: this url is also used on the clientside, this must be a url that is both accessible within the docker network as outside.

tests/development_server.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def create_bucket():
+    client = Minio(
+        "localhost:9000",
+        access_key=MINIO_ROOT_USER,
+        secret_key=MINIO_ROOT_PASSWORD,
+        secure=False
+    )
+    bucket_name = "spooling"
```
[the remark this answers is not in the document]
mdesmet: The bucket name is explicitly referenced in spooling-manager properties. Why we need to randomize it?

etc/spooling-manager.properties
```diff
@@ -0,0 +1,13 @@
...
+s3.region=us-east-1
+s3.aws-access-key=minio-access-key
+s3.aws-secret-key=minio-secret-key
+s3.path-style-access=true
+fs.segment.ttl=5m
+fs.segment.pruning.interval=15s
+fs.segment.pruning.batch-size=250
+# Disable as we don't support SSE-C while writing/reading from S3
```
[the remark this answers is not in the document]
mdesmet: I think we can handle this with a unit test as well. WDYT?

tests/development_server.py
```diff
@@ -0,0 +1,118 @@
...
+            .with_env("TRINO_CONFIG_DIR", "/etc/trino") \
+            .with_bind_ports(DEFAULT_PORT, port)
+
+        root = Path(__file__).parent.parent
+
+        trino = trino \
+            .with_volume_mapping(str(root / "etc/catalog"), "/etc/trino/catalog")
+
+        # Enable spooling config
+        if TRINO_VERSION >= "466":
```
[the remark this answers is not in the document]
mdesmet: The separate if blocks enable us to specify the common trino config only once. I have extracted the if block condition to a variable and reuse it.

tests/development_server.py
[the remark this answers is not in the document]
mdesmet: This bug is also present in the existing tests, will add a preparatory commit fixing it there as well.

etc/spooling-manager.properties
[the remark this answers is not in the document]
wendigo: can we use localhost?

trino/client.py
```diff
@@ -925,3 +977,217 @@ def _parse_retry_after_header(retry_after):
...
+    """
+    A subclass of Segment that handles inline data segments. The data is base64 encoded and
+    requires mapping to rows using the provided row_mapper.
+
+    Attributes:
+        rows (property): The rows of data in the segment, decoded and mapped from the base64 encoded data.
+    """
+    def __init__(self, segment: _InlineSegmentTO, row_mapper: RowMapper) -> None:
+        super().__init__(segment)
+        self._segment = cast(_InlineSegmentTO, segment)
```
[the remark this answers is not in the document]
mdesmet: the segment is passed through the subclass where it becomes `Segment`, but agreed it seems a bit weird.

trino/client.py
mdesmet: The row mapper actually doesn't take a string value but an array of values (each element of the array being a column of the row).

the row mapper only comes into play after decoding the json.

trino/client.py
mdesmet: See above.

trino/client.py
mdesmet: See above.

trino/client.py
mdesmet: See above.

README.md
```diff
@@ -469,6 +469,19 @@ conn = connect(
 )
 ```
 
+## Spooled protocol
+
+Enable the spooled protocol by specifying a supported encoding in the `encoding` parameter (equires a
+Trino server that supports the spooled protocol).
```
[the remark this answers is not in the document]
wendigo: Spooling

README.md
```diff
@@ -469,6 +469,31 @@ conn = connect(
 )
 ```
 
+## Spooled protocol
+
+The client spooling protocol requires [a Trino server that supports the spooling protocol and
```
wendigo, line 474: just `spooling protocol`

wendigo, line 474: Trino server with `spooling protocol support`

trino/client.py
wendigo: It's not extensible in the future

wendigo:
@mdesmet this is a good direction, the only comment I have is that the spooled segment data is resolved eagerly, I think that data fetch should happen in the `SegmentDecoder` when the spooled segment is iterated over, wdyt? (the same applies for ack)

mdesmet:
> @mdesmet this is a good direction, the only comment I have is that the spooled segment data is resolved eagerly, I think that data fetch should happen in the `SegmentDecoder` when the spooled segment is iterated over, wdyt? (the same applies for ack)

Actually it is not eagerly fetched. with `@abstractproperty` it is only fetched at the moment the `.data` property is used (in the `SegmentDecoder` `decode` method.

Following section in `SegmentIterator` manages both the acknowledgment and loading the next segment. So it actually works as you specify above.

```
try:
    self._current_segment = segment = next(self._segments)
    self._rows = iter(self._decoder.decode(segment))
except StopIteration:
    self._finished = True
```

wendigo:
LGTM

etc/spooling-manager.properties
```diff
@@ -0,0 +1,9 @@
+spooling-manager.name=filesystem
+fs.s3.enabled=true
+fs.location=s3://spooling/
+s3.endpoint=http://localstack:4566/
+s3.region=us-east-1
+s3.aws-access-key=test
+s3.aws-secret-key=test
+fs.segment.encryption=true
```
wendigo, line 8: you can drop this, it's a default

etc/spooling-manager.properties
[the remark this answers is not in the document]
mdesmet: Migrated to localstack, it was easier than thought.

trino/client.py
mdesmet: Moved to separate encoders.

tests/development_server.py
```diff
@@ -0,0 +1,127 @@
...
+
+from trino.constants import DEFAULT_PORT
+
+MINIO_ROOT_USER = "minio-access-key"
+MINIO_ROOT_PASSWORD = "minio-secret-key"
+
+TRINO_VERSION = os.environ.get("TRINO_VERSION") or "latest"
+TRINO_HOST = "localhost"
+
+
```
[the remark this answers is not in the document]
mdesmet: We need two lines in between function declarations, so no way around that.

etc/spooling-manager.properties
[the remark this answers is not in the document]
mdesmet: Resolved now, we don't need to access it outside the docker network as we use coordinator proxy.

mdesmet:
Only remaining issue is making the acknowledge executed in a background thread.

wendigo:
@mdesmet please rebase, how is the async ack going?

trino/client.py
```diff
@@ -925,3 +996,284 @@ def _parse_retry_after_header(retry_after):
...
+        def acknowledge_request():
+            try:
+                http_response = self._send_spooling_request(self.ack_uri, timeout=2)
+                if not http_response.ok:
+                    self._request.raise_response_error(http_response)
+            except Exception as e:
+                logger.error(f"Failed to acknowledge spooling request for segment {self}: {e}")
+        # Start the acknowledgment in a background thread
+        thread = threading.Thread(target=acknowledge_request, daemon=True)
+        thread.start()
```
mdesmet, line 1121: @wendigo : apparently I didn't push my latest changes. Here is the async impl.

wendigo:
@mdesmet you can run tests on 467 and drop the last commit

mdesmet:
> @mdesmet you can run tests on 467 and drop the last commit

done

trino/client.py
wendigo: I've pushed small fix as the threading will be expensive to use. PTAL

wendigo: https://github.com/trinodb/trino-python-client/pull/508/commits (last commit)

hashhar:
superseded by https://github.com/trinodb/trino-python-client/pull/509

PR #514, opened 2025-01-13
Make this package PEP-561 compatible by adding a py.typed marker

### LINKED ISSUE #482: Make this package PEP-561 compatible by adding a py.typed marker
### Describe the feature

I see that static type checking was discussed in #69 and mypy was enabled in #91. Do people think it's a good idea to add a [py.typed marker](https://peps.python.org/pep-0561/#specification) so that packages depending on `trino` can type check themselves against it? 

Right now, there's no way to force mypy to type check against this site-packages which lack this marker ([ref](https://github.com/python/mypy/issues/8545)).

_No response_

Priyansh121096:
## Description
Fixes #482 so that dependent packages can type check themselves against `trino`.

## Release notes

```markdown
* Make this package PEP-561 compatible ({issue}`482`)
```


hashhar:
how can I verify this works as expected? Is it as simple as importing trino module and seeing if `mypy` for example can do type checking for a script which calls methods from trino module?

hashhar:
I checked and it wasn't obvious to me that mypy doesn't check types from libraries unless they have py.typed marker.

IDEs for example use type information always (even when using mypy as the backend) so I didn't notice.

Thanks for your change. Merging this.

Priyansh121096:
Thanks @hashhar !

PR #525, opened 2025-01-31
Introduce DecodableSegment and remove Tuple in iterators

wendigo:
This makes the API much easier to consume in the `segments` cursor style:

```
cur = conn.cursor('segment')
cur.execute(sql)
segments = cur.fetchall()
total_row_count = 0
row_mapper = RowMapperFactory().create(columns=cur._query.columns, legacy_primitive_types=False)
for segment in segments:
  rows = list(SegmentIterator(segment, row_mapper))
  print ("rows length is " + str(len(rows)) + " " + segment.encoding)
  total_row_count += len(rows)
print(total_row_count)
```

This will work the same:

```
segments = cur.fetchall()
total_row_count = 0
row_mapper = RowMapperFactory().create(columns=cur._query.columns, legacy_primitive_types=False)

rows = list(SegmentIterator(segments, row_mapper))
print ("rows length is " + str(len(rows)))
total_row_count += len(rows)

print(total_row_count)
```

### DIFF
--- a/tests/integration/test_dbapi_integration.py
+++ b/tests/integration/test_dbapi_integration.py
@@ -29,12 +29,14 @@ from tzlocal import get_localzone_name  # type: ignore
 import trino
 from tests.integration.conftest import trino_version
 from trino import constants
+from trino.client import SegmentIterator
 from trino.dbapi import Cursor
 from trino.dbapi import DescribeOutput
 from trino.dbapi import TimeBoundLRUCache
 from trino.exceptions import NotSupportedError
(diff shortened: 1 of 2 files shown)


wendigo:
@mdesmet I've dropped the Tuple and the API feels more natural now

wendigo:
@hashhar ptal, alas it's a breaking change but it was broken API from the beginning and since it was in a single release so far, I think we should just fix it and move along

hashhar (approved):

wendigo:
@hashhar can we release new version?

PR #530, opened 2025-02-05
Allow authentication over HTTP

koszti:
## Description

The Trino Python client currently rejects all authentication attempts over HTTP. While this seems like is a sensible security measure on the part of the Trino client, using an HTTP URL to access the Trino server does not inherently indicate an insecure environment.

This could be useful for testing purposes or in modern service meshes, such as those using SPIFFE authentication where injection of the spiffe headers and adding TLS are handled by spire and envoy separately in sidecars ([doc here](https://spiffe.io/docs/latest/microservices/envoy/)). Furthermore the trino server also has the option of `http-server.authentication.allow-insecure-over-http=true` ([doc here](https://trino.io/docs/current/security/tls.html#configure-the-coordinator)) so it might be reasonable to not rejecting non-https authentication attempts in the client side.

This pull request includes the following improvements:
* The SQLAlchemy dialect no longer automatically sets `http_scheme` to `https` when authentication is detected in the connection URL.
* Authentication over HTTP is now allowed.
* To authenticate over HTTP using SQLAlchemy, the `http_scheme="http"` connection argument is now respected. If `http_scheme` is not specified, the scheme defaults to `https` when the port is 443; otherwise, it defaults to `http`.

Example auth over HTTP:
```
- DBAPI

    ```python
    from trino.auth import BasicAuthentication

    conn = connect(
        user="<username>",
        auth=BasicAuthentication("<username>", "<password>"),
        http_scheme="http",
        ...
    )
    ```

- SQLAlchemy

    ```python

    # defaults to use http if the port is not 443
    engine = create_engine("trino://<username>:<password>@<host>:<port>/<catalog>")

    # or as connect_args
    from trino.auth import BasicAuthentication
    engine = create_engine(
        "trino://<username>@<host>:<port>/<catalog>",
        connect_args={
            "auth": BasicAuthentication("<username>", "<password>"),
            "http_scheme": "http",
        }
    )
    ```
```

## Non-technical explanation

This PR adds functionality to optionally authenticate over HTTP.

## Release notes

```markdown
* Allow authentication over HTTP ({issue}`530`)
```

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -489,8 +489,6 @@ class TrinoRequest:
         self._exceptions = self.HTTP_EXCEPTIONS
         self._auth = auth
         if self._auth:
-            if self._http_scheme == constants.HTTP:
-                raise ValueError("cannot use authentication with HTTP")
             self._auth.set_http_session(self._http_session)
             self._exceptions += self._auth.get_exceptions()
 
--- a/tests/unit/sqlalchemy/test_dialect.py
+++ b/tests/unit/sqlalchemy/test_dialect.py
@@ -58,7 +58,6 @@ class TestTrinoDialect:
                     catalog="system",
                     user="user",
                     auth=BasicAuthentication("user", "pass"),
-                    http_scheme="https",
                     source="trino-rulez"
                 ),
             ),
@@ -80,7 +79,6 @@ class TestTrinoDialect:
                     catalog="system",
                     user="user",
                     auth=CertificateAuthentication("/my/path/to/cert", "afdlsdfk%4#'"),
-                    http_scheme="https",
                     source="trino-sqlalchemy"
                 ),
             ),
@@ -100,7 +98,6 @@ class TestTrinoDialect:
                     catalog="system",
                     user="user",
                     auth=JWTAuthentication("afdlsdfk%4#'"),
-                    http_scheme="https",
                     source="trino-sqlalchemy"
                 ),
             ),
@@ -168,7 +165,6 @@ class TestTrinoDialect:
                     catalog="system",
                     user="user@test.org/my_role",
                     auth=BasicAuthentication("user@test.org/my_role", "pass /*&"),
-                    http_scheme="https",
                     source="trino-sqlalchemy",
                     session_properties={"query_max_run_time": "1d"},
                     http_headers={"trino": 1},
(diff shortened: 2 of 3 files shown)


trino/client.py
```diff
@@ -462,6 +463,7 @@ def __init__(
         http_session: Optional[Session] = None,
         http_scheme: Optional[str] = None,
         auth: Optional[Authentication] = constants.DEFAULT_AUTH,
+        allow_insecure_auth: bool = False,
```
wendigo, line 466: nit: just insecure

hashhar:
I think we should simply remove the check.

After all the server does reject auth over HTTP unless:

- insecure over HTTP is allowed via config
- OR the connection was forwarded from a LB that terminates TLS (signalled via X-Forwarded-Proto header in the request)

What does the JDBC driver do right now?

trino/client.py
koszti: done

README.md
```diff
@@ -365,6 +365,41 @@ It follows the interface for `KerberosAuthentication`, but is using
     )
     ```
 
+
+### Allow Insecure Authentication
```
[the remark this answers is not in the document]
koszti: done

koszti:
> I think we should simply remove the check.

@hashhar , this needs to be tested, but removing it without introducing a new flag could lead to breaking changes. The SQLAlchemy dialect automatically sets http_scheme="https" for all authentication types (([here](https://github.com/trinodb/trino-python-client/blob/1e95bbfe3cf5b28e7329c21463970f3fc9c8e44d/trino/sqlalchemy/dialect.py#L140))), so we’d also need to remove that along with the actual check (([here](https://github.com/trinodb/trino-python-client/blob/1e95bbfe3cf5b28e7329c21463970f3fc9c8e44d/trino/client.py#L492))).

Would this mean that clients not explicitly specifying http_scheme would end up using an unknown protocol?

hashhar:
I'm saying to only remove these two lines https://github.com/trinodb/trino-python-client/blob/1e95bbfe3cf5b28e7329c21463970f3fc9c8e44d/trino/client.py#L492-L493

koszti:
removing those two lines only doesn’t help unfortunately becauae the `http_sceme` is set to `https` automatically for all auth types. Client actually has no chance to overwrite it by passing it in `connect_args` and they end up with https only whatever they do.

hashhar:
I think the SQLAlchemy dialect code is being too smart for no reason. It should not be automatically setting http_scheme in any case. Same as how the db-api doesn't do it.

hashhar (approved):
LGTM. @damian3031 please take a look too.

PR #531, opened 2025-02-05
Authenticate against external identity provider, and send grant to Trino

RoseGoldIsntGay:
## Description

The new `KeycloakAuthentication` auth flow allows usage of an external OIDC identity provider to authenticate against a Trino cluster secured using OAuth2.

## Non-technical explanation

This authentication type uses credentials for an OpenId Connect provider, such as Keycloak, retrieves a grant token and uses it to authenticate against Trino.

## Release notes

### Keycloak Authentication

The `KeycloakAuthentication` class can be used to connect to a Trino cluster that is configured with the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.html) using an external OIDC identity provider (i.e Keycloak)

It works by sending credentials to the OpenId identity provider and recieving a grant, then passing said grant to the Trino cluster secured using OAuth2

> [!WARNING]
> Client Authentication must be turned off (public access) as the flow does not send a client secret

- DBAPI

    ```python
    from trino.auth import KeycloakAuthentication

    conn = connect(
        user="<username>",
        auth=KeycloakAuthentication(
            username="<username>",
            password="<password>",
            keycloak_url="<keycloak_url>",
            realm="<realm>",
            client_id="<client_id>",
        ),
        ...
    )

    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,35 @@ The OAuth2 token will be cached either per `trino.auth.OAuth2Authentication` ins
     )
     ```
 
+### Keycloak Authentication
+
+The `KeycloakAuthentication` class can be used to connect to a Trino cluster that is configured with the [OAuth2 authentication type](https://trino.io/docs/current/security/oauth2.html) using an external OIDC identity provider (i.e Keycloak)
+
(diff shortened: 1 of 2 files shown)


hashhar:
This seems too custom to your specific implementation of Keycloak.

In general you can use Oauth authentication class to integrate with keycloak already.

PR #541, opened 2025-03-18
Ignore primary key, foreign key and unique constraints for DDL generation

willmostly:
## Description

Define custom visitors in `TrinoDDLCompiler` to ignore constraints.

## Non-technical explanation

By default, SQLAlchemy adds `PRIMARY KEY` and other constraint keywords to table creation DDL when these constraints are defined on a `Table` or class derived from declarative base. Since Trino does not support constraints, this results in invalid SQL. With this change the trino sqlalchemy extension now ignores primary key, foreign key and unique constraints when generating table DDL.

## Release notes

```markdown
* ignore constraints for table DDL generation
```

### DIFF
--- a/trino/sqlalchemy/compiler.py
+++ b/trino/sqlalchemy/compiler.py
@@ -14,6 +14,7 @@ from sqlalchemy.sql import compiler
 from sqlalchemy.sql import sqltypes
 from sqlalchemy.sql.base import DialectKWArgs
 from sqlalchemy.sql.functions import GenericFunction
+from sqlalchemy.util import warn as SAWarn
 
 # https://trino.io/docs/current/language/reserved.html
 RESERVED_WORDS = {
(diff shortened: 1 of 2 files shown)


trino/sqlalchemy/compiler.py
```diff
@@ -181,7 +181,14 @@ def visit_try_cast(self, element, **kw):
 
 
 class TrinoDDLCompiler(compiler.DDLCompiler):
-    pass
+    def visit_foreign_key_constraint(self, constraint, **kw):
+        return None
```
damian3031, line 185: Rather than silently ignoring constraints, maybe we should raise an explicit error indicating that these constraints are not supported in Trino:
```suggestion
from sqlalchemy.exc import CompileError
...
    def visit_foreign_key_constraint(self, constraint, **kw):
        raise CompileError("Trino does not support FOREIGN KEY constraints.")
```
___
Or at least we could add a warning to make it clearer that the constraint is being ignored.
```suggestion
import warnings
from sqlalchemy.exc import SAWarning
...
    def visit_foreign_key_constraint(self, constraint, **kw):
        warnings.warn("Trino does not support FOREIGN KEY constraints. Constraint will be ignored.", SAWarning)
        return None
```

@willmostly @hashhar WDYT?

hashhar: @willmostly PTAL.

muka:
this would enable using sqlalchemy ORM-like features, being aware that PK and FK are not enforced. Right now it's not otherwise possible.

```
sqlalchemy.exc.ProgrammingError: (trino.exceptions.TrinoUserError) TrinoUserError(type=USER_ERROR, name=TYPE_NOT_FOUND, message="line 12:2: Unknown type 'KEY(id)' for column 'PRIMARY'", query_id=20250812_125239_00172_2zeem)
```

hashhar (approved):

Issue #545, opened 2025-04-08
Session Authorization Error


alpaca-attila:
### Expected behavior

The command returns no error:
```
_ = cur.execute("SET SESSION AUTHORIZATION 'username'")
```

### Actual behavior

Related to this PR https://github.com/trinodb/trino-python-client/pull/349
We bumped into some error with the following code
```
TRINO_HOST='***'
TRINO_USER='***'
TRINO_PASS='***'
TRINO_PORT=443

import logging
from trino.dbapi import connect
from trino.auth import BasicAuthentication

logging.basicConfig(level=logging.ERROR) 

conn = connect(
    host=TRINO_HOST,
    port=TRINO_PORT,
    user=TRINO_USER,
    auth=BasicAuthentication(TRINO_USER, TRINO_PASS),
    http_scheme="https",
)
conn._http_session.verify = False
cur = conn.cursor()
_ = cur.execute("SET SESSION AUTHORIZATION 'username'")
cur.close()
conn.close()
```
Producing 

```
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): trino.production.lakehouse.tradetalk.us:443
DEBUG:urllib3.connectionpool:https://trino.production.lakehouse.tradetalk.us:443 "POST /v1/statement HTTP/1.1" 200 370
DEBUG:urllib3.connectionpool:[https://trino.production.lakehouse.tradetalk.us:443](https://trino.production.lakehouse.tradetalk.us/) "GET /v1/statement/queued/20250408_205705_36295_5qmnr/y5a9c6452eeb4228649e5bbb1d3008098a8120cdb/1 HTTP/1.1" 200 369
DEBUG:urllib3.connectionpool:https://trino.production.lakehouse.tradetalk.us:443 "GET /v1/statement/queued/20250408_205705_36295_5qmnr/ye59beed6bc2306af1ebc7960d503377c74a095fc/2 HTTP/1.1" 200 819
---------------------------------------------------------------------------
TrinoUserError                            Traceback (most recent call last)
Cell In[195], line 1
----> 1 a = trino_run()

Cell In[193], line 38, in trino_run()
     36 conn._http_session.verify = False
     37 cur = conn.cursor()
---> 38 data = cur.execute("SET SESSION AUTHORIZATION 'dragonfly'")
     39 cur.close()
     40 conn.close()

[...]

PR #349, opened 2023-03-18
Support SET SESSION AUTHORIZATION on trino-python-client

baohe-zhang:
## Description
This PR is to support `SET SESSION AUTHORIZATION user` and `RESET SESSION AUTHORIZATION` sql commands on the python client. Those 2 sql commands are supported on the server side from this PR https://github.com/trinodb/trino/pull/16067

Related to issue https://github.com/trinodb/trino/issues/2512

Tested locally with the script:
```python
from trino.dbapi import connect

conn = connect(
    host="localhost",
    port=8080,
    user="baozhang",
    catalog="tpch"
)
cur = conn.cursor()

cur.execute("SELECT COUNT(*) FROM tpch.sf1.orders")
rows = cur.fetchall()
assert rows[0][0] == 1500000

cur.execute("SELECT CURRENT_USER")
rows = cur.fetchall()
assert rows[0][0] == 'baozhang'

cur.execute("SET SESSION AUTHORIZATION 'prestodv'")
cur.execute("SELECT CURRENT_USER")
rows = cur.fetchall()
assert rows[0][0] == 'prestodv'

cur.execute("RESET SESSION AUTHORIZATION")
cur.execute("SELECT CURRENT_USER")
rows = cur.fetchall()
assert rows[0][0] == 'baozhang'
```

## Release notes

( ) This is not user-visible or docs only and no release notes are required.
(x) Release notes are required, please propose a release note for me.
( ) Release notes are required, with the following suggested text:

PR #551, opened 2025-04-14
add SESSION_AUTHORIZATION to client capabilities

kempspo:
## Description
PR seeks to resolve issue https://github.com/trinodb/trino-python-client/issues/545.

[Trino expects the Client Capabilities header](https://github.com/trinodb/trino/blob/71390927f48d0aaf1372476e761fdf4a427ccb96/core/trino-main/src/main/java/io/trino/execution/SetSessionAuthorizationTask.java#L67-L69) `SESSION_AUTHORIZATION` exists so that it allows the `SET SESSION AUTHORIZATION`.

[This wasn't added in the initial PR](https://github.com/trinodb/trino-python-client/pull/349)

## Release notes

```markdown
* Fix `SET SESSION AUTHORIZATION` not working. ({issue}`545`)
```

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -135,7 +135,7 @@ def test_request_headers(mock_get_and_post):
         assert headers[constants.HEADER_SESSION] == ""
         assert headers[constants.HEADER_TRANSACTION] is None
         assert headers[constants.HEADER_TIMEZONE] == timezone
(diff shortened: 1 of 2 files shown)


hashhar (commented):
Thanks, I'll make some changes and merge - namely adding constants for the capabilities header value since it ideally should be exactly the same everywhere in the client.

kempspo:
Would be happy to do that @hashhar. 

Where do you think would be the best place to add that in? And are we okay with adding it to this PR?

wendigo (approved):

PR #565, opened 2025-06-10
Fix test_segments_cursor

damian3031:
## Description

`test_segments_cursor` is passing on Trino 475, but failing on Trino 476 with error:
```bash
AttributeError: 'InlineSegment' object has no attribute 'uri'
```

My understanding is that after this change https://github.com/trinodb/trino/pull/25940, some Segments are Inline, whereas before, all of them were Spooled.
Since InlineSegment doesn't contain `uri` and `ack_uri`, I added a check for the existence of these attributes before asserting their types.

## Release notes

```markdown
```


wendigo:
@damian3031 incorrect. Inline and spooled segments were allowed from the beginning. Linked issue has nothing to do with it

PR #568, opened 2025-06-19
Fix Update user-related HTTP headers to fix impersonation

izeigerman:
## Description

The latest version of Trino changed how the HTTP headers are handled: https://trino.io/docs/current/develop/client-protocol.html#client-request-headers when the server returns the `X-Trino-Set-Authorization-User` header. 

Thus, the `X-Trino-Authorization-User` header is no longer supported. Instead, `X-Trino-User` should be set to the authorization user, and X-Trino-Original-User should be set to the originally authenticated user.

Without these changes, the impersonation doesn't work with the latest versions of trino.

## Release notes

```markdown
* Fix impersonation support
```


martint:
@cla-bot check

izeigerman:
Hey @martint @raunaqmorarka @hashhar @wendigo is there anything else preventing this PR from being merged?

PR #576, opened 2025-08-01
Migrate from built-in json to faster orjson

wendigo:
It is way faster than the current json parser: https://github.com/ijl/orjson?tab=readme-ov-file#performance

## Release notes

```markdown
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,6 +84,7 @@ setup(
     python_requires=">=3.9",
     install_requires=[
         "lz4",
+        "orjson >= 3.11.0 ; platform_python_implementation != 'PyPy'",
         "python-dateutil",
         "pytz",
         # requests CVE https://github.com/advisories/GHSA-j8r2-6x86-q33q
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -26,6 +27,10 @@
 with open(os.path.join(here, "README.md"), "r", "utf-8") as f:
     readme = f.read()
 
+orjson_require = []
+if platform.python_implementation() != 'PyPy':
+    orjson_require = ['orjson >= 3.11.0']
+
```
damian3031, line 33:
```suggestion

```

Instead of having logic here and adding it to `install_requires` later, you can use [Environment Marker](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers)
```python
install_requires=[
    ...
    "zstandard",
    "orjson >= 3.11.0 ; platform_python_implementation != 'PyPy'"
],
 ```

hashhar (approved):

PR #580, opened 2025-10-31
Handle secondary TLS port 8443 as https by default

Spikhalskiy:
## Description

Before 0.334.0, both `login@password:host:443` and `login@password:host:8443` URLs were handled as HTTPS because the connection string contained a password, and `https` was enforced by Trino in such situations.

https://github.com/trinodb/trino-python-client/pull/530 removed this enforcement. And all connection strings without an explicitly specified additional schema parameter were treated as `http`.

https://github.com/trinodb/trino-python-client/pull/564 added a fix for the 443 port. And 443 port only.
 
Meanwhile, 8443 is a secondary HTTPS/TLS port, just as 8080 is a secondary port for port 80. Even trino TLS documentation uses 8443 by default: https://trino.io/docs/current/security/tls.html

This PR adds handling of 8443 port as HTTPS by default.

## Non-technical explanation

Connection strings with 8443 port should be handled as HTTPS by default.

## Release notes

( x ) Release notes are required, with the following suggested text: 

```markdown
* 8443 port is handled as HTTPS by default (#580)
```

### DIFF
--- a/tests/unit/test_dbapi.py
+++ b/tests/unit/test_dbapi.py
@@ -325,6 +325,7 @@ def test_description_is_none_when_cursor_is_not_executed():
         ("http://mytrinoserver.domain:9999", None, None, constants.HTTP),
         # Infer from port
         ("mytrinoserver.domain", constants.DEFAULT_TLS_PORT, None, constants.HTTPS),
+        ("mytrinoserver.domain", constants.SECONDARY_TLS_PORT, None, constants.HTTPS),
         ("mytrinoserver.domain", constants.DEFAULT_PORT, None, constants.HTTP),
         # http_scheme parameter has higher precedence than port parameter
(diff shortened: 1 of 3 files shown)


hashhar (commented):
The Trino docs do use 8443 as the default TLS port so I understand the motivation. Hardcoding specific ports is fragile though and there is no limit to ports. The better approach would be to detect the presence of
credentials/auth and infer HTTPS from that (as the client did before #530). That would cover 8443 and any other port without needing special cases. Note that Trino can accept credentials
over HTTP too, so even credential detection isn't perfect, but it's a much better heuristic than port numbers.

As-is, this change risks silently breaking users running HTTP on 8443 who don't pass http_scheme explicitly.

hashhar:
Also current behaviour is intentionally aligned to JDBC driver so that all clients behave similarly.

PR #588, opened 2026-01-12
added heartbeat interval to check if query is still fetching data

### LINKED ISSUE #554: Send query heartbeat while downloading results
### Describe the feature

https://github.com/trinodb/trino/pull/25267

_No response_

nadavtzaysler:
## Description
fixed https://github.com/trinodb/trino-python-client/issues/554
made the python client send heartbeats to the server at fixed configured intervals

## Non-technical explanation
queries will now be able to continue fetching data after 15 min

## Release notes

(X ) Release notes are required, please propose a release note for me.

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -660,6 +660,15 @@ class TrinoRequest:
     def delete(self, url: str) -> Response:
         return self._delete(url, timeout=self._request_timeout, proxies=PROXIES)
 
+    def send_heartbeat(self, url: str) -> Response:
+        """Send HEAD request for query heartbeat."""
+        return self._http_session.head(
+            url,
+            headers=self.http_headers,
+            timeout=self._request_timeout,
+            proxies=PROXIES,
+        )
+
     @staticmethod
     def _process_error(error, query_id: Optional[str]) -> Union[TrinoExternalError, TrinoQueryError, TrinoUserError]:
         error_type = error["errorType"]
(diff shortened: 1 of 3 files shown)


RoeyoOgen (changes requested):

trino/client.py
```diff
@@ -981,6 +1009,66 @@ def cancel(self) -> None:
...
+
+                except Exception:
+                    # On any exception, increment failure count
+                    self._heartbeat_failures += 1
+                    if self._heartbeat_failures >= 3:
+                        self._heartbeat_enabled = False
+                        break
+
+                # Sleep for the heartbeat interval
+                sleep(self._heartbeat_interval_seconds)
```
azawlocki-sbdt, line 1058: The heartbeat thread sleeps here for a relatively long time (the default hearbeat interval in the Java client in 30s) without any possibility for interruption. After the query completes or is cancelled, `_stop_heartbeat()` blocks until the heartbeat thread wakes up, checks the loop condition and exits, which may take for example 30 seconds.

The same problem occurs when the error count from heartbeat requests reaches the threshold. 

A better approach IMO would be to have a `threading.Event`, set it when the query completes, and do `self._heartbeat_cancelled_event.wait(timeout=self._heartbeat_interval)` in the loop instead of `sleep()`.

azawlocki-sbdt: What I said above is not right, `self._stop_heartbeat()` currently waits at most 1s for the thread and may leave the heartbeat thread lingering if it does not exit by then. Better than blocking for up to 30s, but still something that should be avoided.

azawlocki-sbdt:
@nadavtzaysler, #554 is already closed by another PR

PR #593, opened 2026-03-06
Raise error when sending credentials over HTTP

hashhar:
## Description

Raises a `TrinoAuthError` when authentication is configured while using an insecure HTTP connection. This aligns with the Java client behavior where TLS/SSL is strictly required for authentication — the Java client raises a `RuntimeException` with the message \"TLS/SSL is required for authentication\" in the same scenario.

The error message follows the same phrasing as the Java client and includes Python-specific guidance on how to switch to HTTPS.

## Non-technical explanation

Connecting with authentication over plain HTTP is now an error. Use `https://` in the host URL or pass `http_scheme='https'`.

## Release notes

```markdown
* Raise `TrinoAuthError` when authentication credentials are used over an insecure HTTP connection. ({issue}`593`)
```

### DIFF
--- a/tests/unit/test_dbapi.py
+++ b/tests/unit/test_dbapi.py
@@ -18,6 +18,7 @@ import pytest
 from httpretty import httprettified
 from requests import Session
 
+import trino.exceptions
 from tests.unit.oauth_test_utils import _get_token_requests
 from tests.unit.oauth_test_utils import _post_statement_requests
 from tests.unit.oauth_test_utils import GetTokenCallback
@@ -27,6 +28,7 @@ from tests.unit.oauth_test_utils import RedirectHandler
 from tests.unit.oauth_test_utils import SERVER_ADDRESS
 from tests.unit.oauth_test_utils import TOKEN_RESOURCE
 from trino import constants
+from trino.auth import BasicAuthentication
 from trino.auth import OAuth2Authentication
 from trino.dbapi import connect
 from trino.dbapi import Connection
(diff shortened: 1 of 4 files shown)


findepi:
> Adds warning when sending credentials over HTTP. JDBC actually disallows this entirely but it can be a breaking change for Python client now.

Can we imagine a future where Python client does unsafe thing by default in 10 years from now?
I suppose this is not what we want. This means we need to embrace the breakage.

Do you plan to convert warning into failure in a 1-2 release from now?
If not, let's make it failure today.

hashhar:
I plan to convert to failure. Just want to see if there are known applications or setups which explode and provide a version they can pin to while they sort out insecure configs.

hashhar:
@findepi made into an error, was near useless in current shape - i tested with some tools locally

tests/unit/test_dbapi.py
```diff
@@ -362,3 +364,12 @@ def test_default_encoding_zstd():
 def test_default_encoding_all():
     connection = Connection("host", 8080, user="test")
     assert connection._client_session.encoding == ["json+zstd", "json+lz4", "json"]
+
+
+def test_error_when_auth_over_http():
+    with pytest.raises(trino.exceptions.TrinoAuthError, match="TLS/SSL is required for authentication"):
+        Connection("mytrinoserver.domain", auth=BasicAuthentication("u", "p"))
```
damian3031, line 371:
```suggestion
        Connection("mytrinoserver.domain", http_scheme=constants.HTTP, auth=BasicAuthentication("u", "p"))
```
nit: we could pass `http_scheme=constants.HTTP` to make these two tests symmetric and explicit, rather than implicitly relying on HTTP being the default

damian3031 (approved):

PR #594, opened 2026-03-06
Add automated Python version update workflow

hashhar:
## Description

Add a scheduled GitHub Action that keeps the project's Python version support matrix and metadata up to date without manual PRs.

## Release notes

```markdown
```

### DIFF
--- a/.github/scripts/update_python_versions.py
+++ b/.github/scripts/update_python_versions.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python3
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Fetches the official Python release cycle and updates project files to reflect
+the current set of supported Python versions.
+
+Run from the project root:
+    python .github/scripts/update_python_versions.py
+
+Exit code 0 on success (whether or not files were changed).
+The caller can inspect `git diff --quiet` to decide whether to open a PR.
+"""
+import json
+import re
+import sys
+import urllib.request
+from typing import Any
+from typing import Callable
+
+RELEASE_CYCLE_URL = (
+    "https://raw.githubusercontent.com/python/devguide/main/include/release-cycle.json"
+)
+
+
+def _require_sub(
+    pattern: str, repl: str | Callable[[re.Match[str]], str], string: str, label: str, **kwargs: Any
+) -> str:
+    """Like re.sub(), but warns on stderr when the pattern does not match."""
+    result, count = re.subn(pattern, repl, string, **kwargs)
+    if count == 0:
+        print(f"WARNING: no match for {label} pattern: {pattern}", file=sys.stderr)
+    return result
+
+
+def fetch_release_cycle() -> dict[str, Any]:
+    with urllib.request.urlopen(RELEASE_CYCLE_URL, timeout=30) as response:
+        return json.loads(response.read())
+
+
+def version_key(version: str) -> tuple[int, ...]:
+    return tuple(int(x) for x in version.split("."))
+
+
+def compute_version_sets(cycle: dict[str, Any]) -> tuple[set[str], set[str]]:
+    """Return (eol, active) sets of '3.X' version strings."""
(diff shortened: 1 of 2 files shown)


.github/scripts/update_python_versions.py
```diff
@@ -0,0 +1,238 @@
...
+    # Include entries use `python: "X.Y"` format (distinct from the matrix list).
+    # Find whatever version they currently pin (the highest one = last "latest") and bump it.
+    current_include_version = re.findall(r'python: "(\d+\.\d+)"', content)
+    if current_include_version:
+        old_include_version = max(current_include_version, key=version_key)
+        if old_include_version != latest_stable:
+            content = content.replace(
+                f'python: "{old_include_version}"',
+                f'python: "{latest_stable}"',
+            )
```
damian3031:
```suggestion
    current_include_version = re.findall(r'- \{ python: "(\d+\.\d+)"', content)
    if current_include_version:
        old_include_version = max(current_include_version, key=version_key)
        if old_include_version != latest_stable:
            content = content.replace(
                f'- {{ python: "{old_include_version}"',
                f'- {{ python: "{latest_stable}"',
            )
```
nit: `str.replace` on `python: "X.Y"` isn't scoped to the include block - it could match a comment or future job definition. Maybe we can anchor to the include entry format `- { python:` instead.

.github/scripts/update_python_versions.py
```diff
@@ -0,0 +1,238 @@
...
+)
+
+
+def _require_sub(
+    pattern: str, repl: str | Callable[[re.Match[str]], str], string: str, label: str, **kwargs: Any
+) -> str:
+    """Like re.sub(), but warns on stderr when the pattern does not match."""
+    result, count = re.subn(pattern, repl, string, **kwargs)
+    if count == 0:
+        print(f"WARNING: no match for {label} pattern: {pattern}", file=sys.stderr)
```
damian3031: Shouldn't we raise an error instead of warning? If some files get updated but others don't (e.g., due to a file format change), it would be better to fail the entire workflow so someone can investigate, rather than risk merging a PR that's only partially updated

.github/workflows/update-python-versions.yml
```diff
@@ -0,0 +1,84 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-python-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
damian3031:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/update-python-versions.yml
```diff
@@ -0,0 +1,84 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: actions/setup-python@v5
```
damian3031:
```suggestion
      - uses: actions/setup-python@v6
```

.github/workflows/update-python-versions.yml
```diff
@@ -0,0 +1,84 @@
...
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-python-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
damian3031: is that needed?

.github/workflows/update-python-versions.yml
```diff
@@ -0,0 +1,84 @@
+name: Update Python versions
+
+on:
+  schedule:
+    - cron: '0 0 1 * *'  # first of each month
```
damian3031, line 5: Since 2019, Python new versions are released in October. I wonder if we scheduled this workflow just for the first of November?

.github/workflows/update-python-versions.yml
```diff
@@ -0,0 +1,84 @@
...
+          Files updated:
+          - `setup.py`: classifiers and `python_requires`
+          - `tox.ini`: `envlist`
+          - `.github/workflows/ci.yml`: build matrix and Python pin
+          - `.github/workflows/release.yml`: Python version pin
+          - `README.md`: Python version mention
+          - `.pre-commit-config.yaml`: `--pyXX-plus` argument
+
+          > This PR was automatically created by the [Update Python versions](${{ github.server_url }}/${{ github.repository }}/actions/workflows/update-python-versions.yml) workflow.
+          EOF
```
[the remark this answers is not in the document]
damian3031: I think this is a valid point

.github/scripts/update_python_versions.py
hashhar: thanks, great catch

hashhar: made slightly more robust (looking for appearances within `include:` only.

wendigo (approved):

PR #597, opened 2026-04-07
Lazily load spooled result set segments

wendigo:
When using the spooling protocol, TrinoQuery.fetch() materializes all segments into a single list, causing out-of-memory errors on large result sets. Return SegmentIterator directly so rows are decoded on demand, keeping memory usage constant regardless of result set size.

Supersedes https://github.com/trinodb/trino-python-client/pull/590

## Release notes

```markdown
```

### DIFF
--- a/tests/development_server.py
+++ b/tests/development_server.py
@@ -51,7 +51,7 @@ def start_development_server(port=None, trino_version=TRINO_VERSION):
         network = Network().create()
         supports_spooling_protocol = TRINO_VERSION == "latest" or int(TRINO_VERSION) >= 466
         if supports_spooling_protocol:
-            localstack = LocalStackContainer(image="localstack/localstack:latest", region_name="us-east-1") \
+            localstack = LocalStackContainer(image="localstack/localstack:4.14.0", region_name="us-east-1") \
                 .with_name("localstack") \
                 .with_network(network) \
                 .with_bind_ports(4566, 4566) \
--- a/trino/client.py
+++ b/trino/client.py
@@ -39,6 +39,7 @@ import atexit
 import base64
 import copy
 import functools
+import itertools
 import os
 import random
 import re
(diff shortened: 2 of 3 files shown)


trino/client.py
```diff
@@ -933,9 +904,22 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
...
+        # Rows start as an empty list. Early responses often contain only stats,
+        # so we keep fetching until actual data arrives.
+        #
+        # Two protocols produce rows differently:
+        #  - Direct: fetch() returns a list - accumulate into the existing list.
+        #  - Spooling: fetch() returns a lazy iterator - replace rows and stop,
+        #    because we cannot cheaply check iterator length.
+        while not self.finished and not self.cancelled and self._result.rows == []:
+            new_rows = self.fetch()
+            if isinstance(new_rows, list):
```
hashhar, line 918: do we also need this check in `columns`?

tests/development_server.py
```diff
@@ -51,7 +51,7 @@ def start_development_server(port=None, trino_version=TRINO_VERSION):
         network = Network().create()
         supports_spooling_protocol = TRINO_VERSION == "latest" or int(TRINO_VERSION) >= 466
         if supports_spooling_protocol:
-            localstack = LocalStackContainer(image="localstack/localstack:latest", region_name="us-east-1") \
+            localstack = LocalStackContainer(image="localstack/localstack:4.14.0", region_name="us-east-1") \
```
hashhar, line 54: thank you

trino/client.py
```diff
@@ -933,9 +904,22 @@ def execute(self, additional_http_headers=None) -> TrinoResult:
...
-        while not self.finished and not self.cancelled and len(self._result.rows) == 0:
-            self._result.rows += self.fetch()
+        # Block until rows are available, the query finishes, or it is canceled.
+        # Rows start as an empty list. Early responses often contain only stats,
+        # so we keep fetching until actual data arrives.
+        #
+        # Two protocols produce rows differently:
+        #  - Direct: fetch() returns a list - accumulate into the existing list.
+        #  - Spooling: fetch() returns a lazy iterator - replace rows and stop,
+        #    because we cannot cheaply check iterator length.
```
hashhar, line 915: True but we could peek at the element. Right now it seems it can return `None` on the first call but results would arrive later. Hard to consume for users.

Something like:

```python
else:
      try:
          first_row = next(new_rows)
          self._result.rows = itertools.chain([first_row], new_rows)
          break
      except StopIteration:
          self._result.rows = []
```

trino/client.py
hashhar, line 806: Filing an issue about a problem with using iterators here. no need to fix in this PR. Python iterators is example of bad API design. No ability to distinguish StopIteration causes.

hashhar (approved):
LGTM % comment

trino/client.py
wendigo: Why? I don't think we do

hashhar: https://github.com/trinodb/trino-python-client/pull/599

PR #603, opened 2026-04-30
Retry on HTTP 200 with empty response body

### LINKED ISSUE #596: TrinoRequest.process() crashes with JSONDecodeError on empty 200 response instead of retrying
### Expected behavior

When Trino returns an HTTP 200 response with an empty body (which can happen under load), the client should recognize this as a transient error and retry the request, similar to how it retries on 429, 502, 503, and 504 status codes. The built-in _retry_with mechanism should handle this case gracefully.

### Actual behavior

The process() method in client.py checks http_response.ok (which passes for status 200), then immediately calls json.loads(http_response.text) on the empty body, resulting in a JSONDecodeError: Expecting value: line 1 column 1 (char 0). This exception is not in the handled_exceptions list of _retry_with, so no retry is attempted.
This is non-deterministic and occurs when executing multiple queries in a loop (e.g., batched INSERTs via SQLAlchemy). The same query succeeds when run individually. The error originates at the response = json.loads(http_response.text) line in TrinoRequest.process().
A possible fix would be to either:

Check for an empty response body before parsing JSON in process() and raise a retryable exception
Add JSONDecodeError to the handled_exceptions in the retry logic
Add a condition guard in _retry_with that checks for empty response content on 200s

Using SQLAlchemy with trino.sqlalchemy dialect

### Steps To Reproduce

"""
Minimal reproduction: Trino Python client crashes with JSONDecodeError
when receiving HTTP 200 with empty body under load.

Run multiple INSERT...SELECT queries in a loop to trigger the issue.
The exact batch that fails is non-deterministic.
"""

from sqlalchemy import create_engine, text

engine = create_engine("trino://user@trino-host:8080/catalog/schema")

# Simulate batched inserts - adjust table/columns to your environment
issue_ids = list(range(1, 101))  # 100 IDs to process
batch_size = 2

n_batches = (len(issue_ids) + batch_size - 1) // batch_size

for i in range(n_batches):
    batch = issue_ids[i * batch_size : (i + 1) * batch_size]
    ids_csv = ",".join(str(x) for x in batch)

    # Any INSERT...SELECT with a JOIN across wide tables will do.
    # The key conditions to reproduce:
    #   1) Many sequential queries in a tight loop
    #   2) Large result sets or wide tables (many columns)
    #   3) No sleep between iterations
    query = f"""
    INSERT INTO target_table
    SELECT a.*, b.*
    FROM table_a a
    JOIN table_b b ON a.id = b.id
    WHERE a.id IN ({ids_csv})
    """

    session = engine.connect()
    result = session.execute(text(query)).fetchall()  # crashes here
    session.close()

    # No delay between batches — increases likelihood of empty 200 response

### Log output

_No response_

### Operating System

Linux (container based on conda, Python 3.12)

0.333.0

### Trino Server version

479

### Python version

3.12.11

emanuelsdsj:
## Description

fixed #596 

`TrinoRequest.process()` calls `json.loads(http_response.text)`
immediately after checking `http_response.ok`. When Trino returns
HTTP 200 with an empty body under load, this raises a bare
`JSONDecodeError` that is not in the `handled_exceptions` list of
`_retry_with`, so no retry is attempted and the exception
propagates unhandled to the caller.

Two changes to address this:

- Add a retry condition in `max_attempts.setter` for HTTP 200
  responses with an empty body, so the existing exponential
  backoff mechanism retries them — consistent with how 429/502/
  503/504 are already handled.
- Add a guard in `process()` that raises `TrinoConnectionError`
  with a descriptive message if an empty body reaches that point
  (e.g. after retries are exhausted), instead of leaking a raw
  `JSONDecodeError`.

## Non-technical explanation

When the server returned an empty response under load, the client
crashed with an unhelpful `JSONDecodeError`. It now retries
automatically and, if retries are exhausted, raises a clear error
message.

## Release notes

```markdown
* Fix client crash with `JSONDecodeError` on HTTP 200 with empty
  response body by retrying the request. ({issue}`596`)
```

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -641,6 +641,9 @@ class TrinoRequest:
                 # need retry when there is no exception but the status code is 429, 502, 503, or 504
                 lambda response: getattr(response, "status_code", None)
                 in (429, 502, 503, 504),
+                # need retry when the server returns 200 with an empty body (transient under load)
+                lambda response: getattr(response, "status_code", None) == 200
+                and not getattr(response, "text", "").strip(),
             ),
             max_attempts=self._max_attempts,
         )
@@ -723,6 +726,10 @@ class TrinoRequest:
             self.raise_response_error(http_response)
 
         http_response.encoding = "utf-8"
+        if not http_response.text.strip():
+            raise exceptions.TrinoConnectionError(
+                "received empty response from server (status 200)"
+            )
         response = json.loads(http_response.text)
         if "error" in response and response["error"]:
             raise self._process_error(response["error"], response.get("id"))
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -1094,6 +1094,32 @@ def test_429_error_retry(monkeypatch):
     assert post_retry.retry_count == 3
 
 
+def test_empty_200_response_retry(monkeypatch):
+    http_resp = TrinoRequest.http.Response()
+    http_resp.status_code = 200
+    http_resp._content = b""
+
+    post_retry = RetryRecorder(result=http_resp)
+    monkeypatch.setattr(TrinoRequest.http.Session, "post", post_retry)
+
+    get_retry = RetryRecorder(result=http_resp)
+    monkeypatch.setattr(TrinoRequest.http.Session, "get", get_retry)
+
+    attempts = 3
+    req = TrinoRequest(
+        host="coordinator",
+        port=8080,
+        client_session=ClientSession(user="test"),
+        max_attempts=attempts,
+    )
+
+    req.post("SELECT 1")
+    assert post_retry.retry_count == attempts
+
+    req.get("URL")
+    assert get_retry.retry_count == attempts
+
+
 @pytest.mark.parametrize("status_code", [
     501
 ])
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -813,6 +813,28 @@ def test_trino_connection_error(monkeypatch, error_code, error_type, error_messa
     assert error_message in str(error)
 
 
+def test_trino_process_empty_200_response_error():
+    req = TrinoRequest(
+        host="coordinator",
+        port=8080,
+        client_session=ClientSession(
+            user="test",
+            source="test",
+            catalog="test",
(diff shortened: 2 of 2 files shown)


emanuelsdsj:
I believe that both failures are related to the installation of dependencies, even before any tests were run.

hashhar (commented):
this either surfaced an existing flaky test or something actually needs updating.

```
=========================== short test summary info ============================
FAILED tests/integration/test_dbapi_integration.py::test_select_query_stats[json+zstd] - assert 22908 <= 21908
====== 1 failed, 808 passed, 8 skipped, 144 warnings in 165.63s (0:02:45) ======
```

trino/client.py
```diff
@@ -723,6 +726,10 @@ def process(self, http_response: Response) -> TrinoStatus:
             self.raise_response_error(http_response)
 
         http_response.encoding = "utf-8"
+        if not http_response.text.strip():
+            raise exceptions.TrinoConnectionError(
+                "received empty response from server (status 200)"
+            )
```
azawlocki-sbdt, line 732: Is this new code covered by any unit test?

emanuelsdsj: No, i need to create one

emanuelsdsj: Done in [8f19c07](https://github.com/trinodb/trino-python-client/pull/603/changes/8f19c07568d65fc60067969dbdb1971faf8195c7).

azawlocki-sbdt: Thanks!

azawlocki-sbdt:
The PR looks good and the failing test looks unrelated. But it failed on the last two runs. Let me look at it (tomorrow hopefully) to understand what's going on, before I give my ✅.

azawlocki-sbdt:
The failing test is flaky, the same assertion failed in https://github.com/trinodb/trino-python-client/actions/runs/27707800003/job/81960775843#step:5:64 and https://github.com/trinodb/trino-python-client/actions/runs/26816161416/job/79061743155#step:5:64.

azawlocki-sbdt (approved):
LGTM

hashhar (approved):

PR #604, opened 2026-05-18
Send heartbeat when downloading spooled segments

### LINKED ISSUE #554: Send query heartbeat while downloading results

azawlocki-sbdt:
## Description

Resolves https://github.com/trinodb/trino-python-client/issues/554

Trino can abandon a query if the client is silent for too long. When results are delivered via the spooling protocol, downloading a segment from S3/MinIO blocks the client and may cause query timeout if the download takes too long (the timeout is set in Trino configuration, current default is 5 min).

This PR implements a heartbeat mechanism: a background thread is sending periodic HEAD requests to the coordinator's nextUri while a spooled segment is being downloaded. The heartbeat is started and stopped around each `SpooledSegment.data` fetch inside `SegmentIterator._load_next_segment()`. Heartbeat thread is not started if spooling is not used.

Heartbeat requests are emitted in 30s intervals by default. Custom interval can be set using the new `heartbeat_interval` param for `Connection`. Setting it to `None` disables heartbeat. 

Unit tests are added in the new module `tests/unit/test_client_spooling.py`.
An integration test is added in `tests/integration/test_dbapi_integration.py`.

## Release notes

```markdown
```


trino/client.py
```diff
@@ -690,6 +696,14 @@ def get(self, url: str) -> Response:
     def delete(self, url: str) -> Response:
         return self._delete(url, timeout=self._request_timeout, proxies=PROXIES)
 
+    def head(self, url: str) -> Response:
+        return self._http_session.head(
```
damian3031:
```suggestion
        return self._head(
```
head() currently skips the retry wrapper that get/post/delete use. Could you wrap it in `max_attempts.setter` (`self._head = with_retry(self._http_session.head)`) and call self._head() from head()?

tests/development_server.py
```diff
@@ -17,6 +18,29 @@
...
 
+def get_trino_container(port: int):
+    """Find and return a running trino container.
+    Returns None if no matching container is found.
+    """
+    try:
+        client = docker.from_env()
+        for container in client.containers.list():
+            if container.name != "trino":
+                continue
```
damian3031:
```suggestion
        container = client.containers.get("trino")
```
Is this loop needed?

tests/development_server.py
```diff
@@ -17,6 +18,29 @@
...
+            if not any("trinodb/trino" in tag for tag in (container.image.tags or [])):
+                continue
+            host_ports = [
+                binding["HostPort"]
+                for bindings in container.ports.values() if bindings
+                for binding in bindings
+            ]
+            if str(port) in host_ports:
+                return container
+    except Exception:
```
damian3031: Could you narrow exception?
```suggestion
    except docker.errors.DockerException:
```

trino/client.py
```diff
@@ -1274,14 +1293,74 @@ def __repr__(self):
...
+        heartbeat_interval: Optional[float] = None,
+    ) -> None:
         self._segments = iter(segments if isinstance(segments, List) else [segments])
         self._mapper = mapper
         self._decoder = None
         self._rows: Iterator[List[List[Any]]] = iter([])
         self._finished = False
         self._current_segment: Optional[DecodableSegment] = None
+        self._request = request
+        self._heartbeat_interval = heartbeat_interval
```
damian3031:
```suggestion
        request: Optional[TrinoRequest] = None,
        *,
        heartbeat_interval: Optional[float] = None,
    ) -> None:
        self._segments = iter(segments if isinstance(segments, List) else [segments])
        self._mapper = mapper
        self._decoder = None
        self._rows: Iterator[List[List[Any]]] = iter([])
        self._finished = False
        self._current_segment: Optional[DecodableSegment] = None
        self._request = request
        self._heartbeat_interval = heartbeat_interval if request is not None else None
```
The new request parameter is required and positional. Could it be `Optional[TrinoRequest] = None`? When missing, the heartbeat is simply disabled — same semantics as `heartbeat_interval=None`. Keeps the constructor backward-compatible

trino/client.py
```diff
@@ -1274,14 +1293,74 @@ def __repr__(self):
...
+
+    def __init__(self, request: TrinoRequest, interval: float) -> None:
+        self._request = request
+        self._interval = interval
+        self._failures = 0
+        # The event for telling the heartbeat thread to exit
+        self._stop_event = threading.Event()
+
+    def __enter__(self) -> _RequestHeartbeat:
+        executor.submit(self._run)
```
damian3031: I wonder if we should use a dedicated daemon thread instead?
`threading.Thread(..., daemon=True)`
The shared executor (max_workers=4) is already used by SpooledSegment.acknowledge(), so under concurrent queries a heartbeat could be delayed waiting for a free slot.

tests/development_server.py
azawlocki-sbdt: Not really, I've simplified the code in e876eaedc938194cf7a6760ea8f91626282c8746.

tests/development_server.py
azawlocki-sbdt: Right, I've narrowed down both exception and scope in e876eaedc938194cf7a6760ea8f91626282c8746.

trino/client.py
azawlocki-sbdt: Done in e876eaedc938194cf7a6760ea8f91626282c8746

trino/client.py
azawlocki-sbdt: Agreed. Switched to a separate thread in e876eaedc938194cf7a6760ea8f91626282c8746 (making `self._failures` a local var in `_run()` is unrelated, I've just noticed it's not used outside `_run()`).

trino/client.py
azawlocki-sbdt: Agreed. I wanted to avoid handling situations in which the interval is set but the request is `None`, but it's better to keep `SegmentIterator` backwards compatible, as a public class. I've made both optional keyword args in e876eaedc938194cf7a6760ea8f91626282c8746, I'm also raising `ValueError` if only one is specified rather than silently treating both as `None`.

trino/client.py
```diff
@@ -1003,7 +1019,12 @@ def fetch(self) -> Union[List[Union[List[Any], Any]], Iterator[List[Any]]]:
             if self._fetch_mode == "segments":
                 return spooled
             # Return iterator directly, do NOT materialize with list()
-            return SegmentIterator(spooled, self._row_mapper)
+            return SegmentIterator(
+                spooled,
+                self._row_mapper,
+                self._request,
```
damian3031:
```suggestion
                request=self._request,
```
`request` is kwarg-only now

tests/integration/test_dbapi_integration.py
```diff
@@ -1896,7 +1897,7 @@ def test_segments_cursor(trino_connection):
             assert isinstance(segment.segment.ack_uri, str), (
                 f"Expected string for ack_uri, got {type(segment.segment.ack_uri)}"
             )
-        total += len(list(SegmentIterator(segment, row_mapper)))
+        total += len(list(SegmentIterator(segment, row_mapper, cur._query._request)))
```
damian3031:
```suggestion
        total += len(list(SegmentIterator(segment, row_mapper, request=cur._query._request)))
```
`request` is kwarg-only now

tests/unit/test_client_spooling.py
```diff
@@ -0,0 +1,202 @@
...
+    MockHB.assert_called_once_with(req, 30.0)
+    # Make sure MockHB instance is used as a context manager
+    MockHB.return_value.__enter__.assert_called_once()
+    MockHB.return_value.__exit__.assert_called_once()
+
+
+def test_no_heartbeat_for_inline_segment():
+    segment = DecodableSegment("json", None, mock.Mock(spec=InlineSegment))
+    mapper = mock.Mock()
+    iterator = SegmentIterator([segment], mapper, _mock_trino_request(), heartbeat_interval=30.0)
```
damian3031:
```suggestion
    iterator = SegmentIterator([segment], mapper, request=_mock_trino_request(), heartbeat_interval=30.0)
```
`request` is kwarg-only now

tests/unit/test_client_spooling.py
```diff
@@ -0,0 +1,202 @@
...
+    req._next_uri = "http://coordinator/v1/statement/q1/1"
+    query = TrinoQuery(req, query="SELECT 1")
+    query._row_mapper = mock.Mock()
+
+    with mock.patch.object(req, "get", return_value=_spooled_fetch_response()):
+        with mock.patch("trino.client.SegmentIterator") as MockSI:
+            MockSI.return_value = iter([])
+            query.fetch()
+
+    assert MockSI.call_args.args[2] is req
```
damian3031:
```suggestion
    assert MockSI.call_args.kwargs["request"] is req
```
`request` is kwarg-only now

trino/client.py
```diff
@@ -1274,14 +1295,78 @@ def __repr__(self):
...
+    """
+    MAX_FAILURES = 3
+
+    def __init__(self, request: TrinoRequest, interval: float) -> None:
+        self._request = request
+        self._interval = interval
+        # The event for telling the heartbeat thread to exit
+        self._stop_event = threading.Event()
+
+
```
damian3031: please use single blank line, flake8 would raise it 
```suggestion

```

trino/client.py
azawlocki-sbdt: good catch, I have this change locally but forgot to push it 🤦

wendigo:
@cla-bot check

wendigo (approved):

damian3031 (approved):

PR #605, opened 2026-05-25
Batch INSERT statements in executemany to reduce round trips

### LINKED ISSUE #602: SQLAlchemy bulk insert is executed per row
### Expected behavior

I'm using `sqlalchemy@2.0.49` to execute bulk insert statement through `trino`. I expect bulk insert of 1000 rows to be executed in one query, instead there're 1000 separate queries inserting 1 row.

```python3
with engine.connect() as conn:
    rows = [
        {"probe_id": i, "probe_name": random.choice(string.ascii_letters)}
        for i in range(1000)
    ]
    col_list = ", ".join(rows[0].keys())
    placeholders = ", ".join([f":{col}" for col in rows[0].keys()])
    quoted_target_table = f'"{_SCHEMA}"."{table_name}"'
    stmt = f"""
        INSERT INTO {quoted_target_table} ({col_list})
        VALUES ({placeholders})
    """
    conn.execute(text(stmt), rows)
    conn.commit()
```

### Actual behavior

Bulk insert of 1000 rows results in 1000 queries each performing 1 insert.

### Steps To Reproduce

```python3
import random
import string
import uuid

from sqlalchemy import create_engine, text
from sqlalchemy.engine import URL, Engine

from mysettings import trino_settings

_CATALOG = "test_catalog"
_SCHEMA = "test_schema"


def _build_engine() -> Engine:
    url = URL.create(
        "trino",
        username=trino_settings.trino_user,
        host=trino_settings.trino_host,
        port=trino_settings.trino_port,
        database=f"{_CATALOG}/{_SCHEMA}",
    )
    return create_engine(
        url,
        pool_pre_ping=True,
        connect_args={"http_scheme": trino_settings.trino_http_scheme},
    )


def main(engine: Engine, table_name: str) -> None:
    with engine.connect() as conn:
        conn.execute(text(f'CREATE SCHEMA IF NOT EXISTS "{_CATALOG}"."{_SCHEMA}"'))
        conn.commit()

        conn.execute(
            text(
                f'CREATE TABLE IF NOT EXISTS "{_SCHEMA}"."{table_name}" ('
                '"probe_id" INTEGER, "probe_name" VARCHAR)'
            )
        )
        conn.commit()

        rows = [
            {"probe_id": i, "probe_name": random.choice(string.ascii_letters)}
            for i in range(1000)
        ]
        col_list = ", ".join(rows[0].keys())
        placeholders = ", ".join([f":{col}" for col in rows[0].keys()])
        quoted_target_table = f'"{_SCHEMA}"."{table_name}"'
        stmt = f"""
            INSERT INTO {quoted_target_table} ({col_list})
            VALUES ({placeholders})
        """
        conn.execute(text(stmt), rows)
        conn.commit()

        count_result = conn.execute(
            text(f'SELECT COUNT(*) FROM "{_SCHEMA}"."{table_name}"')
        )
        row_count = int(count_result.scalar_one())
        conn.commit()
        print(f"Rows after insert: {row_count}")
        assert row_count == len(rows)


if __name__ == "__main__":
    engine = _build_engine()
    probe_suffix = uuid.uuid4().hex[:10]
    print(f"Running queries for table ID {probe_suffix}")
    table_name = f"tmp_insert_probe_{probe_suffix}"
    main(engine=engine, table_name=table_name)

```

### Log output

_No response_

### Operating System

MacOS Tahoe 26.3.1(a)

0.337.0

### Trino Server version

479

### Python version

3.14.3

srchilukoori:
## Description

Fixes #602.

`Cursor.executemany()` executes INSERT...VALUES one row at a time.
For 1,000 rows that means 1,000 HTTP round trips to Trino.

This adds a regex check for INSERT...VALUES patterns and batches
rows into multi-row VALUES clauses, 100 per request. Non-INSERT
statements still use the existing row-by-row path. Parameters are
formatted with the existing `_format_prepared_param()` method.

| Scenario | Row-by-row | Batched | Speedup |
|----------|-----------|---------|---------|
| 500 rows x 2 cols | 8.9s | 0.23s | ~39x |
| 1,000 rows x 25 cols (mixed types) | timeout | 3.1s (322 rows/s) | — |

## Non-technical explanation

Bulk inserts via SQLAlchemy `text()` or `cursor.executemany()` now
send fewer queries to Trino instead of one per row.

## Release notes

```markdown
* Batch INSERT...VALUES in `executemany()` into multi-row statements to reduce HTTP round trips. ({issue}`602`)
```

### DIFF
--- a/tests/unit/test_executemany.py
+++ b/tests/unit/test_executemany.py
@@ -1,0 +1,255 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from decimal import Decimal
+
+import pytest
+
+import trino.client
+from trino.dbapi import _INSERT_VALUES_RE
+from trino.dbapi import Connection
+from trino.dbapi import Cursor
+
+
+class FakeTrinoQuery:
+    """Hand-written fake replacing TrinoQuery for unit tests.
+
+    Records every SQL string passed and behaves like a successful INSERT.
+    """
+
+    instances = []
+
+    def __init__(self, request, query, legacy_primitive_types=False, fetch_mode="mapped"):
+        self._query = query
+        self._update_type = "INSERT"
+        FakeTrinoQuery.instances.append(self)
+
+    def execute(self):
+        return iter([])
+
+    @property
+    def query(self):
+        return self._query
+
+    @property
+    def update_type(self):
+        return self._update_type
+
+
+class FakeTrinoRequest:
+    pass
+
+
+def _make_cursor(batch_executemany=True):
+    cur = Cursor.__new__(Cursor)
+    cur._connection = Connection.__new__(Connection)
+    cur._connection._client_session = type("cs", (), {"timezone": None})()
+    cur._request = FakeTrinoRequest()
+    cur._iterator = None
+    cur._query = None
+    cur._legacy_primitive_types = False
+    cur._experimental_batch_executemany = batch_executemany
+    return cur
+
+
+@pytest.fixture
+def cursor():
+    """Create a real Cursor wired to FakeTrinoQuery with batching enabled.
+
+    Monkeypatches trino.client.TrinoQuery so the production
+    _executemany_batch_insert runs its real code path but creates
+    fakes instead of making HTTP calls.
+    """
+    FakeTrinoQuery.instances = []
+    original = trino.client.TrinoQuery
+    trino.client.TrinoQuery = FakeTrinoQuery
+
+    yield _make_cursor(batch_executemany=True)
+
+    trino.client.TrinoQuery = original
+
+
+class TestInsertValuesPattern:
+    def test_simple_insert(self):
+        assert _INSERT_VALUES_RE.match("INSERT INTO t (a, b) VALUES (?, ?)") is not None
+
+    def test_insert_with_schema(self):
+        sql = 'INSERT INTO "my_schema"."my_table" (col1, col2) VALUES (?, ?)'
+        assert _INSERT_VALUES_RE.match(sql) is not None
+
+    def test_insert_with_catalog_schema(self):
+        sql = 'INSERT INTO "catalog"."schema"."table" (a, b, c) VALUES (?, ?, ?)'
+        assert _INSERT_VALUES_RE.match(sql) is not None
+
+    def test_multiline_insert(self):
+        sql = '  INSERT INTO "schema"."table" (col1, col2)\n            VALUES (?, ?)\n        '
+        assert _INSERT_VALUES_RE.match(sql.strip()) is not None
+
+    def test_insert_no_columns(self):
+        assert _INSERT_VALUES_RE.match("INSERT INTO t VALUES (?, ?)") is not None
+
+    def test_select_not_matched(self):
+        assert _INSERT_VALUES_RE.match("SELECT * FROM t WHERE a = ?") is None
+
+    def test_update_not_matched(self):
+        assert _INSERT_VALUES_RE.match("UPDATE t SET a = ? WHERE b = ?") is None
+
+    def test_insert_select_not_matched(self):
+        assert _INSERT_VALUES_RE.match("INSERT INTO t SELECT * FROM s") is None
+
+    def test_trailing_semicolon_not_matched(self):
+        assert _INSERT_VALUES_RE.match("INSERT INTO t (a) VALUES (?);") is None
+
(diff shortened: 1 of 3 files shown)


srchilukoori:
@hashhar need your review on this PR

hashhar (commented):
Excellent feature and thanks for the tests.
I need to think a bit about if there are cases where this may do the wrong thing and hence should require an opt-in behaviour for a few releases maybe.

srchilukoori:
> Excellent feature and thanks for the tests. I need to think a bit about if there are cases where this may do the wrong thing and hence should require an opt-in behaviour for a few releases maybe.

@hashhar I added it as opt-in, with a default of false, as suggested. I'll also put together a plan for the edge cases.

srchilukoori:
@hashhar please review the updated PR

trino/dbapi.py
```diff
@@ -83,6 +84,13 @@
 
 logger = trino.logging.get_logger(__name__)
 
+_INSERT_VALUES_RE = re.compile(
+    r"\A(\s*INSERT\s+INTO\s+.+\bVALUES)\s*\([^)]+\)\s*\Z",
+    re.IGNORECASE | re.DOTALL,
+)
+
+_EXECUTEMANY_BATCH_SIZE = 100
```
hashhar, line 92: this needs to be configurable. The actual limit is the query text size configured on the server. Max on the server is 16MB IIRC but default is lower.

trino/dbapi.py
```diff
@@ -83,6 +84,13 @@
 
 logger = trino.logging.get_logger(__name__)
 
+_INSERT_VALUES_RE = re.compile(
+    r"\A(\s*INSERT\s+INTO\s+.+\bVALUES)\s*\([^)]+\)\s*\Z",
```
hashhar, line 88: seems to match all parens grouped things not just things with placeholders? This is only safe to do if ONLY placeholders are present in the VALUES.

trino/dbapi.py
```diff
@@ -658,15 +671,38 @@ def executemany(self, operation, seq_of_params):
...
+
+    def _executemany_batch_insert(self, prefix, seq_of_params):
+        for i in range(0, len(seq_of_params), _EXECUTEMANY_BATCH_SIZE):
+            batch = seq_of_params[i:i + _EXECUTEMANY_BATCH_SIZE]
+            value_rows = []
+            for params in batch:
+                formatted = ", ".join(self._format_prepared_param(p) for p in params)
+                value_rows.append("(%s)" % formatted)
+            sql = "%s %s" % (prefix, ", ".join(value_rows))
+            self._query = trino.client.TrinoQuery(
```
hashhar, line 698: this bypasses the SegmentCursor.

trino/dbapi.py
```diff
@@ -237,6 +246,7 @@ def __init__(
         self._transaction = None
         self.legacy_primitive_types = legacy_primitive_types
         self.legacy_prepared_statements = legacy_prepared_statements
+        self.experimental_batch_executemany = experimental_batch_executemany
```
hashhar, line 249: also update sqlalchemy `_url()`? Also add test for sqlalchemy.

trino/dbapi.py
```diff
@@ -658,15 +671,38 @@ def executemany(self, operation, seq_of_params):
...
+            for params in batch:
+                formatted = ", ".join(self._format_prepared_param(p) for p in params)
+                value_rows.append("(%s)" % formatted)
+            sql = "%s %s" % (prefix, ", ".join(value_rows))
+            self._query = trino.client.TrinoQuery(
+                self._request, query=sql,
+                legacy_primitive_types=self._legacy_primitive_types)
+            self._iterator = iter(self._query.execute())
+            if self._query.update_type is None:
+                raise NotSupportedError("Query must return update type")
```
hashhar, line 703: this is not tested

trino/dbapi.py
hashhar: Also tests for the false positives we want to avoid (e.g. placeholder + literal) are missing which would've caught this

hashhar (changes requested):

tests/unit/test_executemany.py
```diff
@@ -0,0 +1,255 @@
...
+            # Batched: single TrinoQuery with all rows
+            assert len(FakeTrinoQuery.instances) == 1
+            sql = FakeTrinoQuery.instances[0].query
+            assert "(1, 'a')" in sql
+            assert "(2, 'b')" in sql
+            assert "(3, 'c')" in sql
+        finally:
+            trino.client.TrinoQuery = original
+
+    def test_connection_threads_flag_to_cursor(self):
```
hashhar, line 240: weird test since we assign value ourself and then we assert. Should it not instead be testing by creating the connection with kwargs and verifying connection state?

trino/dbapi.py
```diff
@@ -658,15 +671,38 @@ def executemany(self, operation, seq_of_params):
 
         Return values are not defined.
         """
+        if not seq_of_params:
+            self.execute(operation)
+            return self
+
+        match = _INSERT_VALUES_RE.match(operation.strip())
+        if match and self._experimental_batch_executemany:
+            return self._executemany_batch_insert(match.group(1), seq_of_params)
```
hashhar, line 680: shouldn't the rowcount be accumulated across all batches and returned? this just returns value from last batch?

PR #608, opened 2026-06-16
Fix zombie queries on metadata

viniolivieri:
## Description
`_get_partitions` method were leaving zombie queries in Trino with the current implementation.
connection should be closed with or not with context manager usage. The fetch of data here is optionally, but I did implemented because otherwise, query will be consider CANCELLED and not FINISHED.

## Non-technical explanation
Fixed zombie queries from partitions metadata queries.

## Release notes

```markdown
[Queries stuck in FINISHING time](https://github.com/trinodb/trino-python-client/issues/463)
["SEE TABLE SCHEMA" queries to trino remain "FINISHING" when selecting iceberg tables](https://github.com/apache/superset/issues/34992)
```

### DIFF
--- a/tests/unit/sqlalchemy/test_dialect.py
+++ b/tests/unit/sqlalchemy/test_dialect.py
@@ -287,7 +287,7 @@ class TestTrinoDialect:
         assert result == partition_names
 
     def test_get_partitions_iceberg_table_returns_none(self):
-        # An Iceberg ``$partitions`` table has this exact shape and must be
+        # An Iceberg ``$partitions`` table has this exact shape and will be
         # recognized as metadata rather than treated as partition columns.
         partition_names = ["partition", "record_count", "file_count", "total_size", "data"]
         data_types = ["row(...)", "bigint", "bigint", "bigint", "row(...)"]
--- a/tests/unit/sqlalchemy/test_dialect.py
+++ b/tests/unit/sqlalchemy/test_dialect.py
@@ -258,6 +258,100 @@ class TestTrinoDialect:
         isolation_level = self.dialect.get_isolation_level(dbapi_conn)
         assert isolation_level == "SERIALIZABLE"
 
+    @staticmethod
+    def _partitions_connection(partition_names, data_types):
+        """Build a mock SQLAlchemy connection whose execute() returns a result.
+
+        The result exposes ``cursor.description`` shaped like the given
+        partition column names/types and supports both the SQLAlchemy 2.0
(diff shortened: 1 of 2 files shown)


trino/sqlalchemy/dialect.py
```diff
@@ -221,9 +222,20 @@ def _get_partitions(
             SELECT * FROM {schema}."{table_name}$partitions"
         """
         ).strip()
-        res = connection.execute(sql.text(query))
-        partition_names = [desc[0] for desc in res.cursor.description]
-        data_types = [desc[1] for desc in res.cursor.description]
+
+        if SQLALCHEMY_VERSION >= '2.0.0':
+            # Lower versions of SQLAlchemy doesn't support execution of query as ctxt man
```
hashhar, line 227: then why not always do it like the manual way? the context manager is just syntax sugar AFAIK for the same sequence of operations.

hashhar: also this path (> 2.0) isn't fetching data anyway so it'll show CANCELLED too.

BTW I don't agree that showing those as CANCELLED is wrong. We are unnecessarily fetching results for cosmetic benefits.

trino/sqlalchemy/dialect.py
```diff
@@ -221,9 +222,20 @@ def _get_partitions(
...
+            with connection.execute(sql.text(query)) as conn:
+                partition_names = [desc[0] for desc in conn.cursor.description]
+                data_types = [desc[1] for desc in conn.cursor.description]
+        else:
+            conn = connection.execute(sql.text(query))
+            partition_names = [desc[0] for desc in conn.cursor.description]
+            data_types = [desc[1] for desc in conn.cursor.description]
+            res = conn.fetchall() # Fetching data to consider query as FINISHED and not CANCELED
+            if not conn.closed:
```
hashhar, line 236: try ... finally?

trino/sqlalchemy/dialect.py
hashhar, line 1: fix import order and see CI failures

tests/unit/sqlalchemy/test_dialect.py
hashhar, line 1: can we avoid mocked tests since the sequence of responses is now hand constructed and doesn't really test anything meaningful

hashhar (changes requested):

hashhar:
continued at https://github.com/trinodb/trino-python-client/pull/627

PR #609, opened 2026-06-17
fix(sqlalchemy): add literal_processor to TIMESTAMP type

q8webmaster:
## Problem

`TIMESTAMP.literal_processor()` is not implemented, causing a `CompileError` when
`literal_binds=True` is used with a `datetime.datetime` value in a WHERE clause:

```
sqlalchemy.exc.CompileError: No literal value renderer is available for literal value
"datetime.datetime(2026, 6, 17, 9, 57, 43, 244000)" with datatype DATETIME
```

Concrete trigger: Apache Superset's `select_star()` builds a sample query with
`WHERE <partition_col> = <latest_partition_value>` using `literal_binds=True`. Any
table with a TIMESTAMP partition column hits this on the table metadata endpoint.

## Fix

Add `literal_processor` to `TIMESTAMP` emitting a Trino-compatible
`TIMESTAMP 'YYYY-MM-DD HH:MM:SS[.mmm]'` literal.

## Tests

Two unit tests added to `tests/unit/sqlalchemy/test_compiler.py`:
- datetime with sub-second precision → `TIMESTAMP '2026-06-17 09:57:43.244'`
- datetime without microseconds → `TIMESTAMP '2026-06-17 09:57:43'`

### DIFF
--- a/tests/unit/sqlalchemy/test_compiler.py
+++ b/tests/unit/sqlalchemy/test_compiler.py
@@ -9,6 +9,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import datetime
+
 import pytest
 from sqlalchemy import Column
 from sqlalchemy import ForeignKey
@@ -25,6 +27,7 @@ from sqlalchemy.sql import column
 from sqlalchemy.sql import table
 
 from tests.unit.conftest import sqlalchemy_version
+from trino.sqlalchemy.datatype import TIMESTAMP
 from trino.sqlalchemy.dialect import TrinoDialect
 
 metadata = MetaData()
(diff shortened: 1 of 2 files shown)


trino/sqlalchemy/datatype.py
```diff
@@ -81,6 +82,17 @@ def __init__(self, precision=None, timezone=False):
...
         self.precision = precision
 
+    def literal_processor(self, dialect):
+        def process(value):
+            if isinstance(value, datetime.datetime):
+                ts = value.strftime("%Y-%m-%d %H:%M:%S")
+                if value.microsecond:
+                    ts += f".{value.microsecond // 1000:03d}"
+                return f"TIMESTAMP '{ts}'"
+            return str(value)
```
hashhar, line 92: sql injection I think with something like `literal("x'; DROP TABLE t; --", TIMESTAMP())`

hashhar: also correctness issues in case of syntax errors: `WHERE ts = 2026-06-17` would parse as integer arithmetic instead of correctly

trino/sqlalchemy/datatype.py
```diff
@@ -81,6 +82,17 @@ def __init__(self, precision=None, timezone=False):
         super(TIMESTAMP, self).__init__(timezone=timezone)
         self.precision = precision
 
+    def literal_processor(self, dialect):
+        def process(value):
+            if isinstance(value, datetime.datetime):
+                ts = value.strftime("%Y-%m-%d %H:%M:%S")
```
hashhar, line 88: did you intentionally drop the zone? A zone-aware value would now be treated with session zone in Trino side. Maybe see `Cursor._format_prepared_param` for some ideas.

trino/sqlalchemy/datatype.py
```diff
@@ -81,6 +82,17 @@ def __init__(self, precision=None, timezone=False):
...
+                if value.microsecond:
+                    ts += f".{value.microsecond // 1000:03d}"
```
hashhar, line 90: truncates precision. equality predicates may no longer match.

Also `TIMESTAMP(precision=6)` would ignore the specified precision.

tests/unit/sqlalchemy/test_compiler.py
```diff
@@ -206,6 +209,24 @@ def test_try_cast(dialect):
     assert str(query) == 'SELECT try_cast("table".id as VARCHAR) AS id \nFROM "table"'
 
 
+def test_timestamp_literal_processor(dialect):
```
hashhar, line 212: add more tests:

- tz aware values
- higher precision values
- non-datetime values

hashhar (commented):
Also do we need to do the same for `TIME`?

hashhar:
superseded by https://github.com/trinodb/trino-python-client/pull/631

PR #612, opened 2026-06-27
Replace TrinoResult generator with class-based iterator for error recovery

### LINKED ISSUE #598: TrinoResult generator cannot recover from errors during lazy segment iteration
Python generators cannot be resumed after an unhandled exception, the frame is finalized and all subsequent next() calls raise `StopIteration`.

After #597 spooled segment downloads happen lazily inside the generator during iteration. If a transient error (e.g. S3 timeout) occurs mid-iteration, the caller sees the exception, but the generator is permanently dead.
Subsequent fetchone() calls also return None as if the query completed normally and silently drops remaining rows.

Before #597 this was theoretical issue since fetch() materialized all segments eagerly. I/O errors happened outside the generator.

From looking at other projects it seems the pattern is to use a class with `__next__` and `__iter__` instead of a generator and having a field on the class to hold state of the iterator so that we can signal to the caller what happened.

Here's a minimal repro:

```python
from trino.client import TrinoResult

class FailingIterator:
    def __init__(self):
        self._count = 0
    def __iter__(self):
        return self
    def __next__(self):
        if self._count >= 3:
            raise IOError("S3 segment download failed")
        self._count += 1
        return [self._count]

class FakeQuery:
    finished = True
    def fetch(self):
        return []

result = TrinoResult(FakeQuery(), FailingIterator())
it = iter(result)
```

```
next(it)  # [1]
next(it)  # [2]
next(it)  # [3]
next(it)  # raises IOError, caller sees this and retries
next(it)  # raises StopIteration, caller gets None from fetchone() and thinks query is done
```

srchilukoori:
## Problem

Fixes #598.

`TrinoResult.__iter__` uses a generator (`yield`). Python generators are permanently finalized after an unhandled exception propagates through the yield point — all subsequent `next()` calls raise `StopIteration`. Since #597 introduced lazy segment downloads during iteration, a transient I/O error (e.g. S3 timeout) mid-iteration kills the generator. `fetchone()` then returns `None` as if the query completed, silently dropping remaining rows.

## Root cause

Generator frame finalization is a Python language constraint, not a bug in the iteration logic itself. Once an exception propagates through `yield`, the generator's frame is deallocated and cannot be resumed.

## Fix

Replace the generator with a class-based iterator (`__iter__` returning `self`, `__next__` holding state in instance fields). The iteration logic is unchanged — batch prefetch via `fetch()` before exposing rows, same `_rownumber` tracking. The difference is that instance fields (`_row_iter`, `_next_rows`) survive exceptions, so callers can resume iteration after catching a transient error.

## Testing

Added `test_trino_result_iterator_survives_transient_error` — injects a `FailOnceIterator` that raises `IOError` on the 3rd element, verifies the iterator resumes and returns remaining rows after the caller catches the exception.

Full unit test suite passes (123 tests, 0 failures).

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -821,8 +821,10 @@ class TrinoResult:
     """
     Represent the result of a Trino query as an iterator on rows.
 
-    This class implements the iterator protocol as a generator type
-    https://docs.python.org/3/library/stdtypes.html#generator-types
+    This class implements the iterator protocol using __next__ so that
+    transient exceptions during iteration do not permanently kill the
+    iterator (unlike a generator, whose frame is finalized after an
+    unhandled exception).
     """
 
     def __init__(self, query, rows: List[Any]):
@@ -830,6 +832,8 @@ class TrinoResult:
         # Initial rows from the first POST request
         self._rows = rows
         self._rownumber = 0
+        self._row_iter: Optional[Iterator[Any]] = None
+        self._next_rows = None
 
     @property
     def rows(self):
@@ -844,15 +848,27 @@ class TrinoResult:
         return self._rownumber
 
     def __iter__(self):
+        return self
+
+    def __next__(self):
+        # Lazy init: prefetch the next batch before exposing current rows.
         # A query only transitions to a FINISHED state when the results are fully consumed:
         # The reception of the data is acknowledged by calling the next_uri before exposing the data through dbapi.
-        while not self._query.finished or self._rows is not None:
-            next_rows = self._query.fetch() if not self._query.finished else None
-            for row in self._rows:
-                self._rownumber += 1
-                yield row
+        if self._row_iter is None:
+            self._row_iter = iter(self._rows)
+            self._next_rows = self._query.fetch() if not self._query.finished else None
 
-            self._rows = next_rows
+        while True:
+            try:
+                row = next(self._row_iter)
+                self._rownumber += 1
+                return row
+            except StopIteration:
+                if self._next_rows is None:
+                    raise
+                self._rows = self._next_rows
+                self._row_iter = iter(self._rows)
+                self._next_rows = self._query.fetch() if not self._query.finished else None
 
 
 class TrinoQuery:
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -1145,6 +1145,49 @@ def test_trino_query_response_headers(sample_get_response_data):
         assert isinstance(result, TrinoResult)
 
 
+def test_trino_result_iterator_survives_transient_error():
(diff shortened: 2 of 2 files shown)


trino/client.py
```diff
@@ -844,15 +848,27 @@ def rownumber(self) -> int:
...
+            try:
+                row = next(self._row_iter)
+                self._rownumber += 1
+                return row
+            except StopIteration:
+                if self._next_rows is None:
+                    raise
+                self._rows = self._next_rows
+                self._row_iter = iter(self._rows)
+                self._next_rows = self._query.fetch() if not self._query.finished else None
```
azawlocki-sbdt, line 871: Seems to work correctly for the spooling case, when I/O errors originate from `next(self._row_iter)`.

But what if the error is raised by `self._query.fetch()` instead?

As I understand, in the spooling case it is not an issue, since `self._query.fetch()` does not occur once the list of segments is fetched right at the beginning. Except maybe when it does not fit in a single response page? But it can occur if the direct protocol is used and `self._query.fetch()` is called to get the next result page.

azawlocki-sbdt: Ultimately, what contract do we want the iterator to have?

Do we want to guarantee that it can be reused after an I/O error, and return more data without duplication, possibly skipping over some rows? It works that way for spooling but I'm not sure if it works that way for the direct protocol. For example, an error in `self._query.fetch()` in line 871 will leave `self._next_rows` unchanged and I think the iterator would go over them second time after the error.

azawlocki-sbdt: Another thought: if we assume s3 communication is more unreliable than coordinator communication, as #598 seems to assume, maybe we should just have a separate retry/backoff policy for s3 downloading?

azawlocki-sbdt: @hashhar, you're the reporter of #598, WDYT?

hashhar: I don't think your PR fixes it entirely. I actually see it can silently discard data now.

Consider that `_load_next_segment` acks the previous segment and then updates the next segment. The decoder then tries to download the segment and fails.

After you change the TrinoResult is still usable at this point and we call `next` again but we enter _load_next_segment which will ack the failed segment on this call without actually having received the data.

Also @azawlocki-sbdt is correct, for the non-spool case it will duplicate. But the dropped data case also exists. In the lazy-init block `_row_iter` is assigned first so if the next fetch() raises `_next_rows` stays None. On resume init is skipped and the initial rows are consumed and then `_next_rows` is None. Iteration will stop even though the query isn't finished and remaining rows are silently dropped.

srchilukoori:
@hashhar , @azawlocki-sbdt I'm closing this PR as #615 is addressing all issues raised in the reviews above.

PR #615, opened 2026-07-07
Make result iteration recoverable from errors

### LINKED ISSUE #598: TrinoResult generator cannot recover from errors during lazy segment iteration

hashhar:
## Description

TrinoResult implemented iteration as a generator. A generator that raises an unhandled exception is finalized so every subsequent next() raises StopIteration which dbapi's fetchone() maps to None. Since spooled segments are downloaded lazily during iteration, a transient error (e.g. an S3 timeout) mid-iteration permanently killed the iterator and silently dropped the remaining rows as if the query had completed normally.

Rewrite TrinoResult as a class-based iterator that keeps its state on the instance. An error propagates to the caller while the iterator stays usable, so a persistent error keeps surfacing instead of turning into StopIteration and a retried next() resumes where the failure happened.

Also make SegmentIterator retry-safe. Acknowledge and advance past a segment only after it was decoded successfully so retrying after a failed segment download re-decodes the same segment instead of acknowledging and skipping it.

## Release notes

```markdown
```

Fixes #598.

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -1448,3 +1448,95 @@ def test_decoder_factory_raises_with_message_on_missing_lz4():
             match=f"lz4 is not installed so json\\+lz4 encoding is not supported: {error_message}"
         ):
             factory.create("json+lz4")
+
+
+class _FinishedQuery:
+    """Query stub that is already finished. All rows come from the initial batch."""
+    finished = True
+
+    def fetch(self):
+        return []
+
+
+@pytest.mark.parametrize("consecutive_failures", (1, 2, 3))
+def test_trino_result_resumes_after_transient_error_in_rows_iterator(consecutive_failures):
+    class FlakyIterator:
+        """Fails a given number of times at the third row, succeeds when retried."""
+        def __init__(self, failures):
+            self._rows = iter([[1], [2], [3]])
+            self._served = 0
+            self._remaining_failures = failures
+
+        def __iter__(self):
+            return self
+
+        def __next__(self):
+            if self._served == 2 and self._remaining_failures > 0:
+                self._remaining_failures -= 1
+                raise IOError("segment download failed")
+            self._served += 1
+            return next(self._rows)
+
+    result = TrinoResult(_FinishedQuery(), FlakyIterator(consecutive_failures))
+    it = iter(result)
+    assert next(it) == [1]
+    assert next(it) == [2]
+    # Each retry surfaces the error again until the underlying iterator recovers
+    for _ in range(consecutive_failures):
+        with pytest.raises(IOError):
+            next(it)
+    # The iterator stays usable and resumes where the failure happened
+    assert next(it) == [3]
+    with pytest.raises(StopIteration):
+        next(it)
+    assert result.rownumber == 3
+
+
+def test_trino_result_reraises_persistent_error_instead_of_stopping():
+    class FailingIterator:
+        def __init__(self):
+            self._count = 0
+
+        def __iter__(self):
+            return self
+
+        def __next__(self):
+            if self._count >= 3:
+                raise IOError("segment download failed")
+            self._count += 1
+            return [self._count]
+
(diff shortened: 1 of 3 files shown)


tests/unit/test_client.py
```diff
@@ -1448,3 +1448,92 @@ def test_decoder_factory_raises_with_message_on_missing_lz4():
...
+            self._served = 0
+            self._failed = False
+
+        def __iter__(self):
+            return self
+
+        def __next__(self):
+            if self._served == 2 and not self._failed:
+                self._failed = True
+                raise IOError("segment download failed")
```
azawlocki-sbdt: Maybe also add a test case for ensuring that `TrinoResult` can recover from two or three consecutive errors in the underlying iterator, not just a single one.

tests/unit/test_client_spooling.py
```diff
@@ -200,3 +200,52 @@ def test_fetch_passes_request_and_interval_to_segment_iterator(heartbeat_interva
...
+            self._failing_segment = None
+            raise IOError("segment download failed")
+        return self._rows_by_segment[segment]
+
+
+def test_segment_iterator_retries_failed_segment_without_skipping_it():
+    seg1, seg2, seg3 = (_FakeSpooledSegment(name) for name in ("s1", "s2", "s3"))
+    segments = [DecodableSegment("json", None, seg) for seg in (seg1, seg2, seg3)]
+    iterator = SegmentIterator(segments, mapper=None)
+    iterator._decoder = _FlakyDecoder({seg1: [[1]], seg2: [[2]], seg3: [[3]]}, failing_segment=seg2)
```
azawlocki-sbdt: I think the first/last segments failing are corner cases worth covering. Maybe parametrize this test with `failing_segment_index=0,1,2`?

trino/client.py
```diff
@@ -844,15 +853,28 @@ def rownumber(self) -> int:
...
 
     def __iter__(self):
-        # A query only transitions to a FINISHED state when the results are fully consumed:
-        # The reception of the data is acknowledged by calling the next_uri before exposing the data through dbapi.
-        while not self._query.finished or self._rows is not None:
-            next_rows = self._query.fetch() if not self._query.finished else None
-            for row in self._rows:
-                self._rownumber += 1
-                yield row
+        return self
```
azawlocki-sbdt, line 856: NB: Side effect of the changes is that now we have `it_1 = iter(trino_result); it_2 = iter(trino_result); it_1 is it_2`. Previously we had `it_1 is not it_2`, but `it_1` and `it_2` shared the underlying result object anyway and would interfere.

azawlocki-sbdt (approved):
LGTM, code comments make it a good read. My only suggestion would be to cover more scenarios in tests, but it's non-blocking.

PR #616, opened 2026-07-09
Do not double-decode username, password and query values in SQLAlchemy URL

### LINKED ISSUE #611: unquote_plus double-decodes URL credentials, corrupting passwords containing +
## Expected behavior
  Passwords containing URL-safe special characters (like `+`, `%`, `@`) should survive the URL → connect round-trip unchanged:

  trino://user:pass+word@host/catalog  →  BasicAuthentication('user', 'pass+word')

  The password should not be modified beyond SQLAlchemy's own URL decoding.

## Version
  trino 0.337.0

  ## Bug location
  `trino/sqlalchemy/dialect.py` — `TrinoDialect.create_connect_args`

  ## Root cause
  `url.password` and `url.username` are already decoded by SQLAlchemy's `make_url()`. The dialect then applies `unquote_plus` again:

  trino://user:pass%2Bword@host/catalog

  → SQLAlchemy `make_url` decodes `%2B` → `+`
     `url.password = 'pass+word'`

  → `create_connect_args` calls `unquote_plus('pass+word')` → `'pass word'` ✗

  ## Impact
  Any password containing a literal `+` character gets corrupted to a space, causing "Access Denied: Invalid credentials" (401).

  ## Precedent
  SQLAlchemy itself fixed the exact same issue 11 years ago:
  https://github.com/sqlalchemy/sqlalchemy/issues/2873
  Their fix: use `unquote` for passwords, not `unquote_plus`.

  ## Fix
  Replace `unquote_plus(url.username)` and `unquote_plus(url.password)` with `url.username` / `url.password` directly — they're already decoded.

  (Alternatively, use `urllib.parse.unquote` instead of `unquote_plus` for these fields.)

  ## Workaround
  Monkey-patch before creating the engine:

  ```python
  import trino.sqlalchemy.dialect
  from urllib.parse import unquote
  trino.sqlalchemy.dialect.unquote_plus = unquote


### Steps To Reproduce

  ```python
  from trino.auth import BasicAuthentication

  # Password contains a literal '+' character
  url = "trino://user:pass+word@localhost:8080/catalog"

  engine = create_engine(url)

  # The dialect creates BasicAuthentication with corrupted password:
  #   expected: BasicAuthentication('user', 'pass+word')
  #   actual:   BasicAuthentication('user', 'pass word')
  #                       the '+' became a space ↑

  Or test the dialect directly:

  from sqlalchemy.engine.url import make_url
  from trino.sqlalchemy.dialect import TrinoDialect

  url = make_url("trino://user:pass+word@localhost:8080/catalog")
  dialect = TrinoDialect()
  _, kwargs = dialect.create_connect_args(url)

  auth = kwargs["auth"]
  print(auth._username)  # 'user'  
  print(auth._password)  # 'pass word'  ← should be 'pass+word'

  Result: connection fails with error 401: Access Denied: Invalid credentials.
  ```

### Log output

_No response_

### Operating System

Debian GNU/Linux 12 (bookworm)

0.337.0 

### Trino Server version

480

### Python version

python3.10

arpitjain099:
## Description

When connecting through SQLAlchemy, a value that contains a literal `+` (for example a password) was being corrupted into a space, which then failed authentication with a 401. The cause is a double decode: SQLAlchemy's `make_url()` already decodes the URL before the dialect sees it, and `create_connect_args` then ran the values through `unquote_plus` a second time. `unquote_plus` follows the form-encoding rule where `+` means space, so `pass+word` became `pass word`.

Looking at what `make_url()` actually hands over (same on SQLAlchemy 1.3.24, 1.4.54 and 2.0.51):

- userinfo (username and password) is decoded with `unquote`, so `%2B` becomes `+` but a literal `+` stays `+`.
- query values are decoded with `parse_qsl`, which does apply form-encoding, so both `+` and `%2B` are already resolved.
- `url.database` (catalog and schema) is the one component SQLAlchemy leaves url-encoded.

So the second `unquote_plus` was wrong on both the userinfo and the query parameters. A password given as `pass%2Bword` arrives from SQLAlchemy as `pass+word` and was then turned into `pass word`; the same happened to values like `access_token` or `client_tags` that contained a `+`. The fix uses the username, password and query values as-is (matching how the `roles` parameter was already handled) and keeps `unquote_plus` only on the catalog and schema, which come from `url.database`.

Added parameterized tests for a literal `+` and a percent-encoded `+` (`%2B`) in the userinfo, plus round-trip tests for the query parameters through the trino URL builder. The full `tests/unit/sqlalchemy/test_dialect.py` suite passes (24 tests). Closes #611.

## Non-technical explanation

Passwords, tokens and other connection values that contain a plus sign now work when you connect using a SQLAlchemy URL, instead of failing or being silently changed.

## Release notes

```markdown
* Fix SQLAlchemy connections corrupting a `+` character in the username, password or query parameters. ({issue}`611`)
```

### DIFF
--- a/tests/unit/sqlalchemy/test_dialect.py
+++ b/tests/unit/sqlalchemy/test_dialect.py
@@ -271,6 +271,86 @@ def test_trino_connection_basic_auth():
     assert cparams['auth']._password == password
 
 
+@pytest.mark.parametrize(
+    "userinfo, expected_user, expected_password",
+    [
+        # A literal '+' in the userinfo must stay a '+'. The userinfo is not
+        # form-encoded, so '+' does not mean space there.
+        ("user+name:pass+word", "user+name", "pass+word"),
+        # '%2B' is the percent-encoding of '+'. SQLAlchemy decodes it to '+', and
+        # the dialect must not decode it a second time into a space.
+        ("user%2Bname:pass%2Bword", "user+name", "pass+word"),
+        # A percent-encoded space really is a space.
+        ("user%20name:pass%20word", "user name", "pass word"),
+    ],
+)
+def test_trino_connection_basic_auth_with_plus_in_credentials(userinfo, expected_user, expected_password):
+    # SQLAlchemy's make_url already decodes the userinfo, so the dialect must not run
+    # it through unquote_plus again (that would turn a literal '+' into a space).
+    dialect = TrinoDialect()
+    url = make_url(f"trino://{userinfo}@host")
+    _, cparams = dialect.create_connect_args(url)
+
+    assert cparams['user'] == expected_user
+    assert isinstance(cparams['auth'], BasicAuthentication)
+    assert cparams['auth']._username == expected_user
+    assert cparams['auth']._password == expected_password
+
+
+def test_trino_connection_query_params_preserve_plus():
+    # Regression for the same double-decode on the other unquote_plus call sites.
+    # SQLAlchemy already decodes the query values, so a '+' round-tripped through the
+    # trino URL builder must survive. The catalog and schema are the exception: they
+    # come from url.database, which SQLAlchemy leaves url-encoded, so they are still
+    # decoded in the dialect.
+    dialect = TrinoDialect()
+    url = make_url(trino_url(
+        host="host",
+        user="us+er",
+        password="pa+ss",
+        catalog="ca+t",
+        schema="sc+h",
+        source="so+urce",
+        session_properties={"prop": "1+1"},
+        http_headers={"x-trino": "a+b"},
+        extra_credential=[("cred", "va+lue")],
+        client_tags=["ta+g"],
+        roles={"system": "ro+le"},
+    ))
+    _, cparams = dialect.create_connect_args(url)
+
+    assert cparams['user'] == 'us+er'
+    assert cparams['auth']._password == 'pa+ss'
+    assert cparams['catalog'] == 'ca+t'
+    assert cparams['schema'] == 'sc+h'
+    assert cparams['source'] == 'so+urce'
+    assert cparams['session_properties'] == {"prop": "1+1"}
+    assert cparams['http_headers'] == {"x-trino": "a+b"}
+    assert cparams['extra_credential'] == [("cred", "va+lue")]
+    assert cparams['client_tags'] == ["ta+g"]
+    assert cparams['roles'] == {"system": "ro+le"}
+
+
+def test_trino_connection_jwt_auth_preserves_plus():
+    # '%2B' in the token is decoded to '+' by SQLAlchemy and must not be decoded again.
+    dialect = TrinoDialect()
+    url = make_url("trino://host/?access_token=tok%2Ben")
+    _, cparams = dialect.create_connect_args(url)
+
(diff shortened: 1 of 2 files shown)


trino/sqlalchemy/dialect.py
```diff
@@ -127,13 +127,17 @@ def create_connect_args(self, url: URL) -> Tuple[Sequence[Any], Mapping[str, Any
         else:
             raise ValueError(f"Unexpected database format {url.database}")
 
+        # SQLAlchemy's make_url already percent-decodes the userinfo, so we use the
+        # username and password as-is. Running them through unquote_plus again would
+        # corrupt a literal '+' into a space (form-encoding only applies to the query
+        # string, not to the userinfo component).
```
hashhar: is the comment actually true? form-encoding applies to whole URL AFAIK. Can you add test for the other calls to unquote_plus here?

tests/unit/sqlalchemy/test_dialect.py
```diff
@@ -271,6 +271,19 @@ def test_trino_connection_basic_auth():
     assert cparams['auth']._password == password
 
 
+def test_trino_connection_basic_auth_with_plus_in_credentials():
```
hashhar: maybe parameterize the test to also test the value with literal `+` present?

trino/sqlalchemy/dialect.py
hashhar: hmmmm, also does this mean a literal `%2B` in the password for example would corrupt the URL? Can you test?

arpitjain099: You are right, that comment was misleading and I have rewritten it. I checked what make_url actually hands over, and it is the same on 1.3.24, 1.4.54 and 2.0.51:

- userinfo (username/password) is decoded with unquote, so %2B becomes + but a literal + stays a +.
- query values are decoded with parse_qsl, which does apply form-encoding, so there + already means space.
- url.database (catalog/schema) is the one part SQLAlchemy leaves url-encoded.

So form-decoding does happen on the query, which means the other unquote_plus calls had the same double-decode bug: an access_token or client_tag containing a + was getting corrupted too. I removed unquote_plus from the query params as well (that also matches how roles was already handled) and kept it only on catalog/schema. New test is test_trino_connection_query_params_preserve_plus, plus jwt/cert cases for those specific call sites.

tests/unit/sqlalchemy/test_dialect.py
arpitjain099: Done. test_trino_connection_basic_auth_with_plus_in_credentials is now parameterized over user+name:pass+word (literal +), user%2Bname:pass%2Bword (encoded +), and user%20name:pass%20word (encoded space). The first two come out as +, the last as a space.

trino/sqlalchemy/dialect.py
arpitjain099: Yes, confirmed. With the old code, pass%2Bword got corrupted: SQLAlchemy decodes %2B to +, then unquote_plus turned that + into a space, so it ended up as "pass word" and auth failed. The user%2Bname:pass%2Bword case in the parameterized test asserts it now stays "pass+word". The same was happening on the query side (for example access_token=...%2B...), which is why I also dropped unquote_plus from the query params.

PR #619, opened 2026-07-11
Send Content-Type charset on statement POST requests

### LINKED ISSUE #600: POST /v1/statement requests missing charset in Content-Type header
### Describe the bug
 
When submitting queries via `POST /v1/statement`, the `trino-python-client` sends the request body with `Content-Type: text/plain` without a charset parameter. This causes issues with Trino Gateway's `requestAnalyzerConfig` feature, which requires the charset to be explicitly set in order to parse the SQL body for routing rule evaluation.

### Expected behavior
 
The client should send `Content-Type: text/plain; charset=utf-8` since the Trino client protocol uses UTF-8 encoded SQL strings.

### Actual behavior
 
The client sends `Content-Type: text/plain` without charset, causing Trino Gateway to log:
 
```
DEBUG io.trino.gateway.ha.router.TrinoQueryProperties charset is not set in the request
```
 
And the Gateway fails to read the SQL body, resulting in empty `catalogs`, `tables`, and `body` fields in `trinoQueryProperties`. This breaks all SQL-aware routing rules.
 

### Steps to reproduce
 
1. Set up Trino Gateway with `requestAnalyzerConfig.analyzeRequest: true`
2. Submit a query using `trino-python-client`:
```python
 
conn = connect(
    host="gateway-host",
    port=38080,
    user="test",
)
cur = conn.cursor()
cur.execute("SELECT * FROM catalog.schema.table LIMIT 1")
```
 
3. Observe that the Gateway cannot parse the SQL body

### Workaround
 
Set the Content-Type header explicitly via a custom `requests.Session`:
 
```python
import requests
 
http_session = requests.Session()
http_session.headers.update({
    "Content-Type": "text/plain; charset=utf-8"
})
 
conn = connect(
    host="gateway-host",
    port=38080,
    http_session=http_session,
    # ...
)
```

### Log output

_No response_

### Operating System

Redhat

0.337.0

### Trino Server version

477

### Python version

3.13.12

### Additional context
 
Related Trino Gateway issue: https://github.com/trinodb/trino-gateway/issues/1032
 
While the Gateway should ideally handle missing charset by defaulting to UTF-8, the client should also follow best practices by including the charset in the Content-Type header.

hashhar:
## Description

Send "text/plain; charset=utf-8" on the statement POST matching the Trino JDBC client (StatementClientV1). The header is set only on POST and uses setdefault so a user-supplied Content-Type still takes precedence.

The client posted the SQL body to /v1/statement without a Content-Type header so no charset was advertised. This breaks Trino Gateway's request analyzer which needs the charset to parse the SQL body for routing rules (it logs "charset is not set in the request" and leaves catalogs/tables/body empty).

## Non-technical explanation

Fixes #600

## Release notes

```markdown
```

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -169,11 +169,60 @@ def test_request_headers(mock_get_and_post):
 
     req.post("URL")
     _, post_kwargs = post.call_args
-    assert_headers(post_kwargs["headers"])
+    post_headers = post_kwargs["headers"]
+    assert post_headers[constants.HEADER_CONTENT_TYPE] == constants.CONTENT_TYPE_TEXT_UTF8
+    # Content-Type is specific to POST requests. Strip it before reusing the
+    # shared assertions (which count the common headers present on both).
+    del post_headers[constants.HEADER_CONTENT_TYPE]
+    assert_headers(post_headers)
 
     req.get("URL")
     _, get_kwargs = get.call_args
-    assert_headers(get_kwargs["headers"])
+    get_headers = get_kwargs["headers"]
+    assert constants.HEADER_CONTENT_TYPE not in get_headers
+    assert_headers(get_headers)
+
+
+def test_post_sets_content_type_charset(mock_get_and_post):
+    _, post = mock_get_and_post
+
+    req = TrinoRequest(
+        host="coordinator",
+        port=8080,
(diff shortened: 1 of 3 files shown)


azawlocki-sbdt (approved):
LGTM

PR #627, opened 2026-08-14
Fix zombie $partitions queries left running by SQLAlchemy reflection

hashhar:
## Description

_get_partitions never closes the query result after reading cursor.description, so the underlying query is never cancelled and lingers on the coordinator until it eventually times out on its own. Close the result in a finally block so it's always cancelled, whether or not reading cursor.description succeeds.

## Non-technical explanation

Supersedes https://github.com/trinodb/trino-python-client/pull/608

## Release notes

```markdown
```

### DIFF
--- a/tests/integration/test_sqlalchemy_integration.py
+++ b/tests/integration/test_sqlalchemy_integration.py
@@ -10,6 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License
 import math
+import time
 import uuid
 from decimal import Decimal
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_sqlalchemy_integration.py
```diff
@@ -818,32 +772,53 @@ def test_get_view_names_raises(trino_connection):
...
 def _num_queries_containing_string(connection, query_string):
-    statement = sqla.text("select query from system.runtime.queries")
+    statement = sqla.text("select query from system.runtime.queries order by query_id desc offset 1 limit 1")
     result = connection.execute(statement)
     rows = result.fetchall()
     return len(list(filter(lambda rec: query_string in rec[0], rows)))
 
 
+def _assert_no_lingering_query(cur, query_substring, timeout_seconds=10):
+    """Fail if the most recent query matching query_substring is still running.
```
azawlocki-sbdt: nit: fails also if no query matching the given substring is found

azawlocki-sbdt (approved):
LGTM, except for one minor comment re the docstring of an aux function used only in tests

tests/integration/test_sqlalchemy_integration.py
hashhar: removed docstring, improved the fail to clairfy which case we hit for debugging

PR #630, opened 2026-08-18
Fix binding bytes values for VARBINARY columns

### LINKED ISSUE #626: bytes cannot be bound: the literal formatter calls .encode() on the value, so VARBINARY is unreachable from Python
**Versions**

- trino 0.338.0 (installed as `trino[sqlalchemy]`)
- SQLAlchemy 2.0.51
- Python 3.12.9, macOS
- Trino 476 in Docker, memory connector

## What happens

Trino has `VARBINARY` and stores binary fine. `trino.dbapi` can't send it. The literal formatter calls `.encode` on the value it was handed — which is what you'd do to a `str` — so a `bytes` raises before any statement leaves the client:

```
AttributeError: 'bytes' object has no attribute 'encode'
```

Via SQLAlchemy this shows up as a `LargeBinary` column that gets created successfully and then refuses every value. A column that exists and can't be written to is a bit worse than one that couldn't be created, because you find out later.

Everything else in the typed round trip works, for contrast — `Numeric`, `Date`, `DateTime`, `Time` and `Boolean` all came back as the values that went in. `Time` in particular works here and doesn't on two other engines I've tested, so this is a narrow gap rather than a general type problem.

## Suggested fix

Handle `bytes` in the literal formatter and emit `X'…'` (or `from_hex('…')`), both of which the server accepts.

For whatever it's worth, databend-sqlalchemy has the identical bug from the identical cause — it interpolates parameters into the statement text and decodes them as UTF-8 on the way. Rendering parameters into SQL seems to be the thing that gets `bytes` wrong, whoever is doing it.

hashhar:
## Description

`Binary()` in `trino/dbapi.py` called `.encode()` on every value unconditionally, raising `AttributeError` for `bytes`/`bytearray`/`memoryview`. It now passes those through as-is.

SQLAlchemy's default `literal_processor` for binary types also renders values as a quoted string, which breaks for non-UTF-8 bytes. Added a `VARBINARY` type in `trino/sqlalchemy/datatype.py` that renders `X'<hex>'`, wired in via `dialect.colspecs` for reflected and generic `LargeBinary`/`BINARY` columns.

## Non-technical explanation

You can now write `bytes` values into a `VARBINARY` column, whether you're using the raw client or SQLAlchemy.

## Release notes

```markdown
* Fix `AttributeError` when binding `bytes` values for `VARBINARY` columns,
  including through SQLAlchemy `LargeBinary` inserts and literal
  compilation.
  ({issue}`626`)
```

Fixes #626

### DIFF
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -566,8 +566,8 @@ class Cursor:
         if isinstance(param, str):
             return ("'%s'" % param.replace("'", "''"))
 
-        if isinstance(param, (bytes, bytearray)):
-            return "X'%s'" % param.hex()
+        if isinstance(param, (bytes, bytearray, memoryview)):
+            return "X'%s'" % bytes(param).hex()
 
         if isinstance(param, datetime.datetime) and param.tzinfo is None:
             datetime_str = param.strftime("%Y-%m-%d %H:%M:%S.%f")
@@ -811,8 +811,12 @@ def TimeFromTicks(ticks):
     return datetime.time(*datetime.localtime(ticks)[3:6])
 
 
-def Binary(string):
-    return string.encode("utf-8")
+def Binary(value):
+    if isinstance(value, (bytes, bytearray, memoryview)):
+        return bytes(value)
+    if isinstance(value, str):
+        return value.encode("utf-8")
+    raise TypeError(f"Binary() expects bytes, bytearray, memoryview or str, got {type(value).__name__}")
 
 
 class DBAPITypeObject:
--- a/trino/sqlalchemy/datatype.py
+++ b/trino/sqlalchemy/datatype.py
@@ -20,6 +20,7 @@ from typing import Type
 from typing import Union
 
 import sqlalchemy
+from sqlalchemy import exc
 from sqlalchemy import func
 from sqlalchemy import util
 from sqlalchemy.sql import sqltypes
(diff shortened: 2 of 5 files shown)


trino/sqlalchemy/datatype.py
```diff
@@ -89,6 +90,21 @@ def bind_expression(self, bindvalue):
         return func.JSON_PARSE(bindvalue)
 
 
+class VARBINARY(sqltypes.VARBINARY):
+    __visit_name__ = "VARBINARY"
+
+    def literal_processor(self, dialect):
+        def process(value):
+            if isinstance(value, (bytes, bytearray)):
+                return "X'%s'" % value.hex()
```
azawlocki-sbdt: nit: any reason to use `%` instead of an f-string here?

trino/sqlalchemy/datatype.py
```diff
@@ -89,6 +90,21 @@ def bind_expression(self, bindvalue):
         return func.JSON_PARSE(bindvalue)
 
 
+class VARBINARY(sqltypes.VARBINARY):
+    __visit_name__ = "VARBINARY"
+
+    def literal_processor(self, dialect):
+        def process(value):
+            if isinstance(value, (bytes, bytearray)):
```
azawlocki-sbdt: what about `memoryview`?

trino/dbapi.py
```diff
@@ -811,8 +811,10 @@ def TimeFromTicks(ticks):
     return datetime.time(*datetime.localtime(ticks)[3:6])
 
 
-def Binary(string):
-    return string.encode("utf-8")
+def Binary(value):
+    if isinstance(value, (bytes, bytearray, memoryview)):
+        return bytes(value)
+    return str(value).encode("utf-8")
```
azawlocki-sbdt: `Binary(1)` will now return b"1", previously it would raise `AttributeError`. Is it intended?

trino/sqlalchemy/datatype.py
hashhar: TIL. Thanks.

trino/dbapi.py
hashhar: not intentional. Thanks. made to match other dbapi adapters to return TypeError instead.

azawlocki-sbdt (approved):
LGTM

PR #631, opened 2026-08-18
Add literal_processor to TIMESTAMP, TIME and DATE SQLAlchemy types

hashhar:
`literal_binds=True` raised `CompileError` for `TIMESTAMP` and `TIME` columns and compiled `DATE` to a bare varchar literal that Trino rejects. Superset's `select_star()` hits this on partition columns.

The temporal formatting moves into a shared `trino.temporal` module used by both the DBAPI and SQLAlchemy paths so a value compiles to the same SQL:

- Named zones (`ZoneInfo`, pytz) render as zone names (`Asia/Kolkata`) instead of fixed offsets.
- A `ZoneInfo`-aware `time` keeps its zone offset instead of silently dropping it.
- Sub-minute UTC offsets raise instead of silently shifting the instant.
- Wrong Python types raise `CompileError` instead of turning into bad SQL.

Covers reflected columns only. The generic SQLAlchemy `DateTime`/`Time`/`Date` types stay uncovered (pre-existing gap).

## Release notes

```markdown
* Fix compiling `literal_binds` for `TIMESTAMP`, `TIME` and `DATE` columns in
  the SQLAlchemy dialect.
  ({issue}`609`)
```

### DIFF
--- a/tests/unit/sqlalchemy/test_compiler.py
+++ b/tests/unit/sqlalchemy/test_compiler.py
@@ -9,6 +9,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import datetime
+from zoneinfo import ZoneInfo
+
 import pytest
 from sqlalchemy import Column
 from sqlalchemy import ForeignKey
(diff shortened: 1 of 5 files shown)


hashhar:
IMO this is feeling more trouble than it's worth. I wonder if it's better to handle the SQLAlchemy types handling as a bigger more complete effort instead.

PR #632, opened 2026-08-18
Send heartbeats while the caller consumes fetched rows

### LINKED ISSUE #463: Queries stuck in FINISHING time
### Expected behavior

Trino coordinator does not recieved any heartbeat while the clent still fetching the results and the query failed.

### Actual behavior

Hi! 
In our Trino cluster, we are facing an issur that some queries remain stuck in the FINISHING state for an extended period before eventually failing with the error message: "Query was abandoned by the client, as it may have exited or stopped checking for query results." 

After conducting some investigation, it appears that this issue predominantly occurs when querying Trino using the Python client. Here's a breakdown of the observed flow:

1. In the main module, we execute the `TrinoQuery.execute` function with our query.
2. This function initiates a POST request to the Trino coordinator.
3. Subsequently, it sends a GET request to the `nextUri` to retrieve the initial batch of query results.
4. As the results start arriving, the query state transitions to FINISHING.
5. The execution of the `execute` function ends.
6. Following this, the `cursor.fetchall()` function in the main module iterates over the `nextUris`, yielding each received row to the client. However, after a certain duration of fetching query results, the query fails with the "query abandon" error (as mentioned above).

Any assistance on resolving this significant issue would be greatly appreciated.

Thank you!!

### Steps To Reproduce

1. Is it advisable to incorporate heartbeats to the coordinator while fetching results?

2. Would it be feasible to fetch multiple `nextUris` in parallel? I'm uncertain about this possibility due to the need to access `nextUris` as a linked list.

### Log output

_No response_

### Operating System

Windows

0.326.0

### Trino Server version

439

### Python version

3.9.3

hashhar:
## Description

Fixes https://github.com/trinodb/trino-python-client/issues/463.

The fix mirrors the Java client's `StatementClientV1`. `TrinoResult` runs a heartbeat check for every row it serves. Once a full `heartbeat_interval` passes without a request the client sends one HEAD to the query's current `nextUri` which resets the coordinator's abandonment timer. Every processed response defers the next beat, so an actively fetching query sends no extra requests.

There is no background thread. The HEAD goes out on the caller's own thread so it never races other traffic on the `requests.Session` which is not thread safe. The HEAD caps its timeouts at 5 seconds which bounds the pause it can add to a row loop.

Servers older than Trino 475 reject HEAD on the statement endpoint and the client falls back to the old behavior there.

Additionally `Connection.close()` now closes its cursors and cancels their running queries.

## Non-technical explanation

The client pings the coordinator between fetches, so slow consumers no longer fail with "Query was abandoned by the client".

## Release notes

### DIFF
--- a/trino/client.py
+++ b/trino/client.py
@@ -40,6 +40,7 @@ import base64
 import copy
 import functools
 import itertools
+import math
 import os
 import random
 import re
@@ -53,6 +54,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from email.utils import parsedate_to_datetime
 from enum import Enum
+from time import monotonic
 from time import sleep
 from typing import Any
 from typing import Callable
@@ -639,7 +641,6 @@ class TrinoRequest:
             self._get = self._http_session.get
             self._post = self._http_session.post
             self._delete = self._http_session.delete
-            self._head = self._http_session.head
             return
 
         with_retry = _retry_with(
@@ -658,7 +659,6 @@ class TrinoRequest:
         self._get = with_retry(self._http_session.get)
         self._post = with_retry(self._http_session.post)
         self._delete = with_retry(self._http_session.delete)
-        self._head = with_retry(self._http_session.head)
 
     def get_url(self, path: str) -> str:
         return "{protocol}://{host}:{port}{path}".format(
@@ -706,12 +706,16 @@ class TrinoRequest:
     def delete(self, url: str) -> Response:
         return self._delete(url, timeout=self._request_timeout, proxies=PROXIES)
 
-    def head(self, url: str) -> Response:
-        return self._head(
+    def _head(self, url: str, timeout: Union[float, Tuple[float, float]]) -> Response:
+        # requests disables redirect-following for HEAD by default. Follow them so a
+        # heartbeat behind a redirecting gateway still reaches the coordinator (like
+        # the Java client where OkHttp follows redirects).
+        return self._http_session.head(
             url,
             headers=self.http_headers,
-            timeout=self._request_timeout,
+            timeout=timeout,
             proxies=PROXIES,
+            allow_redirects=True,
         )
 
     @staticmethod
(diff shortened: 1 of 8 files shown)


wendigo:
@hashhar are heartbeats sent out of band or they are blocking the main thread/execution?

trino/client.py
```diff
@@ -1374,33 +1415,49 @@ class _RequestHeartbeat:
...
+        self._thread = threading.Thread(target=self._run, daemon=True)
+        self._thread.start()
 
-    def __exit__(self, *_) -> None:
+    def stop(self) -> None:
+        # Wait for an in-flight HEAD to finish. The caller issues its own request on the same
+        # requests.Session right after this returns, and that session is not thread safe.
         self._stop_event.set()
+        if self._thread is not None:
+            self._thread.join()
```
azawlocki-sbdt: I don't know if it's serious but the `join()` may block if the joined thread happens to do the heartbeat HEAD request. This shouldn't take long normally, but I think there's a retry mechanism with exponential backoff so if the server is not responsive the delay due to `join()` may be substantial.

hashhar:
I improved the heartbeating by stealing from the JDBC statement client design and adapting it for Python. The only difference is sync IO instead of async (which requires a much bigger surgery and switching libraries).

The blocking is bounded though and no thread safety issues.

Thanks for asking the question @azawlocki-sbdt, made me wonder if we can do it without threads in the first place.

trino/client.py
```diff
@@ -1477,13 +1511,7 @@ def _load_next_segment(self):
...
                                            .create(self._pending_segment.encoding))
 
-        if isinstance(self._pending_segment.segment, SpooledSegment) and self._request and self._heartbeat_interval:
-            # Downloading a spooled segment may take some time. In the meantime, send heartbeat
-            # requests so the coordinator doesn't think we lost interest and close the query.
-            with _RequestHeartbeat(self._request, self._heartbeat_interval):
-                rows = self._decoder.decode(self._pending_segment.segment)
-        else:
-            rows = self._decoder.decode(self._pending_segment.segment)
+        rows = self._decoder.decode(self._pending_segment.segment)
```
azawlocki-sbdt, line 1514: As far as I remember, the whole point of `_RequestHeartbeat` is to send heartbeat requests asynchronously while the main thread is blocked on `self._decoder.decode()`. Are you sure it's ok to turn it off?

azawlocki-sbdt:
> I improved the heartbeating by stealing from the JDBC statement client design and adapting it for Python. The only difference is sync IO instead of async (which requires a much bigger surgery and switching libraries).
> 
> The blocking is bounded though and no thread safety issues.
> 
> Thanks for asking the question @azawlocki-sbdt, made me wonder if we can do it without threads in the first place.

@hashhar, I like the current shape but I would make the HEAD requests asynchronous. We already have a four threads ThreadPoolExecutor in `client.py` used for sending segment acknowledgements asynchronously, I think it can be used for heartbeat requests as well.

trino/client.py
azawlocki-sbdt: OTOH, the heartbeat during segment download+decoding doesn't give us much, the design in this PR is more robust.

azawlocki-sbdt:
@hashhar, I came up with something like this, trying to make those heartbeat requests run asynchronously and keeping the requests made by a single `requests.Session` thread-safe: https://github.com/hashhar/trino-python-client/pull/4.

This adds extra complexity on top of what you've implemented here, so I don't know...

PR #633, opened 2026-08-19
Add allow_insecure_auth to opt out of the HTTP+auth guard

### LINKED ISSUE #494: Trino python client doesn't respect forwarded HTTP header
### Expected behavior

When Trino coordinator has http-server.process-forwarded=true set, connections over http can be assumed secured by LB or proxy and should be allowed.

### Actual behavior

This client setting prevents such connections from occurring:
https://github.com/trinodb/trino-python-client/blob/4c577742f8835cc2628bbfd55f25804ef4bc720e/trino/client.py#L438

### Steps To Reproduce

Connecting to trino via http using any authentication method will raise the error, with or without headers in place..

### Log output

_No response_

### Operating System

Windows

0.330.0

### Trino Server version

464

### Python version

3.11

hashhar:
## Description

Fixes #494.

The client rejects credentials over plain `http://`, with no opt-out. This breaks topologies where the transport is encrypted below the application layer, e.g. an mTLS-terminating service mesh sidecar: the connection looks like plain HTTP but isn't insecure end-to-end.

Add `allow_insecure_auth` (default `False`) to suppress the guard. If the coordinator has HTTPS enabled, Trino's own `http-server.authentication.allow-insecure-over-http` must also be set server-side.

## Release notes

```markdown
* Add an `allow_insecure_auth` connection option to opt out of the guard
  against sending authentication credentials over plain HTTP for
  connections that are encrypted below the application layer.
  ({issue}`494`)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,6 +469,30 @@ conn = connect(
 )
 ```
 
+### Authentication over insecure transport
+
+By default, the client refuses to send authentication credentials (for example `BasicAuthentication`) over an
+unencrypted `http://` connection, raising a `TrinoAuthError`. This is intentional as it avoids credentials being sent
+in plaintext.
+
+If the connection is already encrypted below the application layer (for example the client only ever talks to a
+Trino coordinator through an mTLS-terminating service mesh sidecar on `localhost`), this check gets in the way even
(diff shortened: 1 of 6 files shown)


azawlocki-sbdt (approved):
LGTM

PR #634, opened 2026-08-21
Parse the host argument once and validate the endpoint

hashhar:
## Description

`Connection` parsed the `host` argument by hand on top of a `urlparse` call that returns no hostname for a scheme-less value. An IPv6 literal, a bare `host:port` and a host carrying a path each built a malformed request URL that failed at connect time with an error naming neither the argument nor the problem.

The `http_scheme` and `port` arguments were not validated either. A wrongly cased scheme, a scheme or port contradicting one given in `host`, and a port outside the valid range all went through unchecked.

Parse the argument once and reject what cannot be represented. The scheme and port rules move out of `Connection` and `TrinoRequest` into helpers both call.

A path in `host` is now rejected rather than carried. The port landed after the path, so no such setup ever reached the wire intact. The gateway-prefix use case behind it is real and still unserved. A dedicated `path_prefix` argument is the shape to add if we want it.

## Release notes

```markdown
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,12 @@ $ pip install trino
 
 Use the DBAPI interface to query Trino:
 
-if `host` is a valid url, the port and http schema will be automatically determined. For example `https://my-trino-server:9999` will assign the `http_schema` property to `https` and port to `9999`.
+`host` accepts a full URL, a hostname on its own, or a `hostname:port`. For
+example `https://my-trino-server:9999` sets `http_scheme` to `https` and `port`
+to `9999`. A scheme or port in `host` that contradicts the `http_scheme` or
+`port` argument is an error. A missing scheme is inferred from the port, and a
+missing port from the scheme. Wrap an IPv6 literal in brackets when it carries
+a port, as in `[::1]:8080`.
 
 ```python
 from trino.dbapi import connect
--- a/trino/dbapi.py
+++ b/trino/dbapi.py
@@ -147,9 +147,15 @@ def _resolve_endpoint(
     hostname = host if parsed_host.hostname is None else parsed_host.hostname + parsed_host.path
 
     if parsed_host.scheme:
-        scheme = parsed_host.scheme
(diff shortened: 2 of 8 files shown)


trino/dbapi.py
```diff
@@ -133,6 +136,94 @@ def connect(*args, **kwargs):
...
+    if given_port is not None and not constants.MIN_PORT <= given_port <= constants.MAX_PORT:
+        raise ValueError(
+            f"Invalid port {given_port!r}, expected {constants.MIN_PORT}-{constants.MAX_PORT}."
+        )
+
+    if host_scheme:
+        try:
+            scheme = trino.client.normalize_http_scheme(host_scheme)
+        except ValueError:
+            raise ValueError(
```
azawlocki-sbdt, line 208: A bit awkward. Maybe have `normalize_http_scheme()` accept an optional `context` argument and, if provided, use it to add "in {context!r}" to the error message?

README.md
```diff
@@ -29,7 +29,12 @@ $ pip install trino
 
 Use the DBAPI interface to query Trino:
 
-if `host` is a valid url, the port and http schema will be automatically determined. For example `https://my-trino-server:9999` will assign the `http_schema` property to `https` and port to `9999`.
+`host` accepts a full URL, a hostname on its own, or a `hostname:port`. For
+example `https://my-trino-server:9999` sets `http_scheme` to `https` and `port`
+to `9999`. A scheme or port in `host` that contradicts the `http_scheme` or
+`port` argument is an error. A missing scheme is inferred from the port, and a
+missing port from the scheme. Wrap an IPv6 literal in brackets when it carries
+a port, as in `[::1]:8080`.
```
azawlocki-sbdt, line 37: To me it reads as if `it` referred to IPv6 literal. Maybe
```suggestion
missing port from the scheme. Wrap an IPv6 literal in brackets when a port
is also present, as in `[::1]:8080`.
```

PR #635, opened 2026-08-31
Replace requests with httpx2 and add an asynchronous client

wendigo:
Migrate the HTTP stack from requests to httpx2, with HTTP/2 negotiated via ALPN on TLS connections and transparent HTTP/1.1 fallback.

- Extract the transport-agnostic protocol logic (headers, response processing, retry decisions, session state, segment decoding) into trino/_protocol.py, shared by the sync and async clients.
- Rework trino.auth: Authentication now contributes get_http_auth() and get_client_arguments() since verify/cert/trust_env are constructor-only in httpx; OAuth2 is an httpx2.Auth generator flow working for both sync and async clients; Kerberos/GSSAPI use an in-repo SPNEGO flow on python-gssapi (trino/_spnego.py) instead of requests_kerberos and requests_gssapi.
- Add trino.aio, an asynchronous DBAPI-like interface on httpx2.AsyncClient with spooled protocol support (async segment downloads, heartbeat task, acknowledgment tasks drained on close).
- Replace httpretty with a hand-written recording MockTrinoServer over httpx2.MockTransport serving sync and async unit tests alike.

## Release notes

```markdown
```


wendigo:
Thoughts @hashhar @azawlocki-sbdt ?

azawlocki-sbdt:
@wendigo, I see three commits, or even PRs, here (requests -> httpx2, refactor to _protocol.py, async API), not just a single one. Also, what is the main motivation for the changes?

hashhar:
i like most of the changes (their direction). I've been working on some similar "design" improvements myself recently and this seems to match that too.
I agree that some of this (like the protocol concerns extraction, lib change etc.) should probably be split.

PR #640, opened 2026-09-22
Do not retry spooling-related GET requests with status code 200 and empty body

azawlocki-sbdt:
## Description

Since #603, the client treats all GET requests to which the server responds with status code 200 and empty body as failed and retries them. But the empty body is a normal response for segment acknowledgements requests, they should not be retried if the status is 200.

This PR fixes that by turning off body emptiness checking (effectively reverting PR #603) for spooling-releted requests. For simplicity, emptiness checking is turned off for spooling data requests, as well as spooling
acknowledgement requests.

Additional minor change is to access `content` rather than `text` when checking for emptiness, to avoid decoding binary responses to text just to see if they are not empty.

## Non-technical explanation

A fix for the bug that resulted in retrying successful segment acknowledgement requests, leading to performance degradation when downloading a large number of segments.

## Release notes

```markdown
```


lozbrown:
Isn't this a duplicate fix if this https://github.com/trinodb/trino-python-client/pull/637

azawlocki-sbdt:
> Isn't this a duplicate fix if this #637

Yes, it is! I wasn't aware of #637, thanks @lozbrown. Closing the current PR in favor of #637 which has wider scope (rightfully) and better description.

